PR #3, opened 2016-05-17
modifies baseurl in config to represent repo name

ablwr:
This modification allows links using the baseurl variable to persist when viewing the site on http://matroska-org.github.io/matroska-specification/


ablwr:
(Oops, being a good Git citizen and not overlapping intellectually-distinct commits)

PR #4, opened 2016-05-18
adds ebml order guidelines page

ablwr:
This adds the order guidelines from https://matroska.org/technical/order/index.html into this repository, converted to markdown (which is then converted to html upon rendering the gh-pages site). Resolves https://github.com/Matroska-Org/matroska-specification/issues/2

### DIFF
--- a/order_guidelines.md
+++ b/order_guidelines.md
@@ -1,0 +1,47 @@
+---
+layout: default
+---
+While in EBML there is no particular order for elements, in Matroska it is necessary to make sure some elements are placed in a certain order for better playback, seeking and editing efficiency. So far there was no guideline on the order of elements, and some existing programs may not feed these, but by following these guidelines your Matroska files should be optimum for playback in various conditions.
+
+# Level 1 elements
+
+Matroska only needs a few level 1 elements to be playable: Segment Info, Track Info, Clusters. These elements have to be present in a Matroska file. And the Segment Info and Track Info <em>must</em> always be found before the Clusters (either because they're located before the first Cluster or they can be found via Meta Seek elements; see below). All the other elements can be omitted although Cues (index) improve the playback experience greatly.
+
+After a Matroska file has been created it may still be edited. For example chapters, tags or cover art attachments can be added. To do that the Meta Seek needs to be updated and also some elements may be voided or extended. Therefore it's a good idea to use some padding after theses Elements. It is better to put that padding at the Level 1 so that they can easily be reshuffled without having to move them at the end.
+
+## Meta Seek
+
+When 1 Meta Seek Head is present, it should be the first element in a Segment as it should reference the other level 1 elements in that segment (except maybe the long list of Clusters). There is also the possibility that in some cases there are two Meta Seek Head sections like for files produced with mkvpropedit. In that case the first one is also placed first with only the position of the level 1 elements except the Clusters. The second Meta Seek is placed at the end and contains a lengthy list of all Clusters (and not the other level 1 elements). Placing the first Meta Seek Head other than the first position of the Segment would make it a lot less useful So it <em>must</em> be the first element of a Cluster.
+
+It is possible to put the first Cluster of the Segment in the Meta Seek to make it easier to find the actual content of the file.
+Meta Seek should also contain some padding (preferably outside and just after the Meta Seek) for a link to Tags, Chapters and Attachments, when they are not already found in the Segment. More padding may also be a good idea to add links in the future for other sections that may be appended in the Matroska file.
+
+## Cues (index)
+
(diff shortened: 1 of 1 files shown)


order_guidelines.md
```diff
@@ -0,0 +1,47 @@
...
+
+After a Matroska file has been created it may still be edited. For example chapters, tags or cover art attachments can be added. To do that the Meta Seek needs to be updated and also some elements may be voided or extended. Therefore it's a good idea to use some padding after theses Elements. It is better to put that padding at the Level 1 so that they can easily be reshuffled without having to move them at the end.
+
+## Meta Seek
+
+When 1 Meta Seek Head is present, it should be the first element in a Segment as it should reference the other level 1 elements in that segment (except maybe the long list of Clusters). There is also the possibility that in some cases there are two Meta Seek Head sections like for files produced with mkvpropedit. In that case the first one is also placed first with only the position of the level 1 elements except the Clusters. The second Meta Seek is placed at the end and contains a lengthy list of all Clusters (and not the other level 1 elements). Placing the first Meta Seek Head other than the first position of the Segment would make it a lot less useful So it <em>must</em> be the first element of a Cluster.
```
robUx4: Do we want to reference actual softwares like mkvpropedit ?
The second meta seek should be discouraged IMO. There's no need to keep a reference of all Clusters, the Cues have that. And if you can edit backward to set the position of the second meta seek, you can surely have a placeholder for the Cues in there.

dericed: I propose the initial commit mirror the current status of the associated page at http://matroska.org/technical/specs/index.html so we're starting from a familiar place and change from there.

order_guidelines.md
```diff
@@ -0,0 +1,47 @@
...
+  <table border="5" align="center"><tr><td><table width="150" border="0"><tr><td bgcolor="#FFCC99">Meta Seek</td></tr><tr><td bgcolor="#FFFFCC">Segment Info</td></tr><tr><td bgcolor="#00FFFF">Tracks Info</td></tr><tr><td bgcolor="#FF6666">Chapters</td></tr><tr><td bgcolor="#66FF00">Attachments</td></tr><tr><td bgcolor="#CC99FF">Tags</td></tr><tr><td bgcolor="#FFCC00">Cues</td></tr><tr><td bgcolor="#66FF99">Clusters</td></tr></table></td></tr></table>
+
+# Cluster Timecode
+
+As each Block+BlockGroup/SimpleBlock in a Cluster needs the Cluster timecode, the Cluster Timecode <em>must</em> be the first element in the Cluster.
+
+# CRC-32
+
+The EBML <a href="/technical/specs/index.html#CRC-32">CRC-32 element</a> value applies to all the data enclosed in its parent EBML element (except for the CRC-32 element itself). So it <em>must</em> be the first element so that the reader can know a CRC is applied and that all coming data have to feed that CRC checker.
```
robUx4: Should say that it must be even before the Cluster Timecode.

robUx4:
OK, we'll fix it little by little.

order_guidelines.md
robUx4: 2016-05-18 7:34 GMT+02:00 Dave Rice <notifications@github.com>:

> In order_guidelines.md
> <https://github.com/Matroska-Org/matroska-specification/pull/4#discussion_r63647333>
> :
>
> > @@ -0,0 +1,47 @@
> > +---
> > +layout: default
> > +---
> > +While in EBML there is no particular order for elements, in Matroska it is necessary to make sure some elements are placed in a certain order for better playback, seeking and editing efficiency. So far there was no guideline on the order of elements, and some existing programs may not feed these, but by following these guidelines your Matroska files should be optimum for playback in various conditions.
> > +
> > +# Level 1 elements
> > +
> > +Matroska only needs a few level 1 elements to be playable: Segment Info, Track Info, Clusters. These elements have to be present in a Matroska file. And the Segment Info and Track Info <em>must</em> always be found before the Clusters (either because they're located before the first Cluster or they can be found via Meta Seek elements; see below). All the other elements can be omitted although Cues (index) improve the playback experience greatly.
> > +
> > +After a Matroska file has been created it may still be edited. For example chapters, tags or cover art attachments can be added. To do that the Meta Seek needs to be updated and also some elements may be voided or extended. Therefore it's a good idea to use some padding after theses Elements. It is better to put that padding at the Level 1 so that they can easily be reshuffled without having to move them at the end.
> > +
> > +## Meta Seek
> > +
> > +When 1 Meta Seek Head is present, it should be the first element in a Segment as it should reference the other level 1 elements in that segment (except maybe the long list of Clusters). There is also the possibility that in some cases there are two Meta Seek Head sections like for files produced with mkvpropedit. In that case the first one is also placed first with only the position of the level 1 elements except the Clusters. The second Meta Seek is placed at the end and contains a lengthy list of all Clusters (and not the other level 1 elements). Placing the first Meta Seek Head other than the first position of the Segment would make it a lot less useful So it <em>must</em> be the first element of a Cluster.
>
> I propose the initial commit mirror the current status of the associated
> page at http://matroska.org/technical/specs/index.html so we're starting
> from a familiar place and change from there.
>

I agree. Merged.


> —
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub
> <https://github.com/Matroska-Org/matroska-specification/pull/4/files/b638e6a2e2f9a812cef38c3ab32398d83846ad07#r63647333>
>



-- 
Steve Lhomme
Matroska association Chairman

PR #15, opened 2016-05-24
Converts the other pages on Matroska.org

ablwr:
This concludes the conversion from Matroska.org to a gh-pages-based site. Some of these might be worth having conversations about on the issues page or via CELLAR (the contributing and license pages come to mine -- Github has these features built-in). But I prefer the separation of intellectually different work and having that done in different PRs/issues.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,3 +1,32 @@
+---
+layout: default
+---
+
+# License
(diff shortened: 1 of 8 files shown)


mbunkus:
Thanks.

PR #16, opened 2016-05-24
order guidelines rewrite

dericed:


dericed:
fyi this pr is dependent on https://github.com/Matroska-Org/matroska-specification/pull/11

mbunkus:
Thanks.

PR #24, opened 2016-07-02
Info elements

dericed:


ebml_matroska.xml
```diff
@@ -63,7 +63,7 @@
       </element>
       TimecodeScale When combined with <a href="http://www.matroska.org/technical/specs/index.html#TimecodeScaleDenominator">TimecodeScaleDenominator</a> the Timestamp scale is given by the fraction TimecodeScale/TimecodeScaleDenominator in seconds.-->
       <element name="Duration" level="2" id="0x4489" type="float" minver="1" range="&gt; 0x0p+0">
-        <documentation lang="en">Duration of the Segment (based on TimecodeScale).</documentation>
+        <documentation lang="en" type="definition">Duration of the Segment measured in nanoseconds when scaled by TimecodeScale.</documentation>
```
mbunkus: I find your new definition more confusing then the one it replaces. Basically the value contained in this element must be multiplied by `TimecodeScale`. The result is the duration in nanoseconds.

The specs use various terms to refer to such situations including but not limited to `based on TimecodeScale`, `scaled by TimecodeScale` and `scaled via TimecodeScale`, simply `scaled` and the corresponding negations. We should probably settle on one type of wording and add an explanatory section.

Maybe my confusion has more to do with my German background. In German we often use the word "wenn" for referring to both "when" and "if… then" (even though we also have "falls" which is much more appropriate for "if … then"). I guess my point is that I don't like introducing yet another choice of wording for scaling, and that we should really settle on a single choice of words.

notes.md
```diff
@@ -101,19 +101,19 @@ The position in some Elements refers to the position, in octets, from the beginn
...
 
-### Segment linking
+### Linked Segments
 
-#### Hard linking
+#### Hard Linking
 
-This linking can also be called splitting. It's the operation of cutting one Segment in several parts. The resulting parts should play as if it was just one part (the original Segment). That means the timecode of each part follows the ones from the previous parts. The track numbers are the same. The chapters only match the current Segment (unless the edition is ordered, where all parts should be in each Segment). And most important, the NextUID and PrevUID points the respective Segment UIDs.
+Hard Linking (also called splitting) is the operation of cutting one Segment in several smaller linked Segments. The resulting Linked Segments should play as if it was the original Segment. Within a Linked Segment the timecode of each Segment MUST follow consequtively and the track numbers and timescale MUST be the same amongst the Segments. The chapters only match the current Segment (unless the Edition is ordered, where all parts should be in each Segment). Within a Linked Segment the NextUID and PrevUID MUST reference the respective Segment UIDs of the next and previous Segments.
```
mbunkus: - `…cutting one Segment into…` instead of `…cutting one Segment in`?
- `…smaller Linked Segments` (capitalize the `L`)?
- `…Linked Segments should play as if they were…` (`Segments` is plural, therefore the verb should use plural form, too, shouldn't it?)
- `consecutively` instead of `consequtively`

notes.md
```diff
@@ -101,19 +101,19 @@ The position in some Elements refers to the position, in octets, from the beginn
...
 
 #### Soft linking
 
-Soft linking is used by codec chapters. They can reference another Segment and jump on that Segment. The way the Segments are described are internal to the chapter codec and unknown to the Matroska level. But there are Elements in the Segment Information (such as `ChapterTranslate`) that can translate a value representing a Segment in the chapter codec and to the current Segment UID. All Segments that could be used in a file/segment this way should be marked as members of the same family (SegmentFamily), so that the player can quickly switch from one to the other.
+Soft Linking is used by codec chapters. They can reference another Segment and jump on that Segment. The way the Segments are described are internal to the chapter codec and unknown to the Matroska level. But there are Elements in the Segment Information (such as `ChapterTranslate`) that can translate a value representing a Segment in the chapter codec and to the current Segment UID. All Segments that could be used in a Linked Segment in this way SHOULD be marked as members of the same family via the SegmentFamily Element, so that the player can quickly switch from one to the other.
```
mbunkus: - `The way… is described…` verb should use singular form

mbunkus: And why is `Soft linking` not capitalized like `Hard linking`?

notes.md
```diff
@@ -101,19 +101,19 @@ The position in some Elements refers to the position, in octets, from the beginn
...
+This kind of linking is a mix between Hard Linking and Soft Linking. Each Segment linked is independant from the others (standalone, unlike hard linked ones). But it should be treated as a hard-link by the player. Medium linking is done through chapters using the `ChapterSegmentUID` Element and only makes sense for ordered editions. The Segment matching the UID should be played as if it was part of the original Segment (Segment it's linked from) and then resume playback in the original Segment. That means the timecodes of the following content should be shifted by the duration of the linked Segment. As for hard-linking, the resulting Segment edition should be played and considered as one.
```
mbunkus: - `…unlike Hard Linked ones…` and similar issues in other places
- And why is `Medium linking` not capitalized like `Hard linking`?
- `…the UID SHOULD be played…`?
- `…following content SHOULD be shifted…`?

mbunkus:
Thanks. Alright, I'm though with reviewing & commenting.

dericed:
through

mbunkus:
You didn't push your changes, it seems.

dericed:
sorry, i was referencing your typo in https://github.com/Matroska-Org/matroska-specification/pull/24#issuecomment-230107561

ebml_matroska.xml
dericed: How about: `Duration of the Segment based on TimecodeScale in nanoseconds`.?

mbunkus:
Aaaaah indeed :)

ebml_matroska.xml
mbunkus: `based on TimecodeScale` sounds good to me.

dericed: Also an issue that TimecodeScale's definition refers to the content with `timestamp` terms.

notes.md
dericed: A vocab issue with `Linked Segment`.
Is a `Linked Segment` a series of individual linked Segments?
And `Linked Segments` refers to the Segments that are linked?
These are very close in meaning, but which should be preferred?

ebml_matroska.xml
mbunkus: Yes, but that's not in the scope of this pull request. I'd rather have one that only deals with timestamp vs. timecode usage.

Problem is if you rename elements building applications will break if the libraries follow suit. And if they don't there's discrepancy between the specs and the libraries that can be confusing.

notes.md
dericed: I'm going to add an intro paragraph to Linked Segments above Hard Linking, because some of the mandates in the Hard Linking section (track consistency) actually apply to each type of Linking.

mbunkus: I see what you mean. We should probably use the capitalized form when we're talking about the aggregate as a whole and the lower-case `linked` when talking about the individual pieces. That's what you're preferring, too, if I read you correctly.

Applied to the definition in question the original `…in severfal smaller linked Segments` is (and all other uses of `linked`/`Linked`) correct.

ebml_matroska.xml
dericed: +1

notes.md
dericed: Sounds good. We'll use Linked Segment and linked Segments.

dericed: I don't like the vocabulary of Hard, Soft and Medium Linking very much as it conveys so little about what it is and these terms aren't used often if at all outside of their definition. For instance the article at http://mod16.org/hurfdurf/?p=8 explain these concepts in great detail but never adopts the vocabulary. Any alternate vocabulary to suggest here?

mbunkus: I see your point, but I don't have any great ideas at the moment :/

dericed: I'm going to change this sentence too. `Hard Linking (also called splitting) is the operation of cutting one Segment into several smaller linked Segments.` To make Hard Linking it is not required to cut one Segment, but one could make the smaller linked Segments from another means.

dericed:
Updated now, with a lot of rewrite to the Linked Segments section. Please re-review.

dericed:
ok, i force pushed once more, done for now

notes.md
```diff
@@ -95,25 +95,37 @@ Overlay tracks should be rendered in the same 'channel' as the track it's linked
...
+#### Hard Linking
 
-#### Soft linking
+Hard Linking (also called splitting) is the process of creating a Linked Segment by relating multiple Segments using the PrevUID and NextUID Elements. Within a Linked Segment the timecode of each Segment MUST follow consecutively in linking order. With Hard Linking, the chapters of any Segment within the Linked Segment MUST only reference the current Segment. With Hard Linking, the NextUID and PrevUID MUST reference the respective Segment UIDs of the next and previous Segments. The first Segment of a Linked Segment MUST have a NextUID Element but not a PrevUID Element. The last Segment of a Linked Segment MUST have a PrevUID Element but not a NextUID Element. The middle Segments of a Linked Segment MUST have both a NextUID Element and a PrevUID Element.
```
mbunkus: I like this introductory chapter. I have only one nitpick: all element names should be set in `backticks`. You're doing it in the other parts of this commit, too, so just do it everywhere, even if the original text didn't.

notes.md
```diff
@@ -95,25 +95,37 @@ Overlay tracks should be rendered in the same 'channel' as the track it's linked
...
 
-Soft linking is used by codec chapters. They can reference another Segment and jump on that Segment. The way the Segments are described are internal to the chapter codec and unknown to the Matroska level. But there are Elements in the Segment Information (such as `ChapterTranslate`) that can translate a value representing a Segment in the chapter codec and to the current Segment UID. All Segments that could be used in a file/segment this way should be marked as members of the same family (SegmentFamily), so that the player can quickly switch from one to the other.
+As an example four Segments may be Hard Linked as a Linked Segment through cross-referencing each other with `SegmentUID`, `PrevUID`, and `NextUID` as in this table.
 
-#### Medium linking
+file name  | SegmentUID                       | PrevUID                          | NextUID
```
mbunkus: The element names should use `backticks`. The file names could use them as well.

notes.md
```diff
@@ -95,25 +95,37 @@ Overlay tracks should be rendered in the same 'channel' as the track it's linked
...
+#### Soft Linking
+
+Soft Linking is used by codec chapters. They can reference another Segment and jump to that Segment. The way the Segments are described are internal to the chapter codec and unknown to the Matroska level. But there are Elements within the `Info` Element (such as `ChapterTranslate`) that can translate a value representing a Segment in the chapter codec and to the current `SegmentUID`. All Segments that could be used in a Linked Segment in this way SHOULD be marked as members of the same family via the SegmentFamily Element, so that the player can quickly switch from one to the other.
```
mbunkus: Backticks for `SegmentFamily`, please.

notes.md
```diff
@@ -95,25 +95,37 @@ Overlay tracks should be rendered in the same 'channel' as the track it's linked
...
+
+#### Medium Linking
+
+WMedium Linking creates relationships between Segments using Ordered Chapters and the `ChapterSegmentUID` Element. A Segment Edition with Ordered Chapters may contain Chapters that reference timestamp ranges from other Segments. The Segment referenced by the Ordered Chapter via the `ChapterSegmentUID` Element should be played as part of a Linked Segment. The timestamps of Segment content referenced by Ordered Chapters MUST be adjusted according to the cumulative duration of the the previous Ordered Chapters.
```
mbunkus: Superfluous `W` before `Medium`

notes.md
```diff
@@ -95,25 +95,37 @@ Overlay tracks should be rendered in the same 'channel' as the track it's linked
...
+
+As an example a file named intro.mkv may have a `SegmentUID` of 0xb16a58609fc7e60653a60c984fc11ead. Another file called program.mkv may use a Chapter Edition that contains two Ordered Chapters. The first chapter references the Segment of intro.mkv with the use of a `ChapterSegmentUID`, `ChapterSegmentEditionUID`, `ChapterTimeStart` and optionally a `ChapterTimeEnd` element. The second chapter references content within the Segment of program.mkv. A player should recognize the Linked Segment created by the use of `ChapterSegmentUID` in an enabled Edition and present the reference content of the two Segments together.
```
mbunkus: - Maybe backtickes for th`intro.mkv`?

mbunkus:
Done with my re-review.

notes.md
dericed: Should `backticking` Elements apply to `Segment`?

dericed: re ^^^ it feels odd, since Segments would have to be re-written as `Segment` Elements to stay plural.

mbunkus: Good points. I do think that we can and should make a distinction between the element `Segment` and the concept of a Segment. The paragraph above talks about the concept, not the concrete elements; therefore not backticking seems fine to me.

dericed: or could do `Segment`s, but that looks wrong

mbunkus: `Segment`s looks so wrong it's not even funny :)

dericed:
@mbunkus I adopted all your suggestions. Please re-re-review.

mbunkus:
Aaaaand merged.

PR #25, opened 2016-07-03
adds wavpack page to spec site

ablwr:
As requested by @mbunkus on CELLAR, corrects oversight of missing page.


mbunkus:
Thanks :)

PR #26, opened 2016-07-04
rewrite Position References as Segment Position

dericed:


dericed:
I realize there's a few elements that use this concept in their definition where I'm not sure if they specifically refer to the beginning of the Data section of the Segment Element or just that of another Element. There's `SeekPosition`, `Position`, `ReferenceVirtual`, `CueClusterPosition`, `CueCodecState`,  `CueRefCluster`, and `CueRefCodecState`.

Are these positions all measured from the first Segment's Data or are the measured from other locations?

mbunkus:
Oh yikes, a very good point.

I'm pretty sure that all players treat `SeekPosition` and `CueClusterPosition` as referring to their parent Segment's Data, not the first Segment's in a Linked Segment case. That's how mkvmerge creates files, and that's how players were developed.

The `CodecState` is not really used that much yet, therefore I'm entirely uncertain how mkvmerge treats it with respect to splitting and linking. I'm guessing it's at least somewhat buggy if not entirely buggy. What probably should happen (and what would produce the least problems to players) is that the last `CodecState` element for each track should be repeated at the beginning of the next output file (mkvmerge does not do this at the moment if I remember correctly). Consequently `CueCodecState` could also refer to its parent Segment's Data (this is what mkvmerge already does when it writes a `CodecState` element).

Next up is `Position`. I've never seen it used in the wild; mkvmerge doesn't write it. I don't know if there are players that support it. However, we can make some assumptions. A cluster cannot be split across mulitple segments. Therefore it makes sense to let `Position` be relative to its parent Segment's Data and not to the first Segment's Data.

That leaves us with `CueRefCluster`, `CueRefCodecState` and `ReferenceVirtual`. All of those are deprecated, and I haven't seen any of those used in the wild. Therefore I'm dispassionate about how to treat them. It would fit into the theme I've got going above if they referred to their parent Segment's Data, too.

So… As I see it none of those are actually measured from the first Segment's Data in a Linked Segment case.

dericed:
The original Positions Reference concept was not really clear on this distinction either. Should we split the concept of Segment Offset into two, Segment Offset and Maybe-Linked Segment Offset?

Are you sure `CueRefCluster`, `CueRefCodecState`, and `ReferenceVirtual` are deprecated. How so? The specdata.xml file doesn't seem to use any indicator for this? https://github.com/Matroska-Org/foundation-source/blob/f4e62175f7fdd5f8641d1f0021f5c1fb0ed664cd/spectool/specdata.xml#L63

dericed:
`no(@minver) and (@webm='0' or no(@webm)))` is ... a bad way to mark deprecated elements.

Can we update these to provide `maxver` attributes? Any reason for this other complex strategy?

mbunkus:
A "bad way" is putting it mildly. I totally concur; `maxver` attributes would be much better.

I honestly have no idea why this is the way it is. @robUx4 do you have any insight?

robUx4:
specdata and the code around it was built to create outputs as close to the original file. That's why deprecated in called `version2`. It dates back when I did the CSS to mark elements as not to be used in Matroska version1 files. IIRC that's also the define in libmatroska to enable/disable deprecated stuff.

To mark elements as deprecated there should probably be a flag in specdata.xml rather that this convoluted test I found matching all the cases.

robUx4:
As for the position, I agree linked segments are not very good for that. It's not even safe from a remux (if it's done with discarding the fact the Segment is linked). It's also not safe if you concatenate files and add (EBML) junk in between files. So maybe we need to write in the section about linked segments that concatenating and remuxing must be done very carefully for such segments.

dericed:
I'm hoping that the EBML Schema format defined in the EBML spec can supersede the specdata format. Is this feasible? (happy to help with xsl and makefiles to help with the transition)

In EBML Schema there is a `maxver` attribute. Currently I mark an Element as DEPRECATED if the [maxver](https://github.com/Matroska-Org/matroska-specification/blob/gh-pages/ebml_matroska.xml#L214) is less than the [version](https://github.com/Matroska-Org/matroska-specification/blob/gh-pages/ebml_matroska.xml#L2) of the doctype in the EBML Schema.

But as @mbunkus notes it is also true that `no(@minver) and (@webm='0' or no(@webm)))` means deprecated. Can we just add `maxver` attributes to those elements instead, I think it would be far clearer.

robUx4:
Yes, whatever becomes the official machine readable file to generate the Matroska semantics will be used to generate the code. That's the whole point of generating the code: so it absolutely matches the specs.

mbunkus:
I agree. Adding `maxver=1` to all those deprecated elements that are currently only covered by `no(@minver) and (@webm='0' or no(@webm)))` is the way to go.

mbunkus:

dericed:
Psst @mbunkus, regarding your comments about `Position` [here](https://github.com/Matroska-Org/matroska-specification/pull/26#issuecomment-230740374), here's some samples in the wild which contain the `Position Element`.

http://www.archive.org/download/BSG54RazorFlashback/BSG54 - Razor Flashback.mkv
http://www.archive.org/download/gov.archives.arc.35924.b/gov.archives.arc.35924.b.mkv
http://www.archive.org/download/gov.archives.arc.35957.b/gov.archives.arc.35957.b.mkv
http://www.archive.org/download/gov.archives.arc.40099.bs/gov.archives.arc.40099.bs.mkv
http://www.archive.org/download/gov.archives.arc.65422.b/gov.archives.arc.65422.b.mkv
http://www.archive.org/download/gov.dod.dimoc.30152.b/gov.dod.dimoc.30152.b.mkv
http://www.archive.org/download/gov.dod.dimoc.30153.b/gov.dod.dimoc.30153.b.mkv
http://www.archive.org/download/MatroskaSampleWithEmbeddedPreservationMetadata/pres_metadata_sample_20110608.mkv
http://www.archive.org/download/org.c-span.172939-1/org.c-span.172939-1.mkv
http://www.archive.org/download/org.c-span.194519-1/org.c-span.194519-1.mkv
http://www.archive.org/download/org.c-span.205093-1/org.c-span.205093-1.mkv
http://www.archive.org/download/org.c-span.206134-1/org.c-span.206134-1.mkv

dericed:
I resolved the conflict but sent a related email to CELLAR at https://mailarchive.ietf.org/arch/msg/cellar/FtezW_ESxfn6Tx9HeiGaO2EgQLY.

dericed:
ping on https://mailarchive.ietf.org/arch/msg/cellar/FtezW_ESxfn6Tx9HeiGaO2EgQLY. As the proposed change contradicts historic Matroska documentation, seeking some comments.

mbunkus:
I'm strongly in favor of adjusting the documentation to match reality. Otherwise the vast majority of existing linked files would be invalidated. It's also unrealistic to assume that existing software that already deals with existing files will be adjusted to match the specs — especially as interesting workarounds would be required to still be able to handle old files where the positions refer to the encapsulating segment.

Another reason I prefer positions to refer to the encapsulating and not the first segment is that position references referring to the first segment can only be resolved properly if all prior segments are currently available. This would mean that you cannot use references in segment 5 if one of the segments 1–4 is missing. It would make for a very bad user experience.

dericed:
I pushed a substantial update to this PR. As discussed, I removed references to Segment Position referencing to the first Segment in the case of Linked Segments, now it just references the containing Segment. This contradicts earlier documentation on Positions Reference, but the change seems to have consensus.

I also updated the ebml_matroska.xml to use the new term.

Note that in a few Elements that use Segment Position, the value of `0` has special meaning. For instance when `SeekHead` contains a `0` it refers the first position within the Segment Element; however when the `Position Element` contains a `0` it means that it is within a live stream.

Should the definition of Segment Position contain a caveat that sometimes the definition of the field could overrule the definition of Segment Position?

notes.md
```diff
@@ -88,9 +88,22 @@ An Octet refers to a byte made of 8 bits.
...
+The `Segment Position` of an `Element` refers to the position of the `Element ID` of that `Element`, measured in octets, from the beginning of the `Element Data` section of the containing `Segment Element`. In other words, the `Segment Position` of an `Element` is the distance in octets from the beginning of its containing `Segment Element` minus the size of the `Element ID` and `Element Data Size` of that `Segment Element`. The `Segment Position` of the first `Child Element` of the `Segment Element` is 0. An `Element` which is not stored within a `Segment Element`, such as the `Elements` of the `EBML Header`, do not have a `Segment Position`.
```
mbunkus: "…refers to the position of the `Element ID`…" — maybe be even more precise? "…refers to the position of the first octet of the `Element ID`…"?

notes.md
```diff
@@ -88,9 +88,22 @@ An Octet refers to a byte made of 8 bits.
...
+
+
+This table presents an example of `Segment Position` by showing a hexadecimal representation of a very small Matroska file with labels to show the offsets in octets. The file contains a `Segment Element` with an `Element ID` of `0x18538067` and a `MuxingApp Element` with an `Element ID` of `0x4D80`.
```
mbunkus: Please insert a sub-heading stating that an example follows.

dericed: I added a `## Example of Segment Position` sub-section header and also introduced a new sub-section.

```
## Segment Position Execption

`Elements` that are defined to store a `Segment Position` MAY define reserved values to indicate a special meaning.
```

dericed:
Nudging @robUx4 for an approval or comment, since I think this PR contradicts earlier documentation that he wrote.

notes.md
```diff
@@ -88,9 +88,27 @@ An Octet refers to a byte made of 8 bits.
...
+The `Segment Position` of an `Element` refers to the position of the first octet of the `Element ID` of that `Element`, measured in octets, from the beginning of the `Element Data` section of the containing `Segment Element`. In other words, the `Segment Position` of an `Element` is the distance in octets from the beginning of its containing `Segment Element` minus the size of the `Element ID` and `Element Data Size` of that `Segment Element`. The `Segment Position` of the first `Child Element` of the `Segment Element` is 0. An `Element` which is not stored within a `Segment Element`, such as the `Elements` of the `EBML Header`, do not have a `Segment Position`.
+
+## Segment Position Execption
```
robUx4: `Exception` typo

PR #34, opened 2016-07-26
ebml_matroska: add DisplayUnit value 4 for unknown Aspect Ratio

robUx4:
As discussed in this thread https://mailarchive.ietf.org/arch/msg/cellar/x1F00MwqytPjrcNru6Kk2CWQ474

We want a way to set an unknown aspect ratio when it's unknown in the source.
Coincidentally the new value will also be unknown to existing players.


mbunkus:
LGTM.

PR #35, opened 2016-07-30
Update LICENSE.md

bastik-tor:
Fixing a simple grammar mistake in line 7 and getting rid of a 'do' in line 26. Two commits so you can cherry-pick, if the 'do' is supposed to be kept.


mbunkus:
Thanks.

PR #38, opened 2016-07-30
Update diagram.md

bastik-tor:
One commit to fix mistakes. The second one can be used if the suggested changes seem reasonable.

I am unaware of what is supposed to happen to sections that are written in first-person ("I"), some of which "promise" later additions to the document.


mbunkus:
All changes look good to me. Thanks.

PR #39, opened 2016-07-30
More deprecated elements

mbunkus:
This pull request adds `minver` and `maxver` attributes to those elements that are currently shown gray-on-gray on the [old specs page](https://www.matroska.org/technical/specs/index.html). The intention of those was to have them deprecated as well, but we never added the version attributes to the corresponding XML file.

As I couldn't figure out when we decided not to use those elements after all I decoded to use 3 as the maximum version.

See also [this comment by @robUx4](https://github.com/Matroska-Org/matroska-specification/pull/26#issuecomment-232134896).

### DIFF
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -105,7 +105,7 @@
         <element name="Block" level="3" id="0xA1" type="binary" minOccurs="1" minver="1">
           <documentation lang="en">Block containing the actual data to be rendered and a timestamp relative to the Cluster Timecode. (see <a href="http://www.matroska.org/technical/specs/index.html#block_structure">Block Structure</a>)</documentation>
         </element>
-        <element name="BlockVirtual" level="3" id="0xA2" type="binary" webm="0">
+        <element name="BlockVirtual" level="3" id="0xA2" type="binary" minver="0" maxver="0" webm="0">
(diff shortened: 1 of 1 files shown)


robUx4:
In the "original" specs all those elements are marked as belonging to no version so I think it's fair to put maxver to 0 for all of them.

mbunkus:
Hmm good point. I've changed the commit to use `minver="0" maxver="0"` and force-pushed.

dericed:
Perhaps we need to explain somewhere that in the context of Matroska there is no version zero, but that version=0 is an alias for invalid or experimental.

dericed:
Reading this late, but am a bit concerned before this patch these elements were valid in many versions of Matroska but after the patch many files in the world are suddenly invalid.

mbunkus:
Can you quantify that? How many files do you have that contain one of those elements? Or is that more of a gut feeling?

robUx4:
Actually the DivX elements exist in the wild. But that's a different question, should be keep track of what exists in other specs ? Like WebM and DivX. IMO we should make it easy for them to define what they support (like extra elements they can add to the XML) but it should not more.
In the case of WebM mkvalidator already differentiate the checks done for WebM and Matroska. But not DivX IIRC. There's also no way to know if a file is DivX or not.

PR #40, opened 2016-07-31
Spelling fixes for notes.md

bastik-tor:
Just a few little fixes.

Commit cef2acf is irrelevant here. I wanted to sync the other changes from the source, yesterday, but that gets me unwanted commit in these pull requests.


mbunkus:
They all look fine to me. Thanks.

PR #42, opened 2016-07-31
add the explicit parent value to all elements

robUx4:
according to https://github.com/Matroska-Org/ebml-specification/pull/83

So noone has to do this boring task later.


robUx4:
At some point we may flatten the XML. The wrapping information is a duplicate information of the `parent` value. The `level` value is probably not needed anymore.

Also the cppname should probably be an item like the `description`. So the `element` only has attributes defined in the EBML schema/RFC.

dericed:
The `<element>` nodes are now all at the same level. Also via this [xsl](https://gist.github.com/dericed/ddb0eaa6bb4f0983825e715d45ff6141) I converted the `parent` attribute to the new `path` attribute and dropped the redundant `level` attribute.

mbunkus:
I'm fine with merging this. Objections?

robUx4:
LGTM, I'll will try to recreate the table of elements on my side to confirm there's no inconsistencies. But that can be fixed later. We should not block this.

PR #54, opened 2016-10-15
Limit seekhead to 2 occurrences

dericed:
> 2 SeekHeads does not seem workable via the SeekHead definition at
> https://github.com/Matroska-Org/matroska-specification/blob/gh-pages/ord
> er_guidelines.md#seekhead


dericed:
https://github.com/Matroska-Org/matroska-specification/pull/42 should be handled first.

robUx4:
Fixed by https://github.com/Matroska-Org/matroska-specification/pull/62

PR #57, opened 2016-10-16
fix the recursive part of the path

robUx4:
subject to changed after https://github.com/Matroska-Org/ebml-specification/pull/113

It may become `1*(\Segment\Chapters\EditionEntry(1*(\ChapterAtom)))` which is less human readable but more accurate


retokromer:
LGTM

PR #62, opened 2016-10-23
Update order_guidelines.md

bastik-tor:
Grammar change.  Wording change to avoid possible misunderstanding.

### DIFF
--- a/order_guidelines.md
+++ b/order_guidelines.md
@@ -17,7 +17,7 @@ As noted by the EBML specification, if a `CRC-32` Element is used then the `CRC-
 
 ## SeekHead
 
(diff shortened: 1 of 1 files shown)


dericed:
I think the commit does clarify the language, but I notice that this section means something different because of a recent change in definition for `Top Level Element` at https://github.com/Matroska-Org/ebml-specification/commit/7b966f6efc31605a7aba61089c98647d085fdea0#diff-01f3b0f2fbf3e44327e6072f64ddc8a0R55.

Before the meaning of `Top Level Element` was:
> An `EBML Element` defined to only occur at `Level 1` within the `Root Element`.

and now it is:
> An `EBML Element` defined as a `Child Element` of the `Root Element`.

If a `Void Element` or `CRC-32 Element` are contained as a `Child Element` of the `Root Element` then in the first definition they are not considered `Top Level Elements` (since they can also occur at other levels), but in the 2nd definition the `Void Element` or `CRC-32 Element` are considered `Top Level Elements` and thus need to be referenced by `SeekHead` if they are used as a `Child Element` of the `Root Element`.

dericed:
I created another pull request in regards to the meaning of `Top Level Element` and don't consider this to block your PR which looks good to me.

PR #65, opened 2016-10-31
set unknownsizeallowed

dericed:
see
https://github.com/Matroska-Org/matroska-specification/issues/60#issueco
mment-257255578

Any others? Does mkvalidator test for the use of unknown sizes?

### DIFF
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <EBMLSchema docType="matroska" version="4">
-  <element name="Segment" path="1*1(\Segment)" id="0x18538067" type="master" minOccurs="1" maxOccurs="1" minver="1">
+  <element name="Segment" path="1*1(\Segment)" id="0x18538067" type="master" unknownsizeallowed="1" minOccurs="1" maxOccurs="1" minver="1">
     <documentation lang="en">The Root Element that contains all other Top-Level Elements (Elements defined only at Level 1). A Matroska file is composed of 1 Segment.</documentation>
   </element>
   <element name="SeekHead" path="0*2(\Segment\SeekHead)" cppname="SeekHeader" id="0x114D9B74" type="master" maxOccurs="2" minver="1">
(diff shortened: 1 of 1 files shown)


ebml_matroska.xml
```diff
@@ -520,7 +520,7 @@
   <element name="ContentSigHashAlgo" path="0*1(\Segment\Tracks\TrackEntry\ContentEncodings\ContentEncoding\ContentEncryption\ContentSigHashAlgo)" id="0x47E6" type="uinteger" minver="1" webm="0" default="0">
     <documentation lang="en">The hash algorithm used for the signature. A value of '0' means that the contents have not been signed but only encrypted. Predefined values:<br/> 1 - SHA1-160<br/> 2 - MD5</documentation>
   </element>
-  <element name="Cues" path="0*1(\Segment\Cues)" id="0x1C53BB6B" type="master" minver="1">
+  <element name="Cues" path="0*1(\Segment\Cues)" id="0x1C53BB6B" type="master" unknownsizeallowed="1" minver="1">
```
robUx4: Do we want Cues ? IMO it's only good as a temporary state and you end up with junk at the end, unless you Void the remaining part each time you write a new Cuepoint, which IMO doesn't exist anywhere.

robUx4:
mkvalidator allows unknown size on any `Master Element`. (for now)

ebml_matroska.xml
dericed: I don't have an opinion about `Cues`, but I could check my sample data to see where unknownsize Elements has been historically used.

robUx4: At first I thought it was a good idea to temporarily write data and a size when it's over. But with other changes on the unknownsize that would mean the Segment needs to be temporarily with an unknownsize. Also in that case Cues may have garbage during writing. So it's probably not a good move. 

If one wants to use unknown size during writing it's fine even on elements were it's not allowed. But proper values should be written when 'finalizing' the file.

mbunkus:

robUx4:
IMO the implications of having `Cues` with an unknown size are too big. If a reader can safely assume that the `Cues` are always with a known size, it can assume to always read it in one big chunk and map it to memory. Since a part of the file that's crucial at startup (usually players read ASAP so they are ready to seek when playback starts, which makes sense since we also have ordered chapters that can seek on startup) I think that's an important optimization we want to provide.

On the other hand the only case the Cues can have an unknown size is a temporary state. That might even be a good indicator if the Cues have an unknown size that the file failed to finish writing properly. 

So I think we should at least split this PR in 2. One with `Segment` and `Clusters` and one with `Cues`.

I'll do the split so we can move on with this PR.

Issue #46, opened 2016-09-24
cleanup table in codec specs


dericed:
The nest table in https://matroska.org/technical/specs/codecid/index.html is broken in the RFC output. The nested form, for instance under V_MPEG4/ISO/??? and V_REAL/????, is challenging to replicate in the simpler formatting of the RFC.
Can these tables be restructured without nesting?

Also the structure of the table in codec spec implies that only four pieces of information are needed to define a codec for use in Matroska:
- Codec ID
- Name
- Description
- Track Type (currently Video, Audio, Subtitle, and ... Button)

Is any more data needed to define an encoding for use in Matroska?

Also the TrackType Element also supports `complex`, `logo`, and `control` track types although none appear to be defined in the codec specs page.

dericed:
Although I pointed out the corresponding page on matroska.org, I should point out that I'm talking about https://github.com/Matroska-Org/matroska-specification/blob/gh-pages/codec_specs.md in the context of the repo.

ablwr:
This seems to no longer be an issue. Re-open if my assumption is false.

dericed:
+1

PR #66, opened 2016-11-02
Moves codec specs out of table

ablwr:
Hello! This pull request seeks, primarily, to break the approved codecs out of their occasionally-hierarchical table and into more simple charting for better viewing in Markdown, HTML, and (most importantly) RFC text. This pull request does NOT attempt to make assertions about the codecs or their descriptions, which will need to be reviewed by people with a stronger historical knowledge.

### DIFF
--- a/codec_specs.md
+++ b/codec_specs.md
@@ -188,33 +188,3 @@ For each line containing the timestamp and file position data is read from the a
 | Buttons |
 | B_VOBBTN | VobBtn Buttons | Based on [MPEG/VOB PCI packets](http://dvd.sourceforge.net/dvdinfo/pci_pkt.html). The file contains a header consisting of the string "butonDVD" followed by the width and height in pixels (16 bits integer each) and 4 reserved bytes. The rest is full [PCI packets](http://dvd.sourceforge.net/dvdinfo/pci_pkt.html). |
 
-
-To be supported later :
-
-'V_MSWMV'; Video, Microsoft Video
-
-'V_INDEO5'; Video, Indeo 5; transmuxed from AVI or created using VfW codec
-
-'V_MJPEG'; Video, MJpeg codec (lossy mode, general)
-
-'V_MJPEG2000'; Video, MJpeg 2000
-
-'V_MJPEG2000LL'; Video, MJpeg Lossless
-
-'V_DV'; Video, DV Video, type 1 (audio and video mixed)
-
-'V_TARKIN'; Video, Ogg Tarkin
-
-'V_ON2VP4'; Video, ON2, VP4
-
-'V_ON2VP5'; Video, ON2, VP5
-
-'V_3IVX'; Video, 3ivX (is D4 decoder downwards compatible?)
-
-'V_HUFFYUV'; Video, HuffYuv, lossless; auch als VfW möglich
-
-'V_COREYUV'; Video, CoreYuv, lossless; auch als VfW möglich
-
-'V_RUDUDU'; Nicola's Rududu Wavelet codec
-
-...... to be continued
--- a/codec_specs.md
+++ b/codec_specs.md
@@ -5,7 +5,7 @@ layout: default
 
 As an additional resource to this page Haali has created a [list of codec IDs in a PDF](http://haali.su/mkv/codecs.pdf).
 
-For each TrackEntry inside [matroska]({{site.baseurl}}/index.html), there has to be a [CodecID]({{site.baseurl}}/index.html#CodecID) defined. This ID is represent the codec used to encode data in the Track. The codec works with the coded data in the stream, but also with some codec initialisation. There are 2 different kind of codec "initialisation" :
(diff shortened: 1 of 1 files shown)


codec_specs.md
```diff
@@ -14,44 +14,99 @@ Each of these elements contain the same kind of data. And these data depend on t
...
+**Codec Name:** Microsoft (TM) Video Codec Manager (VCM)  
+**Description:** The private data contains the VCM structure BITMAPINFOHEADER including the extra private bytes, as [defined by Microsoft](http://msdn.microsoft.com/library/default.asp?url=/library/en-us/gdi/bitmaps_1rw2.asp). The data are stored in little endian format (like on IA32 machines). Where is the Huffman table stored in HuffYUV, not AVISTREAMINFO ??? And the FourCC, not in AVISTREAMINFO.fccHandler ???
+
+  
+**Codec ID:** V_UNCOMPRESSED  
+**Codec Name:** Video, raw uncompressed video frames  
+**Description:** The private data is void, all details about the used colour specs and bit depth are to be put/read from the KaxCodecColourSpace elements.
+
+  
+**Codec ID:** V_MPEG4/ISO/???
```
dericed: This `???` Codec ID and the Real/??? one below aren't real Codec IDs, but just like a Header for the Codec IDs that follow them. I think they could be removed.

codec_specs.md
```diff
@@ -14,44 +14,99 @@ Each of these elements contain the same kind of data. And these data depend on t
...
+
+  
+**Codec ID:** V_MPEG2  
+**Codec Name:** MPEG 2  
+**Description:** The matroska video stream will contain a demuxed Elementary Stream (ES ), where block boundaries are still to be defined. Its RECOMMENDED to use MPEG2MKV.exe for creating those files, and to compare the results with selfmade implementations
+
+  
+**Codec ID:** V_REAL/????  
+**Codec Name:** Real Video(TM)   
+**Description:** The stream is one of the Real Video(TM) video streams listed below. Source for the codec names are from [Karl Lillevold on Doom9](http://forum.doom9.org/showthread.php?s=&amp;threadid=55773&amp;perpage=20&amp;pagenumber=2#post331855). The CodecPrivate element contains a "real_video_props_t" structure in Big Endian byte order as found in [librmff](https://github.com/mbunkus/mkvtoolnix/blob/master/lib/librmff/librmff.h).
```
dericed: Link is dead. @robUx4, any other reference data for this?

codec_specs.md
```diff
@@ -14,44 +14,99 @@ Each of these elements contain the same kind of data. And these data depend on t
...
+
+  
+**Codec ID:** V_QUICKTIME  
+**Codec Name:** Video taken from QuickTime(TM) files  
+**Description:** Several codecs as stored in QuickTime, e.g. Sorenson or Cinepak. The CodecPrivate contains all additional data that is stored in the 'stsd' (sample description) atom in the QuickTime file **after** the mandatory video descriptor structure (starting with the size and FourCC fields). For an explanation of the QuickTime file format read [QuickTime File Format Specification](https://developer.apple.com/library/mac/documentation/QuickTime/QTFF/QTFFPreface/qtffPreface.html).
+
+  
+**Codec ID:** V_THEORA  
+**Codec Name:** Theora  
+**Description:** The private data contains the first three Theora packets in order. The lengths of the packets precedes them. The actual layout is:
```
dericed: Perhaps the term `private data` should be replaced with the more specific `CodecPrivate Element` throughout.

dericed:
I added some comments though they don't necessarily block the PR which is intended to be about formatting rather than content.

ablwr:
Linking to open issue: #46 

This can be cleaned up, but should we merge -- since this is about format rather than content -- and then seek to arrange content?

codec_specs.md
dericed: I think lines 33-37 could be removed.

codec_specs.md
dericed: Checking with @robUx4, but I think lines 68-72 could be removed and then the line:
> The CodecPrivate element contains a "real_video_props_t" structure in Big Endian byte order as found in [librmff](https://github.com/mbunkus/mkvtoolnix/blob/master/lib/librmff/librmff.h).

could be appended to all the V_REAL descriptions below on lines 75, 80, 85, and 90.

codec_specs.md
dericed: not blocking this PR, could resolve the expressions of CodecPrivate in a subsequent PR

dericed:
Replying to https://github.com/Matroska-Org/matroska-specification/pull/66#issuecomment-260223811, you're right I think the PR could be constrained to format and the content could be in a subsequent PR. I'm okay to support this being merged and can help follow with a content adjustment.

mbunkus:
I agree with separating formatting and content changes. I'll merge this as-is.

PR #67, opened 2016-11-03
removes overhead page, as well as removes from Makefile and sidebar

### LINKED ISSUE #48: fix issues in migration of overhead section
matroska.org included an article about [overhead](https://www.matroska.org/technical/overhead/index.html) as part of the spec, which in the repo is represented in markdown at [this page](https://github.com/Matroska-Org/matroska-specification/blob/gh-pages/overhead.md). Many of the tables in the markdown version are missing data (only the first column is migrated).

Relatedly, should this article be considered part of the format specification?

ablwr:
Seeks to resolve #48


dericed:
LGTM

dericed:
Still LGTM :)

PR #68, opened 2016-11-03
Removes EBML-specific information from Matroska specification

### LINKED ISSUE #50: remove material redundant to EBML spec
A few sections in Matroska's draft specification seem fully redundant to the EBML specification. EBML should be a normative reference of Matroska and the redundancy should be reviewed and likely reduced.

Particularly these sections:
https://github.com/Matroska-Org/matroska-specification/blob/gh-pages/notes.md#ebml-class
https://github.com/Matroska-Org/matroska-specification/blob/gh-pages/matroska_schema_section_header.md#ebml-schema-element-attributes

ablwr:
If already exists in EBML specification. Seeks to resolve #50, conversation there.


matroska_schema_section_header.md
```diff
@@ -2,39 +2,6 @@
 
 This specification includes an `EBML Schema` which defines the Elements and structure of Matroska as an EBML Document Type. The EBML Schema defines every valid Matroska element in a manner defined by the EBML specification.
 
-For convenience the section of the EBML specification that defines EBML Schema Element Attributes is restated here:
```
dericed: Seems fine to remove this section looks like an older snapshot of the attribute list from the ebml spec.

notes.md
```diff
@@ -44,11 +44,6 @@ The default value of an Element is assumed when not present in the data stream.
 
 Digital Rights Management. See [Encryption]({{site.baseurl}}/notes.html#Encryption).
 
-# EBML Class
-
-A larger EBML Class typically means the Element has a lower probability/importance. A larger Class-ID can be used as a synch word in case the file is damaged. Elements that are used frequently, but do not need to act as a synch word, SHOULD have a small Class-ID.
```
dericed: Is this point about length and sync in the ebml spec. It could be used in https://github.com/Matroska-Org/ebml-specification/blob/master/specification.markdown#element-id.

dericed: I found this in the ebml spec
> To reduce the risk of false positives while parsing EBML Streams, the IDs of the Root Element and Top-Level Elements SHOULD be at least 4 octets in length. Element IDs defined for use at Level 0 or Level 1 MAY use shorter octet lengths to facilitate padding and optimize edits to EBML Documents; for instance, the EBML Void Element uses an Element ID with a one octet length to allow its usage in more writing and editing scenarios.

which is similar enough to the above statement on Class IDs, so I'm ok with the cut.

dericed: Is it, so I'm ok with this.

mbunkus:
LGTM, too. Even more, I'm always in favor of removing duplication; that often leads to diverging information.

PR #69, opened 2016-11-05
Cues should include the last Cluster for easier seeking to the end

robUx4:
This is useful especially if the duration is missing for the Segment.

### DIFF
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -539,7 +539,7 @@
     <documentation lang="en">The hash algorithm used for the signature. A value of '0' means that the contents have not been signed but only encrypted. Predefined values:<br/> 1 - SHA1-160<br/> 2 - MD5</documentation>
   </element>
   <element name="Cues" path="0*1(\Segment\Cues)" id="0x1C53BB6B" type="master" maxOccurs="1" minver="1">
-    <documentation lang="en">A Top-Level Element to speed seeking access. All entries are local to the Segment. This Element SHOULD be mandatory for non <a href="http://www.matroska.org/technical/streaming/index.hmtl">"live" streams</a>.</documentation>
+    <documentation lang="en">A Top-Level Element to speed seeking access. All entries are local to the Segment. This Element SHOULD be mandatory for non <a href="http://www.matroska.org/technical/streaming/index.hmtl">"live" streams</a>. The last `Cluster` of the `Segment` SHOULD be included in this of `CuePoint`.</documentation>
   </element>
   <element name="CuePoint" path="1*(\Segment\Cues\CuePoint)" id="0xBB" type="master" minOccurs="1" minver="1">
     <documentation lang="en">Contains all information relative to a seek point in the Segment.</documentation>


dericed:
Is there a more specific recommendation or mandate on what should be included. From `diagram.md` it says
> For instance, you can index every single timecode of every Block, in every track if you liked, but you don't really need to. If you have a video file, you really just need to index the keyframes of the video track. 

Should all Clusters with keyframes be referenced in Cues? Should/must the first Cluster be referenced?

mbunkus:
I have my reservations about this change. I know that technically the `Cues` don't imply that the referenced frame is a key frame. However, players often make that assumption. Including the first frame of the last cluster if that cluster doesn't contain any key frame might be detrimental to playback.

dericed:
The Cues documentation currently doesn't infer much about what is good practice for using Cues. It seems that if Cues is used then the only mandate is to list at least one CuePoint (which could be any cluster, first, middle, or last). There's not yet any documentation about practices in using CuePoints for keyframes or not. This may be beyond the scope of this PR but I suggest we better detail:
- What is a minimal expectation of Cues?
- What is a best practice for use of Cues?
- When should Cues reference all Clusters?
- When is it better for Cues to only reference Clusters with keyframes?
- Are CuePoints required to be in any order? Putting them in order by timestamp or position seems logical, but is it recommended to do so?

Perhaps instead of detailing too much in the EBML Schema for Matroska, we could start a cues.md as we have for other Top-Level Elements.

mbunkus:
Tough but good questions.

> * What is a minimal expectation of Cues?

Good one. Unless you're creating a live stream I'd say that each file SHOULD contain cues in order to facilitate seeking. Files without cues are valid, though, therefore only SHOULD and not MUST.

If cues are present then they SHOULD either be located in front of the first cluster or they SHOULD be reachable via a meta seek element located before the first chapters. Again, if this is not the case then the file is still valid, hence no MUST, but a user can also not assume that a player will be able to use the cues.

> * What is a best practice for use of Cues?

Incomplete list from the top of my head:

* Always create cues unless in live streaming situations.
* Write meta seek elements before the first cluster and reference the cues from them.
* For each video track present: include each key frame in the cues.
* For each subtitle track present: include each frame in the cues and use CueDuration elements for them.
* If no video track is present, add cues for some but not all of the audio key frames (e.g. at most one every 500ms or some other interval that won't result in too much overhead).
* Only include key frames in cues in order not to have the player waste time.
* Write CueRelativePosition elements unless the referenced element is the first SimpleBlock or BlockGroup element within the cluster.
* If CodecState elements are used in a cluster, then always reference them from cues via the CueCodecState element.
* Sort cues by their CueTime element. However, the order of elements with the same CueTime doesn't matter (e.g. don't sort them at all, sort by CueTrack, whatever).

> * When should Cues reference all Clusters?

I don't see any situation where this is useful. Being able to quickly reach a cluster without being able to start decoding from it doesn't give the player much.

The one theoretical exception I can think of is a cluster that doesn't include a single key frame but where a CodecState element is used.

> * When is it better for Cues to only reference Clusters with keyframes?

Always, I'd say, as that's how current multiplexers work. Current players expect it that way.

> * Are CuePoints required to be in any order? Putting them in order by timestamp or position seems logical, but is it recommended to do so?

Similar as above: it's not required (we've never formalized it, therefore we shouldn't make this a requirement now in order not to invalidate existing files with potentially unordered enties), but multiplexers should sort them in order not to confuse players which expect them to be ordered due to historical use.

dericed:
A question based upon @mbunkus' answers.

> For each video track present: include each key frame in the cues.
> Only include key frames in cues in order not to have the player waste time.

Is this recommendation as true with an i-frame only video stream (i.e. all frames are keyframes)? Because later on the topic of have a Cue for ALL Clusters, you say:

> I don't see any situation where this is useful.

dericed:
@mbunkus, regarding:

> If cues are present then they SHOULD either be located in front of the first cluster or they SHOULD be reachable via a meta seek element located before the first chapters.

we already have a mandate in the [order guidelines](https://github.com/Matroska-Org/matroska-specification/blob/95534acaac0301ad8ff5ae5ee75df9e13bca6adf/order_guidelines.md#seekhead)

> Whether one or two SeekHead Element(s) are used, the SeekHead Element(s) MUST collectively reference the identity and position of all Top-Level Elements except for the first SeekHead itself.

So the second part of your test must already be true.

Also do you mean to reference `chapters` or do you mean `clusters`?

dericed:
regarding prior comment, I forgot that the SeekHead itself is not mandatory, so a recommendation that the `Cues` appear before the first `Cluster` is okay if there is no `SeekHead`.

Also maybe we need to change:

> Whether one or two SeekHead Element(s) are used, the SeekHead Element(s) MUST collectively reference the identity and position of all Top-Level Elements except for the first SeekHead itself.

to

> If one or two SeekHead Element(s) are used, the SeekHead Element(s) MUST collectively reference the identity and position of all Top-Level Elements except for the first SeekHead itself.

mbunkus:

dericed:
@robUx4, after https://github.com/Matroska-Org/matroska-specification/pull/74, do we still need this PR?

robUx4:
Closing this PR as this case is covered by the SHOULD suggestions in PR #74

PR #77, opened 2016-12-31
add codec specs for DVB subtitles (codec ID S_DVBSUB)

mbunkus:
This is the graphical subtitle format used by the Digital Video Broadcasting standard. There are several programs that have been writing Matroska files with such tracks for years. mkvmerge has only gained support for it after release 9.7.1.

This pull request makes that storage method official.

### DIFF
--- a/subtitles.md
+++ b/subtitles.md
@@ -523,3 +523,30 @@ When TextST subtitles are stored inside Matroska, the only allowed character set
 Each HDMV text subtitle stream in a Blu-ray can use one of a handful of character sets. This information is not stored in the MPEG2 Transport Stream itself but in the accompanying Clip Information file.
 
 Therefore a muxer MUST parse the accompanying Clip Information file. If the information indicates a character set other than UTF-8, it MUST re-encode all text Dialog Presentation Segments from the indicated character set to UTF-8 prior to storing them in Matroska.
+
+
+# Digital Video Broadcasting (DVB) subtitles
+
+The specifications for the Digital Video Broadcasting subtitle bitstream format (short: DVB subtitles) can be found in the document "ETSI EN 300 743 - Digital Video Broadcasting (DVB); Subtitling systems". The storage of DVB subtitles in MPEG transport streams is specified in the document "ETSI EN 300 468 - Digital Video Broadcasting (DVB); Specification for Service Information (SI) in DVB systems".
+
+## Storage of DVB subtitles
+
+### CodecID
+
+The CodecID to use is `S_DVBSUB`.
+
+### CodecPrivate
+
+The CodecPrivate element is five bytes long and has the following structure:
+
+*    2 bytes: composition page ID (bit string, left bit first)
+*    2 bytes: ancillary page ID (bit string, left bit first)
+*    1 byte: subtitling type (bit string, left bit first)
+
+The semantics of these bytes are the same as the ones described in section 6.2.41 "Subtitling descriptor" of ETSI EN 300 468.
+
+### Storage of DVB subtitles in Matroska Blocks
+
+Each Matroska Block consists of one or more DVB Subtitle Segments as described in segment 7.2 "Syntax and semantics of the subtitling segment" of ETSI EN 300 743.
+
+Each Matroska Block SHOULD have a Duration indicating how long the DVB Subtitle Segments in that Block SHOULD be displayed.
--- a/codec_specs.md
+++ b/codec_specs.md
@@ -358,6 +358,11 @@ Each frame is kept intact, including the CRC32. The header and seektable are dro
 **Description:** Basic image based subtitle format; The subtitles are stored as images, like in the DVD. The timestamp in the block header of matroska indicates the start display time, the duration is set with the Duration element. The full data for the subtitle bitmap is stored in the Block's data section.
 
   
+**Codec ID:** S_DVBSUB  
+**Codec Name:** Digital Video Broadcasting (DVB) subtitles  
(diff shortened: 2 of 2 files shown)


subtitles.md
```diff
@@ -523,3 +523,32 @@ When TextST subtitles are stored inside Matroska, the only allowed character set
...
+
+### CodecPrivate
+
+The CodecPrivate element is five bytes long and has the following structure:
+
+*    2 bytes: composition page ID (bit string, left bit first)
+*    2 bytes: ancillary page ID (bit string, left bit first)
+*    1 byte: subtitling type (bit string, left bit first)
+
+The semantics of these bytes bytes are the same as the ones described in section 6.2.41 "Subtitling descriptor" of ETSI EN 300 468.
```
dericed: `bytes bytes`

mbunkus: Meh :) Fixed.

subtitles.md
```diff
@@ -523,3 +523,32 @@ When TextST subtitles are stored inside Matroska, the only allowed character set
...
+
+### Storage of DVB subtitles in Matroska Blocks
+
+Each Matroska Block consists of one or more DVB Subtitle Segments as described in segment 7.2 "Syntax and semantics of the subtitling segment" of ETSI EN 300 743.
+
+DVB Subtitle Segments are normally shown until a subsequent Subtitle Segment is encountered. Therefore the Matroska Block MAY have no Duration. In that case a player MUST display Subtitle Segments within a Matroska Block until the next Subtitle Segments are encountered.
```
dericed: Re: `DVB Subtitle Segments are normally shown until a subsequent Subtitle Segment is encountered.` But if the Block has a duration then it should not be shown until the subsequent Subtitle Segment, right?

dericed: Also I think `SHOULD be displayed` may be better than `are normally shown`.

dericed: Re: `In that case a player MUST display Subtitle Segments`. This makes it seem as if the player does not have the option to not display the subtitle, such as when the subtitle is disabled.

mbunkus: Well, in the codec specs we always describe what a player must do in case it actually displays the track. I don't think we should litter everything with "but only if the track is enabled and selected for display" or something like that.

subtitles.md
```diff
@@ -523,3 +523,32 @@ When TextST subtitles are stored inside Matroska, the only allowed character set
...
+
+A multiplexer MAY use a Duration, e.g. by calculating the distance between two subsequent Blocks containing Subtitle Segments. If a Matroska Block has a Duration, a player MUST display that Segment only for the duration of the Block's Duration.
```
dericed: I think the logic of these last two paragraphs may be more clear if reordered like below.

> A multiplexer MAY use a Duration, e.g. by calculating the distance between two subsequent Blocks containing Subtitle Segments. If a Matroska Block has a Duration, a player MUST display that Segment only for the duration of the Block's Duration. If a Matroska Block has no Duration, a player MUST display Subtitle Segments within a Matroska Block until the next Subtitle Segments are encountered.

Also, I think `Segment` in this context should be avoided to prevent confusion with the `Segment Element`, using `Subtitle Segment` may be better. Also noting that somethings `Segment` or `Subtitle Segment` is plural and something singular, should be consistent.

subtitles.md
mbunkus: I can merge this and the following paragraph. How about…

> A Matroska Block MAY have a duration. If it does, a player MUST show the Segments for that Duration or until the next Block for the track is due to be shown, whichever occurs first. If the Block does not have a Duration, a player MUST show the Segments until the next block for the track is due to be shown.

subtitles.md
mbunkus: I'd rather not say anything about how a multiplexer calculates a duration. Let's only state how a demultiplexer/player must work. That's how most codec specs are written, too.

mbunkus:
I've just force-pushed modifications, especially for the handling of the display duration. Please re-review.

subtitles.md
```diff
@@ -523,3 +523,30 @@ When TextST subtitles are stored inside Matroska, the only allowed character set
...
+*    2 bytes: ancillary page ID (bit string, left bit first)
+*    1 byte: subtitling type (bit string, left bit first)
+
+The semantics of these bytes are the same as the ones described in section 6.2.41 "Subtitling descriptor" of ETSI EN 300 468.
+
+### Storage of DVB subtitles in Matroska Blocks
+
+Each Matroska Block consists of one or more DVB Subtitle Segments as described in segment 7.2 "Syntax and semantics of the subtitling segment" of ETSI EN 300 743.
+
+A Matroska Block MAY have a Duration. If a Matroska Block has a Duration, a player MUST display that Segment only for the duration of the Block's Duration. If a Matroska Block has no Duration, a player MUST display Subtitle Segments within a Matroska Block until the next Subtitle Segments are encountered.
```
robUx4: Not sure the duration explanation belongs. It's probably how the subtitle (and video) duration should always be handled, as opposed to audio duration for example.

robUx4: On the other it should say that the DVB duration should be written in the Matroska `Block`.

mbunkus: Yeah, you're probably right. How long a packet should be displayed belongs to the general Matroska specs, not codec-spcific description.

mbunkus:
I've removed the section about how to determine how long to display the segments and replaced it with a note stating that each Matroska Block SHOULD have a Duration. Result's been force-pushed.

PR #79, opened 2016-12-31
defining codec specs

dericed:
Starting a process to better define how to define codec specs. This starts to detail how to create a Codec ID. And splits the concept of Description into Description, Citation, and Initialisation.

### DIFF
--- a/codec_specs.md
+++ b/codec_specs.md
@@ -1,21 +1,61 @@
 ---
 layout: default
 ---
-# CodecID
+# Codec Mappings
 
-As an additional resource to this page Haali has created a [list of codec IDs in a PDF](http://haali.su/mkv/codecs.pdf).
+A `Codec Mapping` is a set of attributes to identify, name, and contextualize the format and characteristics of encoded data that can be contained within Matroska Clusters.
 
-For each TrackEntry inside [matroska]({{site.baseurl}}/index.html), there has to be a [CodecID]({{site.baseurl}}/index.html#CodecID) defined. This ID is represent the codec used to encode data in the Track. The codec works with the coded data in the stream, but also with some codec initialisation. There are 2 different kind of codec "initialisation":
+Each TrackEntry used within Matroska MUST reference a defined `Codec Mapping` using the `Codec ID` to identify and describe the format of the encoded data in its associated Clusters. This `Codec ID` is a unique registered identifier that represents the encoding stored within the Track. Certain encodings MAY also require some form of codec initialisation in order to provide its decoder with context and technical metadata.
 
-*   CodecPrivate in the TrackEntry
-*   CodecState in the BlockGroup
+The intention behind this list is not to list all existing audio and video codecs, but rather to list those codecs that are currently supported in Matroska and therefore need a well defined `Codec ID` so that all developers supporting Matroska will use the same `Codec ID`. If you feel we missed support for a very important codec, please tell us on our development mailing list (cellar at ietf.org).
 
-Each of these elements contain the same kind of data. And these data depend on the codec used.
+## Defining Matroska Codec Support
 
-Important Note:
+Support for a codec is defined in Matroska with the following values.
 
-The intention behind this list is `NOT` to list all existing audio and video codecs, but rather to list those codecs that are `currently supported` in Matroska and therefore need a well defined codec ID so that all developers supporting Matroska will use the same ID. If you feel we missed support for a very important codec, please tell us on our development mailing list (matroska-devel at lists.matroska.org).
+### Codec ID
 
+Each codec supported for storage in Matroska MUST have a unique `Codec ID`. Each `Codec ID` MUST be prefixed with the string from the following table according to the associated type of the codec. All characters of a `Codec ID Prefix` MUST be capital letters (A-Z) except for the last character of a `Codec ID Prefix` which MUST be an underscore ("_").
+
+Codec Type | Codec ID Prefix
+-----------|----------------
+Video      | "V_"
+Audio      | "A_"
+Subtitle   | "S_"
+Button     | "B_"
+
+Each `Codec ID` MUST include a `Major Codec ID` immediately following the `Codec ID Prefix`. A `Major Codec ID` MAY be followed by an OPTIONAL `Codec ID Suffix` to communicate a refinement of the `Major Codec ID`. If a `Codec ID Suffix` is used, then the `Codec ID` MUST include a forward slash ("/") as a separator between the `Major Codec ID` and the `Codec ID Suffix`. The `Major Codec ID` MUST be composed of only capital letters (A-Z) and numbers (0-9). The `Codec ID Suffix` MUST be composed of only capital letters (A-Z), numbers (0-9), and formard slash ("/").
+
+The following table provides examples of valid `Codec IDs` and their components:
+
+Codec ID Prefix | Major Codec ID | Separator | Codec ID Suffix | Codec ID
+A_              | AAC            | /         | MPEG2/LC/SBR    | A_AAC/MPEG2/LC/SBR
+V_              | V_MPEG4        | /         | ISO/ASP         | V_MPEG4/ISO/ASP
+V_              | MPEG1          |           |                 | V_MPEG1
+
+### Codec Name
+
+Each encoding supported for storage in Matroska MUST have a Codec Name. The Codec Name provides a readable label for the encoding.
+
+### Description
+
+An optional description for the encoding. This value is only intended for human consumption.
+
+### Initialisation
+
+Each encoding supported for storage in Matroska MUST have a defined Initialisation. The Initialisation MUST describe the storage of data necessary to initialize the decoder, which MUST be stored within the `CodecPrivate Element`. When the Initialisation is updated within a track then that updated Initialisation data MUST be written into the `CodecState Element` of the first `Cluster` to require it. If the encoding does not require any form of Initialisation then `none` MUST be used to define the Initialisation and the `CodecPrivate Element` SHOULD NOT be written and MUST be ignored.
+
+### Citation
+
+Documentation of the associated normative and informative references for the codec is RECOMMENDED.
+
+### Deprecation Date
+
+A timestamp, expressed in [@!RFC3339] that notes when support for the `Codec Mapping` within Matroska was deprecated. If a `Codec Mapping` is defined with a `Deprecation Date`, then it is RECOMMENDED that Matroska writers SHOULD NOT use the `Codec Mapping` after the `Deprecation Date`.
+
+### Superseded By
+
+A `Codec Mapping` MAY only be defined with a `Superseded By` value, if it has an expressed `Deprecation Date`. If used, the `Superseded By` value MUST store the `Codec ID` of another `Codec Mapping` that has superseded the `Codec Mapping`.
 
 ## Video Codecs
 


codec_specs.md
```diff
@@ -1,21 +1,59 @@
...
-*   CodecPrivate in the TrackEntry
-*   CodecState in the BlockGroup
-
-Each of these elements contain the same kind of data. And these data depend on the codec used.
+Each of these Elements contain the same kind of data. And these data depend on the codec used.
 
 Important Note:
 
-The intention behind this list is `NOT` to list all existing audio and video codecs, but rather to list those codecs that are `currently supported` in Matroska and therefore need a well defined codec ID so that all developers supporting Matroska will use the same ID. If you feel we missed support for a very important codec, please tell us on our development mailing list (matroska-devel at lists.matroska.org).
+The intention behind this list is `NOT` to list all existing audio and video codecs, but rather to list those codecs that are `currently supported` in Matroska and therefore need a well defined codec ID so that all developers supporting Matroska will use the same ID. If you feel we missed support for a very important codec, please tell us on our development mailing list (cellar at ietf.org).
```
retokromer: Why `NOT`? I don't think this in conforming RFC 2119.

codec_specs.md
```diff
@@ -1,21 +1,59 @@
...
+
+Each codec supported for storage in Matroska MUST have a Codec Name. The Codec Name provides a readable label for the codec.
+
+### Description
+
+An optional description for the codec.
+
+### Initialisation
+
+Each codec supported for storage in Matroska MUST have a defined Initialisation. If the codec does not require any form of Initialisation then `none` SHOULD be used to define the Initialisation. The Initialisation MUST document in detail how data MUST be stored within the CodecPrivate and CodecState Elements in order to initialize the codec.
```
retokromer: «The Initialisation MUST document in detail how data MUST be stored» looks very weird to me. I suggest to replace the second MUST. – Yet this may be only my weird taste.

codec_specs.md
mbunkus: I agree with @retokromer. Let's correct it now: lower-case "not" without backticks. Additionally I'd remove the backticks around "currently supported". It's supposed to be an emphasis, I guess, but we usually use backticks for official terminology like element names.

codec_specs.md
```diff
@@ -14,6 +14,46 @@ Important Note:
...
+Each codec supported for storage in Matroska MUST have a unique Codec ID. Each Codec ID MUST be prefixed with a two character string according to the associated type of the codec as listed in the following table.
+
+Codec Type | Codec ID Prefix
+-----------|----------------
+Video      | V_
+Audio      | A_
+Subtitle   | S_
+Button     | B_
+
+Each Codec ID MUST then use a `Major Codec ID` and may optionally use a `Minor Codec ID` and `Micro Codec ID` with a forward-slash (`/`) delimiter between each component.
```
mbunkus: Unfortunately this clashes a bit with existing practice from way back when Matroska was very young. We used the following monstrosities for AAC:

* `A_AAC/MPEG2/LC`
* `A_AAC/MPEG2/MAIN`
* `A_AAC/MPEG2/LC/SBR`
* `A_AAC/MPEG2/SSR`
* `A_AAC/MPEG4/LC`
* `A_AAC/MPEG4/LTP`
* `A_AAC/MPEG4/MAIN`
* `A_AAC/MPEG4/LC/SBR`
* `A_AAC/MPEG4/SSR`

As you can see some of them have one more layer. No `CodecPrivate` was used with them.

However, those IDs have been discontinued in favor of `A_AAC` with `CodecPrivate` 2005-12-07 (mkvmerge 1.6.5).

All other codec IDs that mkvmerge handles keep to major/minor/micro.

Looking forward I'm in favor of restricting official IDs to major/minor/micro.

codec_specs.md
mbunkus: @retokromer well, both MUSTs refer to different things: the first is a restriction on how to write the specs, the second a restriction on how multiplexers write the file. I'd rather keep both.

retokromer: Exactly! I agree with you, @mbunkus, and for this very reason I would personally prefer to divide into two sentences these two behaviours.

mbunkus: I'd be fine with that, of course.

dericed: I removed the second MUST. I think that the Initialization statement for each codec itself should clarify the level of mandate.

codec_specs.md
dericed: In some versioning systems, the fourth value is called `qualifier`. I suppose this term could fit here.

Also from codec_specs.md and https://matroska.org/technical/specs/codecid/index.html it doesn't seem possible to `know` that these Codec IDs are deprecated or what they are superseded by. Are these only discontinued by mkvmerge or by the Matroska format?

Should we have two other descriptors:
- Deprecation Date: optional but gives a date when the Matroska specification maintain deprecates a particular Codec ID.
- Superseded By: if Deprecation Date is used, then it should be highly recommended to say what Codec ID has superseded the deprecated one.

For players, they could then use the full registry for playback support, though writers should avoid using deprecated Codec IDs.

mbunkus: Ugh. We've never deprecated the IDs officially, nor have we officially added the ID that's actually used — `A_AAC`. I clearly remember there being a discussion and consensus either on the mailing list or on the IRC channel about this change.

I don't know of a single Matroska muxer that still writes those old IDs.

I'd consider them de facto deprecated, with the specs page needing to catch up.

I like both of the descriptors you've proposed.

dericed:
Forced in a signification update. This describes a term `Codec Mapping` to describe the information required to define encoding support in Matroska. I updated the description of `Codec ID` and `Initialization` and added `Adoption Date`, `Deprecation Date`, and `Superseded By`.

Also what does `The private data is void` mean in the old specification document. For instance if using `A_MPEG/L3` then `the private data is void`. I'm confused by the use of the term `void` here. Does this mean that?
- a `CodecPrivate Element` MUST be present and be an `Empty Element`
- there MUST be no written `CodecPrivate Element`

Also I could use help describing when `CodecState Elements` should be used.

mbunkus:
Thanks. I'll do a re-review later or tomorrow.

> I'm confused by the use of the term void here.

Yeah :) It means the latter: the codec doesn't require private initialization data, and therefore a `CodecPrivate Element` MUST NOT exist. Technically an empty `CodecPrivate Element` would probably work just as well, however, there's no upside to allowing it apart from confusing players and trying to trigger bugs; therefore I'm in favor of disallowing that.

codec_specs.md
```diff
@@ -1,21 +1,67 @@
...
+
+An optional description for the encoding.
+
+### Initialisation
+
+Each encoding supported for storage in Matroska MUST have a defined Initialisation. The Initialisation MUST describe the storage of data necessary to initialize the decoder, which MUST be stored within the `CodecPrivate Element`. When the Initialisation is updated within a track then that updated Initialisation data MUST be written into the `CodecState Element` of the first `Cluster` to require it. If the encoding does not require any form of Initialisation then `none` MUST be used to define the Initialisation and the `CodecPrivate Element` MUST be void.
+
+### Citation
+
+Documentation of the associated normative and informative referneces for the codec is RECOMMENDED.
```
JeromeMartinez: references

codec_specs.md
```diff
@@ -1,21 +1,67 @@
...
+
+### Adoption Date
```
JeromeMartinez: I don't think it is easy to say the adoption date because it would be actually when the RFC is accepted, and it would be redundant with the RFC release date.
I suggest to remove this part.

codec_specs.md
JeromeMartinez: I think the proposal is too limitative, and we should let each ID free to define what is after "/".
I suggest:
"Codec ID Prefix"_"Major Codec ID"/ "Codec ID Specific"
(the "/" + right part being optional)

dericed: I agree with Jerome's suggestion, I rewrote to use `Codec ID Prefix`, `Major Codec ID`, and `Codec ID Suffix`.

codec_specs.md
dericed: fixed locally

codec_specs.md
dericed: removed Adoption Date

dericed:
Updated according to some suggestions from @JeromeMartinez and @mbunkus. Please re-review.

codec_specs.md
```diff
@@ -1,21 +1,61 @@
...
+Each TrackEntry used within Matroska MUST reference a defined `Codec Mapping` using the `Codec ID` to identify and describe the format of the encoded data in its associated Clusters. This `Codec ID` is a unique registered identifier that represents the encoding stored within the Track. Certain encodings MAY also require some form of codec initialisation in order to provide its decoder with context and technical metadata.
 
-*   CodecPrivate in the TrackEntry
-*   CodecState in the BlockGroup
+The intention behind this list is not to list all existing audio and video codecs, but rather to list those codecs that are currently supported in Matroska and therefore need a well defined Codec ID so that all developers supporting Matroska will use the same Codec ID. If you feel we missed support for a very important codec, please tell us on our development mailing list (cellar at ietf.org).
```
retokromer: twice \`Codec ID\`

dericed: updated. thx

codec_specs.md
```diff
@@ -1,21 +1,61 @@
...
-Each of these elements contain the same kind of data. And these data depend on the codec used.
+## Defining Matroska Codec Support
 
-Important Note:
+Support for a codec is defined in Matroska with the following values.
 
-The intention behind this list is `NOT` to list all existing audio and video codecs, but rather to list those codecs that are `currently supported` in Matroska and therefore need a well defined codec ID so that all developers supporting Matroska will use the same ID. If you feel we missed support for a very important codec, please tell us on our development mailing list (matroska-devel at lists.matroska.org).
+### Codec ID
 
+Each codec supported for storage in Matroska MUST have a unique `Codec ID`. Each `Codec ID` MUST be prefixed with the string from the following table according to the associated type of the codec. All characters of a `Codec ID Prefix` MUST be capital letters (A-Z) except for the last character of a `Codec ID Prefix` which MUST be a hypen ("-").
```
mbunkus: Actually, the last character is not a hyphen but an underscore :)

codec_specs.md
```diff
@@ -1,21 +1,61 @@
...
+
+Each encoding supported for storage in Matroska MUST have a Codec Name. The Codec Name provides a readable label for the encoding.
+
+### Description
+
+An optional description for the encoding.
+
+### Initialisation
+
+Each encoding supported for storage in Matroska MUST have a defined Initialisation. The Initialisation MUST describe the storage of data necessary to initialize the decoder, which MUST be stored within the `CodecPrivate Element`. When the Initialisation is updated within a track then that updated Initialisation data MUST be written into the `CodecState Element` of the first `Cluster` to require it. If the encoding does not require any form of Initialisation then `none` MUST be used to define the Initialisation and the `CodecPrivate Element` MUST be void.
```
mbunkus: I don't like the wording that an element "must be void". I'd prefer something like "must not be used" or "must be ignored" or something like that.

And there's a missing , before "then" in the last sentence.

codec_specs.md
```diff
@@ -1,21 +1,61 @@
...
+V_              | V_MPEG4        | /         | ISO/ASP         | V_MPEG4/ISO/ASP
+V_              | MPEG1          |           |                 | V_MPEG1
+
+### Codec Name
+
+Each encoding supported for storage in Matroska MUST have a Codec Name. The Codec Name provides a readable label for the encoding.
+
+### Description
+
+An optional description for the encoding.
```
mbunkus: Maybe we should clarify that `Codec Name` and `Description` are meant solely for human consumption and have no effect on the processing of Matroska data by a demuxer.

codec_specs.md
```diff
@@ -1,21 +1,61 @@
...
+
+Each encoding supported for storage in Matroska MUST have a defined Initialisation. The Initialisation MUST describe the storage of data necessary to initialize the decoder, which MUST be stored within the `CodecPrivate Element`. When the Initialisation is updated within a track then that updated Initialisation data MUST be written into the `CodecState Element` of the first `Cluster` to require it. If the encoding does not require any form of Initialisation then `none` MUST be used to define the Initialisation and the `CodecPrivate Element` MUST be void.
+
+### Citation
+
+Documentation of the associated normative and informative references for the codec is RECOMMENDED.
+
+### Deprecation Date
+
+A timestamp, expressed in [@!RFC3339] that notes when support for the `Codec Mapping` within Matroska was deprecated. If a Codec is defined with a `Deprecation Date`, then it is RECOMMENDED that Matroska writers SHOULD NOT use the `Codec Mapping` after the `Deprecation Date`.
```
mbunkus: "Codec Mapping" instead of "Codec"?

codec_specs.md
retokromer: I agree regarding «void».

codec_specs.md
dericed: I used `Codec Mapping` rather than `Codec` because I didn't want saying that `A_AAC/MPEG2/LC/SBR` was deprecated could imply that `AAC` as an encoding format was deprecated. Since it is just that particular mapping of the codec to Matroska was is deprecated and not the codec itself.

codec_specs.md
dericed: Added:
> This value is only intended for human consumption.

codec_specs.md
dericed: argh, right. Fixed locally.

codec_specs.md
dericed: I agree but want the statement to affect the writer as well as the reader. I updated to:
> ...  and the `CodecPrivate Element` SHOULD NOT be written and MUST be ignored.

dericed:
PR is adjusted to feedback and force pushed.

codec_specs.md
mbunkus: Sorry for explaining so poorly what I meant. I'm fine with `Codec Mapping`. However, you've written: `If a Codec is defined…`, and my point was that this should probably be `If a Codec Mapping is defined…` in order to be consistent.

I did not mean to advocate changing `Codec Mapping` back to `Codec`.

dericed: thx for clarifying, fixed and pushed

dericed:
Once again, ready for review.

codec_specs.md
```diff
@@ -1,21 +1,61 @@
 ---
 layout: default
 ---
-# CodecID
+# Codec Mappings
 
-As an additional resource to this page Haali has created a [list of codec IDs in a PDF](http://haali.su/mkv/codecs.pdf).
+A `Codec Mapping` is a set of attributes to identify, name, and contextualize the format and characteristics of encoded data that can be contained within Matroska Clusters.
```
robUx4: It's actually in the `Blocks` not the `Clusters` which contain more generic stuff.

robUx4: More generally I think it should say that the codec data correspond to the Payload data found in `Block` and `SimpleBlock`,

PR #81, opened 2017-01-15
Cues unknownsize

robUx4:
I am personally against it for reasons given in PR #65 

Especially this comment https://github.com/Matroska-Org/matroska-specification/pull/65#issuecomment-272684623

### DIFF
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -538,7 +538,7 @@
   <element name="ContentSigHashAlgo" path="0*1(\Segment\Tracks\TrackEntry\ContentEncodings\ContentEncoding\ContentEncryption\ContentSigHashAlgo)" id="0x47E6" type="uinteger" maxOccurs="1" minver="1" webm="0" default="0">
     <documentation lang="en">The hash algorithm used for the signature. A value of '0' means that the contents have not been signed but only encrypted. Predefined values:<br/> 1 - SHA1-160<br/> 2 - MD5</documentation>
   </element>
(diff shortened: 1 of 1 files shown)


dericed:
+0

robUx4:
I suppose I agree with the idea of dropping this PR ?
@mbunkus any opinion as you rely on temporary writing of Cues in mkvmerge ?

mbunkus:
I actually do not rely on it. Cues are built in memory and only written en bloc when the file is finalized. The only element mkvmerge ever sets an unknown on is the Segment itself.

Personally I don't see a use-case for writing Cues with an unknown size. If the size isn't known, then a writer intends to write more clusters. However, in that case it wouldn't have writen the Cues yet anyway. For example, the following temporary layout during writing doesn't make much sense:

```
Segment(unknown size) → Cues(unknown size) Cluster1 Cluster2 Cluster3…
```

Here the writer must either have reserved some space for the Cues (e.g. via a following Void element), in which case the current size is known and could be written, or it has not reserved more space, in which case the size is definitely known.

On the other hand if the Cues are written at the end, then the size is almost certainly known.

Addressing Steves comment about readers reading the Cues as one big chunk: I don't know a Matroska parser that actually works that way. Which doesn't mean that there aren't any, of course, but it's more of a theoretical concern in my eyes.

In the end I don't have any strong feelings one way or the other. I lean in the direction of not allowing unknown-sized Cues.

retokromer:
> In the end I don't have any strong feelings one way or the other. I lean in the direction of not allowing unknown-sized Cues.

+1

robUx4:
Closing. Feel free to reopen if needed.

PR #82, opened 2017-01-15
remove docs on EncryptedBlock, VirtualBlock

dericed:


dericed:
As discussed at https://mailarchive.ietf.org/arch/msg/cellar/YYloMpauJJCRftQsUQvm0BFoi4M

PR #84, opened 2017-01-20
move example subtitles to codeblocks

dericed:


dericed:
I think beyond formatting this section needs a lot of work. Some suggestions:

- reduce the size of some of the examples in code blocks, some seem unnecessarily long
- use RFC2119 language
- the initial paragraphs are a bit casual and should be rewritten to better contextualize the section as recommendations and mandates for subtitle use
- add references

PR #85, opened 2017-01-20
fix a codec_spec table

dericed:


mbunkus:
That one's simple enough to merge directly :)

PR #86, opened 2017-01-22
update the TrackTimecodeScale min and max spec value to match the specs

robUx4:
It was incorrectly set to 0/0 in commit d649f83355c956cd61f321ca06884ed842161300.


dericed:
LGTM

PR #87, opened 2017-01-22
XML Schema: add a cppname to projection elements

robUx4:
In libmatroska all sub video elements have the Video prefix in their name.


robUx4:
@mbunkus do you use the elements already in mkvtoolnix ? Otherwise we can just use the regular names generated for these classes.

robUx4:
Given libmatroska was not released yet with these elements I would guess you don't.

dericed:
LGTM

PR #95, opened 2017-02-19
Clean Makefile

robUx4:
Using make -j2 will also build the .txt and .html outputs at the same time in parallel
Update the XSLT script will also trigger rebuilding the proper files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 _site
 .sass-cache
 *lhomme-cellar-matroska-*
-merged.md
 ebml_matroska_elements.md
 ebml_matroska_elements4rfc.md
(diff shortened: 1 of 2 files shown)


dericed:
I tested this and it works well, produces the same output, but is far better written. However, when running `make` there is now no `draft-lhomme-cellar-matroska-01.xml` output and this is the file which is submitted to the ietf document tracker. Can you add `draft-lhomme-cellar-matroska-01.xml` back? Otherwise, LGTM.

robUx4:
I put the regular expression change first, as otherwise none of these changes build in the end for me. It's nicer to always have something that builds first.

I also added a commit to keep the XML file.

dericed:
LGTM

retokromer:
LGTM

robUx4:
Waiting on #97 before merging this.

PR #97, opened 2017-02-19
Mmark clean

robUx4:
On top of the Makefile cleaning branch

The `layout:` lines are not cleaned properly for me and then mmark produces junk output.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,13 +4,13 @@ all: draft-lhomme-cellar-matroska-01.html draft-lhomme-cellar-matroska-01.txt
 
 draft-lhomme-cellar-matroska-01.html: index.md
 	xsltproc transforms/ebml_schema2markdown4rfc.xsl ebml_matroska.xml > ebml_matroska_elements4rfc.md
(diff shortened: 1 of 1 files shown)


retokromer:
On Slackware 14.1 as well: https://github.com/Matroska-Org/matroska-specification/pull/97/commits/b230559e4d37765ce883a564e0ea0e006f8c71c6

LGTM

mbunkus:
Hmm, this PR includes the changes from #95, doesn't it? Makes it a bit harder to review what changed in this particular PR only.

robUx4:
Yes, that's only the last commit. It depends on the other commits being merged. If they're not I will make a different version. But it won't merge/rebase cleanly into branch #95.

robUx4:
Don't we want to cover just this problematic section that is (oddly) generated ?

dericed:
LGTM

mbunkus:
Like I said: `grep -v '^---\|^layout:\|^---\b'` is equivalent to `grep -v '^---\|^layout:'` (as `^---` always matches where `^---\b` matches). As I prefer clarity, I'd ask you to change this to `grep -v '^---\|^layout:'`.

robUx4:
I updated the commit to match this.

PR #99, opened 2017-02-22
Chapters

dericed:
related to
https://github.com/Matroska-Org/matroska-specification/issues/98

### DIFF
--- a/chapters.md
+++ b/chapters.md
@@ -93,8 +93,10 @@ In this example a movie is split in different chapters. It could also just be an
 This would translate in the following matroska form :
 
 ```xml
+<?xml version="1.0"?>
 <Chapters>
   <EditionEntry>
+    <EditionUID>16603393396715046047</EditionUID>
     <ChapterAtom>
       <ChapterUID>1193046</ChapterUID>
       <ChapterTimeStart>00:00:00.000000000</ChapterTimeStart>
@@ -103,6 +105,8 @@ This would translate in the following matroska form :
         <ChapterString>Intro</ChapterString>
         <ChapterLanguage>eng</ChapterLanguage>
       </ChapterDisplay>
+      <ChapterFlagHidden>0</ChapterFlagHidden>
+      <ChapterFlagEnabled>1</ChapterFlagEnabled>
     </ChapterAtom>
     <ChapterAtom>
       <ChapterUID>2311527</ChapterUID>
@@ -116,6 +120,8 @@ This would translate in the following matroska form :
         <ChapterString>Avant le crime</ChapterString>
         <ChapterLanguage>fra</ChapterLanguage>
       </ChapterDisplay>
+      <ChapterFlagHidden>0</ChapterFlagHidden>
+      <ChapterFlagEnabled>1</ChapterFlagEnabled>
     </ChapterAtom>
     <ChapterAtom>
       <ChapterUID>3430008</ChapterUID>
@@ -129,6 +135,8 @@ This would translate in the following matroska form :
         <ChapterString>Le crime</ChapterString>
         <ChapterLanguage>fra</ChapterLanguage>
       </ChapterDisplay>
+      <ChapterFlagHidden>0</ChapterFlagHidden>
+      <ChapterFlagEnabled>1</ChapterFlagEnabled>
     </ChapterAtom>
     <ChapterAtom>
       <ChapterUID>4548489</ChapterUID>
@@ -142,6 +150,8 @@ This would translate in the following matroska form :
         <ChapterString>Après le crime</ChapterString>
         <ChapterLanguage>fra</ChapterLanguage>
       </ChapterDisplay>
+      <ChapterFlagHidden>0</ChapterFlagHidden>
+      <ChapterFlagEnabled>1</ChapterFlagEnabled>
     </ChapterAtom>
     <ChapterAtom>
       <ChapterUID>5666960</ChapterUID>
@@ -155,7 +165,11 @@ This would translate in the following matroska form :
         <ChapterString>Générique</ChapterString>
         <ChapterLanguage>fra</ChapterLanguage>
       </ChapterDisplay>
+      <ChapterFlagHidden>0</ChapterFlagHidden>
+      <ChapterFlagEnabled>1</ChapterFlagEnabled>
     </ChapterAtom>
+    <EditionFlagDefault>0</EditionFlagDefault>
+    <EditionFlagHidden>0</EditionFlagHidden>
   </EditionEntry>
 </Chapters>
 ```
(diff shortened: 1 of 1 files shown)


dericed:
This is some prep work on chapters.md to prepare for implementing some of hubblec4's recommendations on the cellar listserv. It tries to fix some logical errors in the Chapter Flag section. Please review carefully that both Hidden and Enabled follow the logic in the rewritten Chapter Flag section.

chapters.md
```diff
@@ -222,4 +78,188 @@ The next 4 following octets correspond to the [User Operation flags](http://dvd.
...
+*   27500ms - 38000ms : The killer arrested
+*   38000ms - 43000ms : Credits
+
+This would translate in the following matroska form :
+
+```xml
+<Chapters>
+  <EditionEntry>
+    <ChapterAtom>
+      <ChapterUID format="hex">12 34 56</ChapterUID>
```
hubblec4: ChapterUID in hex-format, is this correct?

dericed: Hmm, ChapterUID is defined as an unsigned integer, so I should change that. I used the practice of `format=hex` from @mbunkus's examples at https://github.com/mbunkus/mkvtoolnix/blob/029a6bf943cc9e4262da80d0851fcca574fceb2e/examples/example-segmentinfo-1.xml; however I neglected that some `UID` Elements are binary and some are unsigned integer.

dericed:
I updated the PR with new commits. Now the `ChapterUID` elements use unsigned integers (thx @hubblec4). Also the `ChapterTime*` expressions were converted from unsigned integers to time expressions (see comments at https://github.com/Matroska-Org/ebml-specification/issues/132). The XML fragments will now work with mkvpropedit.

Please review.

chapters.md
```diff
@@ -222,4 +78,236 @@ The next 4 following octets correspond to the [User Operation flags](http://dvd.
...
+        <ChapterLanguage>eng</ChapterLanguage>
+      </ChapterDisplay>
+      <ChapterFlagHidden>0</ChapterFlagHidden>
+      <ChapterFlagEnabled>1</ChapterFlagEnabled>
+    </ChapterAtom>
+    <EditionFlagDefault>0</EditionFlagDefault>
+    <EditionFlagHidden>0</EditionFlagHidden>
+  </EditionEntry>
+</Chapters>
+```
```
hubblec4: Minimum chapter.xml sample

I  have noticed that some program's output of a chapter.xml is not Matroska conform. 
(Missing UID's or Flags)
As I started my capterEditor project, I  would have been glad to find such info.

Here is a minium sample in xml:
<Chapters>
  <EditionEntry>
    <EditionFlagHidden>0</EditionFlagHidden>
    <EditionFlagDefault>0</EditionFlagDefault>
    <ChapterAtom>
      <ChapterUID>1</ChapterUID>
      <ChapterTimeStart>00:00:00.000000000</ChapterTimeStart>
      <ChapterFlagHidden>0</ChapterFlagHidden>
      <ChapterFlagEnabled>1</ChapterFlagEnabled>
    </ChapterAtom>
  </EditionEntry>
</Chapters>

And then we could have an "absolute minimum", because the Flags(Hidden, Enabled) can use the default values. 
<Chapters>
  <EditionEntry>
    <ChapterAtom>
      <ChapterUID>1</ChapterUID>
      <ChapterTimeStart>00:00:00.000000000</ChapterTimeStart>
    </ChapterAtom>
  </EditionEntry>
</Chapters>

mbunkus:
About the XML representation of chapters. There's always some confusion about it. Originally the XML format was invented by me for use in MKVToolNix. It was modeled closely after the element names used in the Matroska specification, but it doesn't quite match them in some cases. The same is true for tags.

To make this clear: there's no official XML representation for chapter and tag information from the spec's side, and the formats that mkvmerge uses aren't official either.

For example:

* mkvmerge will automatically add random UIDs if the `EditionUID` / `ChapterUID` elements are absent. Matroska requires their presence, though.
* mkvmerge only supports decimal number formats for signed/unsigned integer elements. In the spec one could easily write them as hexadecimal numbers, though.
* Some of the element names mkvmerge uses don't correspond to the official element names in the spec but rather the C++ class names (e.g. `ChapterLanguage` vs `ChapLanguage` — or the other way around).

The examples produced by @dericed's patches get that last part wrong, I think. The spec should use a format that's program-agnostic, which also means that it should use the official element names, not the C++ class names. Your examples use e.g. `ChapterString` which is the C++ class name for the `ChapString` element. Dito `ChapterLanguage` (C++ class name) vs. `ChapLanguage` (element name according to `ebml_matroska.xml`).

Alternatively the examples should state clearly that they represent a format understood by the tools from the MKVToolNix software package.

hubblec4:
Yes, thanks for this heads up Moritz.
You told me this for a long time ago. 
The xml-content files are not Matroska, a very important point to mention.

dericed:
Based on @mbunkus's comment I think I prefer to change the examples to use a program-agnostic representation of the chapters. Perhaps this needs to be defined somewhere in EBML, as I expect that having a textual representation of an EBML Document fragment will be a recurring need.

dericed:
I added a new commit to change the xml representations to use the spec's Element names rather than to adhere to mkvtoolnix's xml form. I didn't squash yet in case we'd like to revert. Please re-review.

chapters.md
```diff
@@ -220,14 +220,15 @@ The one and only command existing for the moment is `GotoAndPlay( ChapterUID );`
...
-| 0x08 | CN | Cell 1 | Cell 2 | Cell 3 | Cell 4 | Cell 5 | Cell 6 | - |
+ChapProcessPrivate | DVD Name | Hierarchy                                           | Commands Possible | Comment
+-------------------|----------|-----------------------------------------------------|-------------------|--------
+0x30               | SS       | DVD domain                                          | -                 | First Play, Video Manager, Video Title
+0x2A               | LU       | Language Unit                                       | -                 | Contains only PGCs
+0x28               | TT       | Title                                               | -                 | Contains only PGCs
+0x20               | PGC      | Program Group Chain (PGC)                           | *                 |
+0x18               | PG       | Program 1 / Program 2 / Program 3                   | -                 |
+0x10               | PTT      | Part Of Title 1 / Part Of Title 2                   | -                 | Equivalent to the chapters on the sleeve.
+0x08               | CN       | Cell 1 / Cell 2 / Cell 3 / Cell 4 / Cell 5 / Cell 6 | -                 |
```
dericed: @robUx4, could you re-test or explain, I don't see the HTML break here

hubblec4:
+1 for changing element names to the spec's Element names. So it is Matroska native.
But mkvtoolnix's xml form is better known and that will confuse people at first.

dericed:
The xml from mkvtoolnix is certainly better known, but in the context of the spec it is very hard to explain because of the inconsistencies.

dericed:
ping for re-review/re-approval. This should be merged before we start on https://github.com/Matroska-Org/matroska-specification/issues/114.

mbunkus:
I'm still OK with this.

dericed:
merging then, thx

PR #105, opened 2017-02-26
deprecate TimeSlice and LaceNumber

dericed:


robUx4:
I think some of the descriptions should go away too, they are redundant with the maxver setting.

dericed:
updated to reduce redundancy

retokromer:
It’s better, in my opinion.

robUx4:
LGTM

PR #106, opened 2017-02-26
WIP A_Opus

dericed:
partly resolves
https://github.com/Matroska-Org/matroska-specification/issues/100

### DIFF
--- a/codec_specs.md
+++ b/codec_specs.md
@@ -612,6 +612,16 @@ Description: The Wavpack packets consist of a stripped header followed by the fr
 
 Initialisation: none
 
+### A_OPUS
+
+Codec ID: A_OPUS
+
+Codec Name: Opus interactive speech and audio codec
+
+Description: Opus is designed to handle a wide range of interactive audio applications, including Voice over IP, videoconferencing, in-game chat, and even live, distributed music performances.  It scales from low bitrate narrowband speech at 6 kbit/s to very high quality stereo music at 510 kbit/s.  Opus uses both Linear Prediction (LP) and the Modified Discrete Cosine Transform (MDCT) to achieve good compression of both speech and music. Opus is defined in [RFC6716](https://tools.ietf.org/html/rfc6716).
+
+Initialisation: The `Private Data` is set to the Opus `Identification Header` as defined in [@!RFC7845].
+
 ## Subtitle Codec Mappings
 
 ### S_TEXT/UTF8


dericed:
This needs a definition of the Initialization, ping @JeromeMartinez.

Here's an example of CodecPrivate in MediaTrace
```
000001D5    CodecPrivate (22 bytes)
000001D5     Header (3 bytes)
000001D5      Name:                                 9122 (0x023A2)
000001D7      Size:                                 19 (0x013)
000001D8     Identification (19 bytes)
000001D8      opus_codec_id:                        OpusHead
000001E0      opus_version_id:                      1 (0x01)
000001E1      channel_count:                        1 (0x01)
000001E2      preskip:                              312 (0x0138)
000001E4      rate:                                 48000 (0x0BB80)
000001E8      ouput_gain:                           0
000001EA      channel_map:                          0
```

JeromeMartinez:
Look like Opus header is not required for decoding, only for metadata (i.e. channel mapping), defaults could be used (gain of 0? pre-skip from Matroska SeekPreRoll element) for other items.
Do we put Opus header mandatory or not?

mbunkus:
Back when I added Opus support to mkvmerge I decided to use [Xiph's Opus in Matroska mapping](https://wiki.xiph.org/MatroskaOpus). The elements `DiscardPadding` (child of a `BlockGroup`), `SeekPreRoll` and `CodecDelay` (children of `TrackEntry`) were introduced for handling Opus' various requirements, and the Xiph wiki was later updated to name those new elements.

I strongly recommend to make the header mandatory. You could use defaults for the channels, but not for the `output gain` field of the [Opus identification header](https://tools.ietf.org/html/rfc7845#section-5.1).

mbunkus:
To be clear: `CodecPrivate` stores the very same data structure called "Identification Header" as described in [section 5.1 of RFC 7845](https://tools.ietf.org/html/rfc7845#section-5.1).

JeromeMartinez:
For the moment, as all other Matroska elements are already in the spec, I see only

`Initialisation: The `Private Data` MUST contain the Identification Header packet, as defined in [RFC7845](https://tools.ietf.org/html/rfc7845), and no other data.`

is it correct from your point of view?

mbunkus:
It is, yes.

dericed:
It setting `Private Data` to the `Identification Header` the only consideration. Should we also say

> `SampleFrequency` MUST be set to `48000`. `SeekPreRoll` MUST be set to `80000000`. `Channels` MUST be set to the number of output PCM channels.

or address `DiscardPadding`, `SeekPreRoll` and `CodecDelay` in the Codec Mapping for Opus.

If the `Private Data` is the only consideration than the PR is updated for that.

dericed:
Ping to @mbunkus @JeromeMartinez. I'm unsure what the status is here. Should we say that A_Opus infers some additional defaults on other Elements (IMHO codec-specific defaults makes for complex semantics).

mbunkus:
> Should we also say
>> SampleFrequency MUST be set to 48000. SeekPreRoll MUST be set to 80000000. Channels MUST be set to the number of output PCM channels.
>
> or address DiscardPadding, SeekPreRoll and CodecDelay in the Codec Mapping for Opus.

No, we should not say that. Those values are valid for the current edition/version of the Opus specs. But the Matroska specs should not tie themselves to a certain version of a codec spec in such a way. What if Opus v3 decides to use a different pre-roll? We would have to adjust the Matroska specs.

If anything, we should say that `SampleFrequency` and `SeekPreRoll` be set according to the Opus version the track contains.

codec_specs.md
```diff
@@ -612,6 +612,16 @@ Description: The Wavpack packets consist of a stripped header followed by the fr
 
 Initialisation: none
 
+### A_OPUS
+
+Codec ID: A_OPUS
+
+Codec Name: Opus interactive speech and audio codec
+
+Description: Opus is designed to handle a wide range of interactive audio applications, including Voice over IP, videoconferencing, in-game chat, and even live, distributed music performances.  It scales from low bitrate narrowband speech at 6 kbit/s to very high quality stereo music at 510 kbit/s.  Opus uses both Linear Prediction (LP) and the Modified Discrete Cosine Transform (MDCT) to achieve good compression of both speech and music. Opus is defined in [RFC6716](https://tools.ietf.org/html/rfc6716).
```
mbunkus: The description is meant to be about the mapping of the codec to Matroska: what blocks contain, how header fields have to be set (if anything special has to be observed) etc. It's not supposed to be a description of the codec itself. The Matroska specs are not codec specs or their marketing department.

PR #108, opened 2017-03-02
avoid reference to ffv1 version 2

dericed:
This patch considers the conversation at http://ffmpeg.org/pipermail/ffmpeg-devel/2017-March/207699.html. Here it avoids reference to FFV1 version 2 which is not described by the FFV1 specification. It is rewritten to say that FFV1 versions 0 and 1 do not have private data, but that version 3 does have private data.

### DIFF
--- a/codec_specs.md
+++ b/codec_specs.md
@@ -248,7 +248,7 @@ Codec Name: FF Video Codec 1
 
 Description: FFV1 is a lossless intra-frame video encoding format designed to efficiently compress video data in a variety of pixel formats. Compared to uncompressed video, FFV1 offers storage compression, frame fixity, and self-description, which makes FFV1 useful as a preservation or intermediate video format. [Draft FFV1 Specification](https://datatracker.ietf.org/doc/draft-niedermayer-cellar-ffv1/)
 
-Initialisation: For FFV1 versions 2 or less, `Private Data` SHOULD NOT be written. For FFV1 version 3 and greater, the `Private Data` MUST contain the FFV1 Configuration Record structure, as defined in https://tools.ietf.org/html/draft-niedermayer-cellar-ffv1-01#section-4.1, and no other data.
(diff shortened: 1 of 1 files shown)


codec_specs.md
```diff
@@ -248,7 +248,7 @@ Codec Name: FF Video Codec 1
 
 Description: FFV1 is a lossless intra-frame video encoding format designed to efficiently compress video data in a variety of pixel formats. Compared to uncompressed video, FFV1 offers storage compression, frame fixity, and self-description, which makes FFV1 useful as a preservation or intermediate video format. [Draft FFV1 Specification](https://datatracker.ietf.org/doc/draft-niedermayer-cellar-ffv1/)
 
-Initialisation: For FFV1 versions 2 or less, `Private Data` SHOULD NOT be written. For FFV1 version 3 and greater, the `Private Data` MUST contain the FFV1 Configuration Record structure, as defined in https://tools.ietf.org/html/draft-niedermayer-cellar-ffv1-01#section-4.1, and no other data.
```
jamrial: You should probably leave the "and greater" part. The FFV1 spec in the section 4.1 states

> In the case of a bitstream with "version >= 3", a Configuration Record is stored in the underlying container, at the track header level.

Rest seems fine.

jamrial:
For reference, https://tools.ietf.org/html/draft-niedermayer-cellar-ffv1-01#section-4.8.1 is where it mentions version 2 is not meant to be considered at all.

retokromer:
As long as I know, the version 2 has been exclusively an experimental one during the switching from version 1 to version 3. I was told (by Peter Bubestinger-Steindl at the JTS in Singapore last year, if I remember carefully) that no «real» file of version 2 should exist.

codec_specs.md
dericed: thx @jamrial, I updated the patch to say "version 3 and greater"

dericed:
Patch updated, please re-review.

retokromer:
LGTM

dericed:
will merge soon, unless there's a comment otherwise

robUx4:
LGTM, merged

PR #112, opened 2017-03-14
add T_QUICKTIME codec mapping

dericed:
intended as alternative to https://github.com/Matroska-Org/matroska-specification/pull/102

### DIFF
--- a/codec_specs.md
+++ b/codec_specs.md
@@ -23,6 +23,7 @@ Video      | "V_"
 Audio      | "A_"
 Subtitle   | "S_"
 Button     | "B_"
+Timecode   | "T_"
 
 Each `Codec ID` MUST include a `Major Codec ID` immediately following the `Codec ID Prefix`. A `Major Codec ID` MAY be followed by an OPTIONAL `Codec ID Suffix` to communicate a refinement of the `Major Codec ID`. If a `Codec ID Suffix` is used, then the `Codec ID` MUST include a forward slash ("/") as a separator between the `Major Codec ID` and the `Codec ID Suffix`. The `Major Codec ID` MUST be composed of only capital letters (A-Z) and numbers (0-9). The `Codec ID Suffix` MUST be composed of only capital letters (A-Z), numbers (0-9), underscore ("_"), and forward slash ("/").
 
@@ -713,3 +714,15 @@ Codec ID: B_VOBBTN
 Codec Name: VobBtn Buttons
 
 Description: Based on [MPEG/VOB PCI packets](http://dvd.sourceforge.net/dvdinfo/pci_pkt.html). The file contains a header consisting of the string "butonDVD" followed by the width and height in pixels (16 bits integer each) and 4 reserved bytes. The rest is full [PCI packets](http://dvd.sourceforge.net/dvdinfo/pci_pkt.html).
+
+## Timecode Codec Mappings
+
+### T_QUICKTIME
+
+Codec ID: T_QUICKTIME
+
+Codec Name: QuickTime timecode track
+
+Description: `T_QUICKTIME` is used to identify Timecode tracks as stored in QuickTime. The `QuickTime Timecode Sample Data` is stored within the Block's data section. For an explanation of `QuickTime Timecode Sample Data` read [QuickTime File Format Specification](https://developer.apple.com/library/content/documentation/QuickTime/QTFF/QTFFChap3/qtff3.html#//apple_ref/doc/uid/TP40000939-CH205-57421).
+
+Initialisation: The `Private Data` contains the `Timecode Sample Description` which in QuickTime is stored within the 'tmcd' atom after the mandatory `data reference index` value. For an explanation of the `Timecode Sample Description` read [QuickTime File Format Specification](https://developer.apple.com/library/content/documentation/QuickTime/QTFF/QTFFChap3/qtff3.html#//apple_ref/doc/uid/TP40000939-CH205-69831).
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -201,6 +201,7 @@
       <enum value="1" label="video"/>
       <enum value="2" label="audio"/>
       <enum value="3" label="complex"/>
+      <enum value="4" label="timecode"/>
       <enum value="16" label="logo"/>
       <enum value="17" label="subtitle"/>
       <enum value="18" label="buttons"/>


dericed:
Related message at https://mailarchive.ietf.org/arch/msg/cellar/flGPjiu-YBdpAgDQ-IUZNBY1wfk

dericed:
This PR was rewritten to split T_QUICKTIME into T_QUICKTIME/SINGLE and T_QUICKTIME/MULTIPLE. The SINGLE version only uses a single Cluster and timecode value to associate a virtual continuous QuickTime timecode with the Matroska timecode. This is the way most QuickTime timecode tracks work, a single value stored, and every other value calculated based off private data. The MULTIPLE version infers that more than one value may be stored. I may add some example math here but wanted some review of the concepts here. Ping to @jeromemartinez.

codec_specs.md
```diff
@@ -727,3 +728,24 @@ Codec Name: VobBtn Buttons
...
 
+## Timecode Codec Mappings
+
+### T_QUICKTIME/SINGLE
+
+Codec ID: T_QUICKTIME/SINGLE
+
+Codec Name: QuickTime timecode track (continuous)
+
+Description: `T_QUICKTIME/SINGLE` is used to identify Timecode tracks as stored in QuickTime. The `QuickTime Timecode Sample Data` is stored within the Block's data section. For an explanation of `QuickTime Timecode Sample Data` read [QuickTime File Format Specification](https://developer.apple.com/library/content/documentation/QuickTime/QTFF/QTFFChap3/qtff3.html#//apple_ref/doc/uid/TP40000939-CH205-57421). In the context of Matroska, `T_QUICKTIME/SINGLE` tracks MUST only relate to exactly one `Cluster` which MUST have a `Cluster Timecode` of zero and exactly one `SimpleBlock` or `BlockGroup Element`. The storage of the `QuickTime Timecode Sample Data` with the `SimpleBlock` or `BlockGroup Element` allows the `QuickTime Timecode` to be mapped to the `Matroska Timecode` at a single point. The `QuickTime Timecode` value MAY then be evaluated at any point in the `Matroska Timecode` by using the initialised `Timecode Sample Description`, the single `QuickTime Timecode Sample Data` value, and the `Matroska Timecode`.
```
retokromer: British `initialised` or American `initialized`? The rest seems to me American.

codec_specs.md
```diff
@@ -727,3 +728,24 @@ Codec Name: VobBtn Buttons
...
+
+Initialisation: The `Private Data` contains the `Timecode Sample Description` which in QuickTime is stored within the 'tmcd' atom after the mandatory `data reference index` value. For an explanation of the `Timecode Sample Description` read [QuickTime File Format Specification](https://developer.apple.com/library/content/documentation/QuickTime/QTFF/QTFFChap3/qtff3.html#//apple_ref/doc/uid/TP40000939-CH205-69831).
```
retokromer: same here

codec_specs.md
mbunkus: Two things I'd like to discuss:

> In the context of Matroska, `T_QUICKTIME/SINGLE` tracks MUST only relate to exactly one `Cluster` which MUST have a `Cluster Timecode` of zero and exactly one `SimpleBlock` or `BlockGroup Element`.

I honestly don't understand what this is supposed to mean, especially the verb "relate". Do you mean to say that for a track with the CodecID `T_QUICKTIME/SINGLE` there MUST be exactly one `BlockGroup` or `SimpleBlock`, that this singular block is stored in a cluster with `ClusterTimecode` set to 0, and that there MUST NOT be any other cluster in the file with `ClusterTimecode` set to 0?

If yes, why such severe restrictions? Why not allow multiple `Clusters` that start at 0 (by simply omitting the restriction)? What about files whose first cluster doesn't start at 0 — how would that be interpreted? And why are you saying something about the `Cluster Timecode` but not about the timecode parts of the blocks?

In other words: can you please describe all of your thoughts in much more detail? We can then see how to turn that into unambiguous verbiage.

mbunkus: The second issue I have is with linking to other web pages as part of the format specification. I'd rather we refer to existing PDF documentation or publications. In the case of linking to the QuickTime specification this is both possible and a bit difficult: possible because there is [a PDF version](https://developer.apple.com/standards/qtff-2001.pdf) of the QuickTime specs, and harder because it doesn't contain numbered chapters and sections. I still prefer to say something like this:

> For an explanation of the format read the section "Timecode Media", sub-section "Timecode Sample Description", of Apple's [QuickTime File Format Specification (2001)](https://developer.apple.com/standards/qtff-2001.pdf).

codec_specs.md
mbunkus: Same issue with referencing web pages vs. a published standard (as a PDF).

codec_specs.md
dericed: In this case, I suggest that `T_QUICKTIME/SINGLE` implies that only a single Cluster is associated with that track. I didn't mean that imply that other tracks could not share the same timecode value. Upon reflection I think the requirement to use a Cluster Timecode is unnecessary. What is needed in T_QUICKTIME/SINGLE is for a single `QuickTime Timecode Sample Data` value to be stored in a single `BlockGroup` or `SimpleBlock` of a single `Cluster`, then the `QuickTime Timecode Sample Data` (within the Block) and the `Timecode Sample Description` (in Private Data) and the `Cluster Timecode` and the `Block Timecode` can be used to calculate what the QuickTime Timecode value should be relative to any Matroska Timecode value. Still for a Matroska Reader that needs to quickly associate any Matroska Timecode value with a QuickTime Timecode value it would be good to know where that Cluster is quickly. I had intended to have Cluster as 0 timecode to mean that it should be one of the first Clusters, but of course that is not a good expectation.

mbunkus: If finding the (single) `BlockGroup`/`SimpleBlock` is the goal, then just say that the `BlockGroup`/`SimpleBlock` SHOULD be referenced via `Cues`.

I don't like requiring that the first or one of the first cluster contains that data as that seems problematic wrt streaming Matroska files.

dericed: Are there any other Codec Mappings that require Cues?

mbunkus: None that I know of.

mcr:
This may require changes in Matroksa, and it may not be possible to do this in the same document.

dericed:
I think this idea is much lower priority (and perhaps unneeded) since there https://github.com/cellar-wg/matroska-specification/pull/348. Close?

JeromeMartinez:
> I think this idea is much lower priority (and perhaps unneeded) since there #348. Close?

The "format" is different between QuickTime time code (there is a CodecPrivate part with "24Hours" etc flags, then an offset from 00:00:00) and SMPTE ST 12 (hours, minutes, seconds, frames), and conversion can be lossy e.g. if hours count is 30 (possible in QuickTime, not possible in SMPTE ST 12).

Low priority: right.
Unneeded: I disagree, not the same target of users (one is lossless transwrapping from QuickTime, the other is lossless transwrapping from MXF).

They are just 2 different formats with no bijection between them.

codec_specs.md
```diff
@@ -23,6 +23,7 @@ Video      | "V_"
 Audio      | "A_"
 Subtitle   | "S_"
 Button     | "B_"
+Timecode   | "T_"
```
robUx4, line 26: There is currently no "timecode" track type. There is a "metadata" track type which seems to be fitting. So maybe the prefix should be `"M_"` and the type called `Metadata` ?

We can also add a "timecode" track type. In the end it's more informational for the player what this is about, although it should be able to tell what kind of metadata we're talking about just by looking at the codec. But at the container level (don't know what's in codecs) it might be usefull to tell if we have timecodes or not.

codec_specs.md
```diff
@@ -727,3 +728,24 @@ Codec Name: VobBtn Buttons
 
 Description: Based on [MPEG/VOB PCI packets](http://dvd.sourceforge.net/dvdinfo/pci_pkt.html). The file contains a header consisting of the string "butonDVD" followed by the width and height in pixels (16 bits integer each) and 4 reserved bytes. The rest is full [PCI packets](http://dvd.sourceforge.net/dvdinfo/pci_pkt.html).
 
+## Timecode Codec Mappings
+
+### T_QUICKTIME/SINGLE
+
+Codec ID: T_QUICKTIME/SINGLE
```
robUx4, line 735: We need to tell what track type to use "metadata", "subtitle", "timecode" ? Maybe this should be explained in the chapter above.

robUx4 (changes requested):
we need to tell the track type to use

robUx4:
Moving this to Matroska v5 because:

- timecode is not an existing track type
- the codec doesn't exist yet
- this may be a new block addition

PR #116, opened 2017-04-21
fix link to diagram.md

retokromer:
Should resolve https://github.com/Matroska-Org/matroska-specification/issues/111.


mbunkus:
LGTM, so…

PR #117, opened 2017-04-29
offer recommendations for new codec mappings

dericed:


robUx4:
LGTM

codec_specs.md
```diff
@@ -58,6 +58,18 @@ A timestamp, expressed in [@!RFC3339] that notes when support for the `Codec Map
 
 A `Codec Mapping` MAY only be defined with a `Superseded By` value, if it has an expressed `Deprecation Date`. If used, the `Superseded By` value MUST store the `Codec ID` of another `Codec Mapping` that has superseded the `Codec Mapping`.
 
+## Recommendations for the Creation of New Codec Mappings
+
+Creators of new Codec Mappins to be used in the context of Matroska:
```
mbunkus: "Mappins" is missing a "g"

mbunkus: Backtick all special terms just as above: `Codec Mapping`, `Codec ID` and `Codec Name`

dericed:
I updated the PR according to @mbunkus's suggestions regarding typos and backticks; however we should take this message from cellar into consideration https://mailarchive.ietf.org/arch/msg/cellar/NDLFWY58Nhnw4xaNX86wCFoPF0A.

Copy/paste:
>It's important to note that while IANA deprecated the X- prefix, they *replaced
*it with the Vendor and Personal trees.
> 
> https://tools.ietf.org/html/rfc6838#section-3
> 
> So, you can certainly say "don't use X_".  But, if someone has an encoding
that doesn't fit any of the existing prefixes, then using "Y_" is just as
bad (or worse).

codec_specs.md
```diff
@@ -58,6 +58,18 @@ A timestamp, expressed in [@!RFC3339] that notes when support for the `Codec Map
...
 
+## Recommendations for the Creation of New Codec Mappings
+
+Creators of new `Codec Mappings` to be used in the context of Matroska:
+
+- SHOULD assume that all `Codec Mappings` they create might become standardized, public, commonly deployed, or usable across multiple implementations.
+
+- SHOULD employ meaningful values for `Codec ID` and `Codec Name` that they have reason to believe are currently unused.
+
+- SHOULD NOT prefix their `Codec ID` with "X-" or similar constructs.
```
retokromer: Should this not read `X_`? (underline, not dash)

mbunkus: Indeed :)

dericed: `X-` is leftover from the similar recommendation from rfc6838. Yes `X_` would be more in Matroska style. Still we in addition to saying what not to do, we should recommend what to do. For example, enough info so that the poster of https://mailarchive.ietf.org/arch/msg/cellar/igOaI6HoCGx_K8nFceXMetoAoVQ could asssign a Codec ID.

dericed: Not sure. For instance if an organization using Matroska starts using their own `V_EXPERIMENTAL/VP123`, then a year later proposes it for official support, it doesn't make sense to leave that experimental name as is, but we'd likely prefer `V_VP123`. If the original users simply used `V_VP123` from the beginning and the experimental nature was just evident since it wasn't initially in the Codec Mapping, then there wouldn't be an issue with changing Codec IDs.

dericed:
I rebased to change a use of `X-` to Matroska style `X_`. There are still open issues that @robUx4 listed in https://github.com/Matroska-Org/matroska-specification/pull/117#discussion_r117633284. At the moment, I propose considering to merge this as a suggestion of what-not-to-do. We can consider what-to-do in a later PR.

mbunkus:
I concur.

PR #118, opened 2017-04-29
extend docs for ColourSpace

dericed:


robUx4:
We are losing the fact that the values are usually borrowed from the same element in AVI (to be found in the specs).

dericed:
From https://msdn.microsoft.com/en-us/library/windows/desktop/dd318229(v=vs.85).aspx I see this definition:
```
biCompression

    For compressed video and YUV formats, this member is a FOURCC code, specified as a DWORD in little-endian order. For example, YUYV video has the FOURCC 'VYUY' or 0x56595559. For more information, see FOURCC Codes.

    For uncompressed RGB formats, the following values are possible:

    Value	            Meaning
    BI_RGB:           Uncompressed RGB.
    BI_BITFIELDS: Uncompressed RGB with color masks. Valid for 16-bpp and 32-bpp bitmaps.

    See Remarks for more information. Note that BI_JPG and BI_PNG are not valid video formats.

    For 16-bpp bitmaps, if biCompression equals BI_RGB, the format is always RGB 555. If biCompression equals BI_BITFIELDS, the format is either RGB 555 or RGB 565. Use the subtype GUID in the AM_MEDIA_TYPE structure to determine the specific RGB type.
```

I think this doesn't match Matroska practice, as I've never seen ColourSpace used for a compressed format, but only for uncompressed. For `V_UNCOMPRESSED` video I suggest that either this field is mandatory or that Private Data is defined (the first option preferred since it is used in practice).

robUx4:
OK but we still have no idea where to find and how to interpret this fourcc value.

retokromer:
I agree with @robUx4 and I am still struggling with it.

I also suggest to use the `FourCC` writing.

dericed:
IMHO the concept of `FourCC` is poorly standardized. Given that AVI has no maintenance in a long time, is it worth defining the fourcc's for this element here (namely the uncompressed video ones)? Else we could defer to another Fourcc source, but not many good ones that I can find. I think that someone reading this spec and its normative citations should understand how to read, for instance, a V_UNCOMPRESSED file with I420 or UYVY as ColourSpace. We don't really have a V_UNCOMPRESSED authority to point to, so the interpretation of this element is very critical.

retokromer:
I am still struggling with it :-(

robUx4:
I agree it's better to leave any uncertainty out. I guess we should define it somewhere in the specs/notes and point to that. And make it match with AVI stuff where possible.

dericed:
I added this sentence to reference AVI's use of the value:
> This value is similar in scope to the biCompression value of AVI's BITMAPINFOHEADER.

Please re-review

ebml_matroska.xml
```diff
@@ -404,7 +404,7 @@
     </restriction>
   </element>
   <element name="ColourSpace" path="0*1(\Segment\Tracks\TrackEntry\Video\ColourSpace)" cppname="VideoColourSpace" id="0x2EB524" type="binary" maxOccurs="1" minver="1" webm="0" size="4">
-    <documentation lang="en">Same value as in AVI (32 bits).</documentation>
+    <documentation lang="en">Specify the pixel format used for the Track's data as a fourcc. This value is similar in scope to the biCompression value of AVI's BITMAPINFOHEADER. This Element is MANDATORY in TrackEntry when the CodecID Element of the TrackEntry is set to "V_Uncompressed".</documentation>
```
mbunkus: CodecIDs must be all-uppercase: `V_UNCOMPRESSED`, not `V_Uncompressed`

mbunkus: `FourCC` is actually an abbreviation for [four-character code](https://en.wikipedia.org/wiki/FourCC). Therefore `fourcc` should be `FourCC`.

dericed:
Updated for capitalization per @mbunkus's suggestions at:
https://github.com/Matroska-Org/matroska-specification/pull/118#discussion_r125188519 https://github.com/Matroska-Org/matroska-specification/pull/118#discussion_r125188540

Please re-review

PR #119, opened 2017-05-13
diagram rewrite

dericed:


dericed:
Hi all, In the prior RFC drafts for Matroska, the content of the diagram.md page was left out. Overall I think a structural overview would be helpful so I did some work to rewrite this document. In general the original diagram page (see https://matroska.org/technical/diagram/index.html) includes a concise summary of Top Level Elements, then a longer summary of Top Level Elements, and some diagrams. I consolidated this down to one summary of the Top Level Elements and converted the diagrams from HTML to something more RFC-friendly.

For diagrams, I used two different styles to represent Matroska structure. Any preference?

Style 1, ascii art tables
```
+--------------------------------+
| SeekHead | Seek | SeekID       |
|          |      |--------------|
|          |      | SeekPosition |
+--------------------------------+
```

Style 2, bullets
* Portion of Block
  * Data Type
     * Bit Flag
* Header
  * TrackNumber
  * Timecode
  * Flags
    * Gap
    * Lacing
    * Reserved
* Optional
  * FrameSize
* Data
  * Frame

retokromer:
Hmm… I presume, in RFC it should be ASCII art.

diagram.md
```diff
@@ -2,240 +2,284 @@
...
+|         |-------------|
+|         | Tags        |
++-----------------------+
+```
+
+## Matroska Top Level Elements
+
+The Matroska `EBML Schema` defines eight `Top Level Elements`: `SeekHead`, `Info`, `Tracks`, `Chapters`, `Cluster`, `Cues`, `Attachments`, and `Tags`.
+
+The `SeekHead Element` (also known as `MetaSeek`) contains an index of where other `Top Level Elements` of the `Segment` are located in order to let the parser know where the other major parts of the file are. This element isn't technicaly REQUIRED, but without a `SeekHead Element` a `Matroska Parser` would have to search the entire file to find all of the other Top Level Elements. This is because Matroska has flexible ordering requirements; for instance, the `Chapters Element` could be stored in the middle of the `Cluster Elements`.
```
mbunkus: "…stored in the middle of the `Cluster Elements`" — I'd prefer a more realistic example, one that's actually used in the wild: just change "in the middle of" to "after".

diagram.md
```diff
@@ -2,240 +2,284 @@
...
+- Segment
+
+A more complex Matroska file consisting of an `EBML Stream`  (consisting of two `EBML Documents`) could be represented like this:
+
+- EBML Header
+- Segment
+- EBML Header
+- Segment
+
+The following diagram represents a simple Matroska file, comprised of an `EBML Document` with an `EBML Header`, a `Segment Element` (the `Root Element`), and all eight Matroska `Top Level Elements`.
```
mbunkus: Maybe add some explanation that horizontal spacing expresses parent-child relationship/containment (e.g. the `Info` element is contained within the `Segment` element) whereas vertical alignment indicates the order in the file.

diagram.md
```diff
@@ -2,240 +2,284 @@
...
+|        |            | Name         |
+|        |            |--------------|
+|        |            | Language     |
+|        |            |--------------|
+|        |            | CodecID      |
+|        |            |--------------|
+|        |            | CodecPrivate |
+|        |            |--------------|
+|        |            | CodecName    |
+|        |            |--------------|-------------------+
```
mbunkus: In the very first table you've made the `EBML Head` entry span the width of the whole table, but starting with this table you don't make the upper elements as wide as the table. I think this should be consistent over all tables.

diagram.md
```diff
@@ -2,240 +2,284 @@
...
+|          |              |                    | ChapterDisplay    | ChapString   |
+|          |              |                    |                   |--------------| 
+|          |              |                    |                   | ChapLanguage | 
++---------------------------------------------------------------------------------+
+```
+Figure: Representation of the `Chapters Element` and a selection of its `Descendant Elements`.
+
+The `Cluster Elements` contain all of the video frames and audio for each track. In a given Matroska file, there are usually many `Cluster Elements`. The Clusters help to break up the `SimpleBlock` or `BlockGroup Elements` and help with seeking and error protection. There is no set limit to how much data a Cluster can contain, or how much time they can span, but a limit at 5 seconds or 5 megabytes is RECOMMENDED. Every Cluster contains a timecode, usually the timecode that the first Block in the Cluster SHOULD be played back, but it doesn't have to be. Then there are one or more (usually many more) `BlockGroups` or `SimpleBlocks` in each Cluster. A BlockGroup can contain a Block of data, and any information relating directly to that Block.
```
mbunkus: It's not quite correct that "there is no set limit…". One limit is implied by the fact that the `BlockGroup` and `SimpleBlock` elements use a 16bit signed integer to represent the relative timestamp (relative to the cluster's `ClusterTimestamp` element), scaled by the track's `TimecodeScale` element. Therefore the amount of time one can place in a cluster is definitely limited. For example, with the default resolution of 1ms you can only store ~32s of data (~65s if you use negative relative values, too).

diagram.md
retokromer: \` missing for the second occurrence of \`Top Level Elements`

diagram.md
```diff
@@ -2,240 +2,284 @@
...
+|        |            |              |-------------------|
+|        |            |              | AspectRatioType   |
+|        |            |              |-------------------|
+|        |            |              | Color             |
+|        |            |--------------|-------------------|
+|        |            | Audio        | SamplingFrequency |
+|        |            |              |-------------------|
+|        |            |              | Channels          |
+|        |            |              |-------------------|
+|        |            |              | BitDpeth          |
```
retokromer: BitDepth

diagram.md
```diff
@@ -2,240 +2,284 @@
...
+
+The `ReferenceBlock` in the BlockGroup, is used instead of the basic "P-frame"/"B-frame" description. Instead of simply saying that this Block depends on the Block directly before, or directly afterwards, we put the timecode of the needed Block. And because you can have as many `ReferenceBlock Elements` as you want for a Block, it allows for some extremely complex referencing.
+
+The `Cues Element` is used to seek when playing back a file by providing a temporal index for each of the tracks. It is similar to the `SeekHEad Element`, but this is used for seeking to a specific time when playing back the file. Without this it is possible to seek, but it is much more difficult because the player has to 'hunt and peck' through the file looking for the correct timecode. `Cues` contains `CuePoint Elements` which store the timecode (`CueTime`) and then a listing for the exact position in the file for each of the tracks for that timecode. The `Cues` are pretty flexible for what exactly you want to index. For instance, you can index every single timecode of every `Block` or index selectively. If you have a video file, it is RECOMMENDED to index at least the keyframes of the video track.
```
retokromer: \`SeakHead` (second "e" not capitalised)

diagram.md
```diff
@@ -2,240 +2,284 @@
...
+|             |              | FileReferral      |
+|             |              |-------------------|
+|             |              | FileUsedStartTime |
+|             |              |-------------------|
+|             |              | FileUsedEndTime   |
++------------------------------------------------+
+```
+Figure: Representation of a `Attachments Element`.
+
+The `Tags Element` contains metadata that describes the `Segment` and potentially its `Tracks`, `Chapters`, and `Attachments`. Each Track or Chapter that those tags applies to has its UID listed in the tags. The Tags contain all extra information about the file, script writer, singer, actors, directors, titles, edition, price, dates, genre, comments, etc. And it allows you to enter many of these (title, edition, comments, ect) in different languages.
```
retokromer: "etc." in "(title, edition, comments, ect)"

diagram.md
dericed: thx. I added
> In the following diagrams of this section, horizontal spacing expresses a parent-child relationship between Matroska Elements (e.g. the `Info Element` is contained within the `Segment Element`) whereas vertical alignment represents the storage order within the file.

diagram.md
dericed: I simplified the sentence down to:
> It is RECOMMENDED the size of each individual `Cluster Element` be limited to store no more than 5 seconds or 5 megabytes.

But is this recommendation still valid for all track types. For instance would it differ based on if a track was video or text?

dericed:
I updated the PR according to @mbunkus and @retokromer's reviews and added a question at https://github.com/Matroska-Org/matroska-specification/pull/119#discussion_r116365282.

diagram.md
retokromer: Good point! What is the reason to limit it to 5 MB?

dericed: The recommendation is the same before and after the pr, so I recommend that the issue not block the pr unless it can be resolved quickly, else let's move to an issue.

retokromer: No, this should not block the PR. By reading it, I was just thinking that 5 MB is «a little» ridiculous today…

robUx4:
LGTM.
For a patch that large we shouldn't focus on wording or little details, it can be merged and refined later.

Issue #122, opened 2017-05-17
range for LuminanceMax and LuminanceMin should be updated to hex floating point constants


dericed:
see range attributes in https://github.com/Matroska-Org/matroska-specification/blob/fce29626d0c7d7d7b0a0543f82cb3da924bce8d4/ebml_matroska.xml#L548-L553

robUx4:
This is fun

robUx4:
Also, do we actually need ranges ? Just positive would be enough ? It also doesn't cover the minimum precision (0.1 cd/m² or 0.001 cd/m²) ?

retokromer:
Hmm… _intuitively_, I would say that for luminance positive is enough, but the precision should be added.

robUx4:
IMO just saying they must be positive is enough. You don't know what precision people will want to store. Namely for scientific recordings they would record the sun or very few photons in dark environments. We should be able to handle that and not reject such files. Negative luminance doesn't make much sense though.

retokromer:
Yep, agreed: let’s go for positive only.

dericed:
ping to @FrankGalligan

retokromer:
FYI: I asked two of my former students working on scientific imaging (at CERN and EPFL). Both support positive only and both suggest to give the possibility (not mandatory) to store also the precision.

robUx4:
The problem with the precision is that it's really dependent on the float format and the actual value. I don't think current video standards define the precision, just the range. The lower bound can be seen as a precision but it's not.

retokromer:
I understand plenty your argument, but in the scientific world a value without a precision is not one. That said, I am personally fine with positive only.

robUx4:
I think for scientific purpose the precision of the instrument would be stored separately. The precision may also vary depending on the value (low range, mid range, high range).

retokromer:
The idea is to store as much as possible with the file itself. That’s also my dream e.g. for the LUTs and the sensor’s metadata that determine the «raw» data – this defines the way the data are cooked.

robUx4:
Fixed after the merged changes.

PR #127, opened 2017-05-26
change the luminance ranges to postive

robUx4:
We don't know if there won't be very small values or very large values someday.
Negative values there won't be.

Fixes issue #122

### DIFF
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -545,10 +545,10 @@
   <element name="WhitePointChromaticityY" path="0*1(\Segment\Tracks\TrackEntry\Video\Colour\MasteringMetadata\WhitePointChromaticityY)" cppname="VideoWhitePointChromaY" id="0x55D8" type="float" maxOccurs="1" range="0-1" minver="4" webm="0">
     <documentation lang="en">White Y chromaticity coordinate as defined by CIE 1931.</documentation>
   </element>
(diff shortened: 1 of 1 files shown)


dericed:
ping @frankgalligan

dericed:
Don't you mean `>=`? After patch a zero value is not valid but before it is ok.

robUx4:
Mh, I would think 0 is an invalid value. Yet again, it might be technically possible to film in a black box where there's absolutely no light by default. So I agree.

dericed:
@FrankGalligan's original range for LuminanceMax included 0. An exactly black recording is certainly possible so I suggest `range="&gt;= 0x0p+0"` for both LuminanceMax and LuminanceMin.

retokromer:
I discussed this at CERN last Friday. They don’t see a case study for `LuminanceMax = 0`. Yet I’m fine with either way allowing or not allowing it, as it don’t hurt.

dericed:
I suggest that Matroska and webm keep compatibility where we can. With this change a webm would be valid with LuminanceMax=0 but when remuxed to Matroska would be invalid. See: https://www.webmproject.org/docs/container/.

retokromer:
I fully agree on this!

robUx4:
OK I updated to use `>=0` for both.

PR #128, opened 2017-05-26
deprecate AttachmentLink

robUx4:
Following the discussions on https://github.com/Matroska-Org/matroska-specification/pull/115


dericed:
Not `maxver='4'`?

dericed:
Actually the only examples I can find of AttachmentLink (which is only 2 divx files) are both Matroska version 1, so I'm ok with setting this as maxver=3. @mbunkus?

mbunkus:
I'm fine with this change, too.

PR #130, opened 2017-05-26
unify British & typos

retokromer:
I also suggest (not corrected here yet) to use:

- `trans-muxed` instead of `transmuxed`
- `bit-depth` instead of `bitdepth`
- `sample-rates` instead of `samplerates`
- `sub-stream` instead of `substream`
- `seek-table` instead of `seektable`

I strongly suggest to write Dolby without the stupid <sup>TM</sup>.

### DIFF
--- a/codec_specs.md
+++ b/codec_specs.md
@@ -3,7 +3,7 @@ layout: default
 ---
 # Codec Mappings
 
-A `Codec Mapping` is a set of attributes to identify, name, and contextualize the format and characteristics of encoded data that can be contained within Matroska Clusters.
+A `Codec Mapping` is a set of attributes to identify, name, and contextualise the format and characteristics of encoded data that can be contained within Matroska Clusters.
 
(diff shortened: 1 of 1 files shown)


retokromer:
Should it read `seekable` instead of `seektable`?

dericed:
`seekable` and `seektable` aren't analogous. FLAC is still `seekable` with or without a `seektable`. Potentially we could say something like the degree to wish it is `seekable` is impacted by the presence of a `seektable`.

retokromer:
If I remember carefully, there is one `seektable` and some `seekable`. I will check when back from Amsterdam.

dericed:
Argh, I realize that my above comments were written while thinking about the flac rather than the mkv repo. This PR LGTM. Nitpick, I'd suggest `seek table` over `seektable` as that's how it's expressed in http://tausoft.org/wiki/True_Audio_Codec_Format.

PR #135, opened 2017-06-11
add ebml header/global elements to ebml schema

dericed:
As discussed at
https://github.com/Matroska-Org/ebml-specification/issues/128. This is
the same as added to the ebml specification’s example of ebml schema,
except Matroska adds a header constaint for EBMLMaxSizeLength to say
that the range should be ‘1-8’ rather than ‘not 0’.

### DIFF
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -1,5 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
 <EBMLSchema docType="matroska" version="4">
+  <!-- constraints on EBML Header Elements -->
+  <element name="EBMLMaxIDLength" path="1*1(\EBML\EBMLMaxIDLength)" id="0x42F2" minOccurs="1" maxOccurs="1" range="4" default="4" type="uinteger"/>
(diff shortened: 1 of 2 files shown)


dericed:
This PR is dependent on https://github.com/Matroska-Org/ebml-specification/pull/152 being merged. Please do not merge this before that.

robUx4:
Do we have to define elements that are different from the base specs or all of them ?
I'd prefer the former as it might avoid issues. And if EBML evolves it evolves automatically with it. On the other hand it might no be desirable for some. So for example the EBMLVersion might be forced to a fixed value.

dericed:
This PR should wait until https://github.com/Matroska-Org/ebml-specification/pull/152 is resolved and merged.

dericed:
Now that https://github.com/Matroska-Org/ebml-specification/pull/152 is resolved, this PR is updated. Now the EBML Schema for Matroska adds two constrained uses of the EBML Header Elements. These constraints are already documented in https://github.com/Matroska-Org/matroska-specification/blob/e45973165e9019fa6244125ed9dbd12257e397b8/index.md#added-constaints-on-ebml in narrative form. In that narrative form, I changed `8 or less` to `between 1 and 8 inclusive` to be more clear.

Please re-review.

PR #138, opened 2017-07-01
Fix a typo

lu-zero:


lu-zero:
We should maybe keep this this PR open while I keep reading the specification

lu-zero:
I have @robUx4 next to me and hopefully I'll send some comments in the shape of patches  in a bit : )

mbunkus:
Just let us know when we can start the review. And thanks!

lu-zero:
So far those are all the typos I found, but as you noticed I moved my focus on the ebml specification =)

dericed:
Will merge soon, as today is the doc freeze for IETF specs before IETF99.

PR #139, opened 2017-07-03
version 03

dericed:


dericed:
I don't think there's anything else we intend to squeeze through today. I'll merge this soon.

PR #147, opened 2017-07-16
add a local website generator to the Makefile

### LINKED ISSUE #146: Generate the github website from the Makefile
It may just be running some Jekyll commands as a Makefile target (like `website`).

That would help keep our Markdown and website generator happy together for longer.

robUx4:
Similar to the one generated automatically on Github:
http://matroska-org.github.io/matroska-specification/

Fixes #146

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,10 @@ $(OUTPUT).md: rfc_frontmatter.md index.md diagram.md matroska_schema_section_hea
 ebml_matroska_elements.md: ebml_matroska.xml transforms/ebml_schema2markdown.xsl
 	xsltproc transforms/ebml_schema2markdown.xsl $< > $@
 
+website:
+	jekyll b
+
 clean:
-	rm -f $(OUTPUT).txt $(OUTPUT).html $(OUTPUT).md $(OUTPUT).xml ebml_matroska_elements.md ebml_matroska_elements4rfc.md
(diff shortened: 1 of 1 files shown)


ablwr:
Well, look at you go!

robUx4:
That part is easy.
I wanted to remove the front-matter from each MarkDown file via the `_config.yml` but it doesn't seem possible.

robUx4:
It's possible to have just ```---
---``` at the beginning of files. Which would generate clean IETF output. But the `layout: default` information is lost and it seems forcing it in `_config.yml` doesn't work :(

ablwr:
I'll give it a try -- just now got my computer set up to have the `make` command build things properly instead of throwing errors at me! With regards to this specific change, 👍

robUx4:
I created a branch with my failed attempt at removing the `layout: default` from each Markdown file.
https://github.com/Matroska-Org/matroska-specification/tree/layout-default

ablwr:
I definitely remember struggling with this ~1 year ago because I didn't want the front-loaded YAML in there, either, but wasn't able to get a workaround (and didn't want to spend too much time on it). But Jekyll and I have both matured since then...

ablwr:
This does seem to just be a limitation of Jekyll. I will update the _config.yml file with the default scope so the pages themselves are cleaner.

Issue #145, opened 2017-07-16
Fix internal documentation links


robUx4:
Right now the Markdown code has a lot of `{{site.baseurl}}/index.html#Cluster` links. But they fail to link properly in the code.
I'm not sure how to fix this properly. Also I'm not sure it is good pratice to put links inside an RFC at all.

ablwr:
This, too, is tied to my response to the `layout: default` issue you brought up on the CELLAR list: measures taken to automatically build and replicate the HTML structure.

ablwr:
I see that the link you are referring to, specifically, no longer links to anything, anyway. We should remove the broken ones.

Intra-document links sometimes work in the generated html, if formatted correctly (all lowercase, swapping spaces for hyphens). But, like our other conversations about this subject -- the most important source is the generated RFC, not the website.

I think it's best to remove the links for now, also because the document is in flux and some headers cease to exist, some new ones spring up, and others change. This would break the existing links and cause extra confusion!

ablwr:
I will go through and update the links accordingly. If it's an intra-document link created just to be helpful, I will remove it. If it's a regular link, I will leave it. If it's an intra-document like that is phrased in a "See [this]", I will do my best to keep it.

robUx4:
Fixed

PR #150, opened 2017-07-16
removes non-functional intra-document links

ablwr:
Resolves formatting issue presented in #145

### DIFF
--- a/chapters.md
+++ b/chapters.md
@@ -23,17 +23,17 @@ The `EditionFlagHidden` Flag behaves similarly to the `ChapterFlagHidden` Flag: 
 
 The menu features are handled like a _chapter codec_. That means each codec has a type, some private data and some data in the chapters.
 
-The type of the menu system is defined by the ChapProcessCodecID parameter. For now only 2 values are supported : 0 matroska script, 1 menu borrowed from the DVD. The private data depend on the type of menu system (stored in ChapProcessPrivate), idem for the data in the chapters (stored in ChapProcessData).
+The type of the menu system is defined by the `ChapProcessCodecID` parameter. For now only 2 values are supported : 0 matroska script, 1 menu borrowed from the DVD. The private data depend on the type of menu system (stored in ChapProcessPrivate), idem for the data in the chapters (stored in ChapProcessData).
 
 ### Matroska Script (0)
 
-This is the case when ChapProcessCodecID = 0\. This is a script language build for Matroska purposes. The inspiration comes from ActionScript, javascript and other similar scripting languages. The commands are stored as text commands, in UTF-8\. The syntax is C like, with commands spanned on many lines, each terminating with a ";". You can also include comments at the end of lines with "//" or comment many lines using "/* \*/". The scripts are stored in ChapProcessData. For the moment ChapProcessPrivate is not used.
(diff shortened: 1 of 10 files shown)


ablwr:
To see the differences more clearly, you can check out this branch locally and run `git diff origin --word-diff` to see the difference at the word-level, instead of sentence-level present in this WebUI. It makes it easier to review.

robUx4:
LGTM. I think the RFC might not have internal links at all anyway. If we want to have some back, we might as well do a proper pass on said document (likely after splitting).

ablwr:
Agreed. I can trim out the other few internal links that I demoted to just anchor-tags (but functional ones).

dericed:
An RFC can include links to section headers. I don't think other types of anchors can be used. Each MKV Element is given it's own section header when the whole doc is built so such anchors can be used. But when the RFC is rendered the link appears as a reference according to the section number. For instance in the EBML document

> `Master Elements` MUST NOT use an unknown size unless the `unknownsizeallowed` attribute of their `EBML Schema` is set to true (see [the section on the `unknownsizeallowed` attribute](#unknownsizeallowed)).

is rendered in the RFC as 

> "Master
>   Elements" MUST NOT use an unknown size unless the
 >  "unknownsizeallowed" attribute of their "EBML Schema" is set to true
  > (see Section 14.1.4.10).

chapters.md
```diff
@@ -27,13 +27,13 @@ The type of the menu system is defined by the ChapProcessCodecID parameter. For
 
 ### Matroska Script (0)
 
-This is the case when [ChapProcessCodecID]({{site.baseurl}}/index.html#ChapProcessCodecID) = 0\. This is a script language build for Matroska purposes. The inspiration comes from ActionScript, javascript and other similar scripting languages. The commands are stored as text commands, in UTF-8\. The syntax is C like, with commands spanned on many lines, each terminating with a ";". You can also include comments at the end of lines with "//" or comment many lines using "/* */". The scripts are stored in ChapProcessData. For the moment ChapProcessPrivate is not used.
+This is the case when ChapProcessCodecID = 0\. This is a script language build for Matroska purposes. The inspiration comes from ActionScript, javascript and other similar scripting languages. The commands are stored as text commands, in UTF-8\. The syntax is C like, with commands spanned on many lines, each terminating with a ";". You can also include comments at the end of lines with "//" or comment many lines using "/* \*/". The scripts are stored in ChapProcessData. For the moment ChapProcessPrivate is not used.
```
dericed: could consider backtick-quoting elements
> This is the case when ChapProcessCodecID = 0\.

to
> This is the case when the `ChapProcessCodecID Element` = 0\.

codec_specs.md
```diff
@@ -621,39 +621,39 @@ Codec ID: S_TEXT/UTF8
 
 Codec Name: UTF-8 Plain Text
 
-Description: Basic text subtitles. For more information, please look at the [Subtitle specifications]({{site.baseurl}}/subtitles.html).
+Description: Basic text subtitles. For more information, please look at the Subtitle section of the specification.
```
dericed: Seems pointed enough to be worth a sectional reference

> For more information, see \[the section on the \`Subtitles\`\](#subtitles).

ablwr:
Great, thanks @dericed! I'll make your recommended changes and this can go through review again.

ablwr:
It works! ✨

PR #151, opened 2017-07-16
adds make dependencies and instructions for readme

ablwr:
No related issue, but this updates the matroska-specification README with instructions to make it easier for new people to get involved and build/test locally.


robUx4:
LGTM

PR #152, opened 2017-07-17
cleans up html for webpage rendering, removes unneeded layouts

ablwr:
This, too, cleans up by removing some unused layout partials and straightens up the default and base layouts.


robUx4:
LGTM.

PR #153, opened 2017-07-23
Add tags as supported elements in webm.

nbirkbeck:
According to https://www.webmproject.org/docs/container/, the following elements are allowed: Tags, Tag, Targets, TargetTypeValue, TargetType, TagTrackUID, SimpleTag, TagName, TagLanguage, TagDefault, TagString, TagBinary.


mbunkus:
Thanks. But please re-format the commit message. The first line should be short (if possible no longer than ~75 characters), followed by an empty line, followed by arbitrary amounts of text but line-wrapped at ~75 characters. A good first line would be your first sentence.

Why? Because a lot of tools only show the first line as a summary, and long lines are messing up the overview.

Thanks.

nbirkbeck:
Thanks for quick review. I've amended the comment. 

Also, I guess a similar changes needs to go to foundation-source? (https://github.com/Matroska-Org/foundation-source)

mbunkus:
I think we simply copy over the whole XML file whenever we make new releases for libMatroska, hence a separate pull request against the foundation-source repository is not required.

robUx4:
LGTM merged !

mbunkus:
@nbirkbeck As @robUx4 has told me recently: we're not using the XML file from the "foundation-source" repository for library generation anymore. Instead we generate directly from `ebml_matroska.xml` in this repository. Just for the record.

PR #157, opened 2017-07-30
cleans up syntax and style for ordered_guidelines

ablwr:
Similar to https://github.com/Matroska-Org/matroska-specification/pull/156 but for ordered_guidelines.md.

Overall, I am wondering if the content in ordered_guidelines should be checked for duplication in the rest of the specification and then left out of the Makefile. It seems to me that most of this section is either repeated elsewhere (diagram.md -- with conflicting information, as I commented in the above PR: https://github.com/Matroska-Org/matroska-specification/pull/156#discussion_r130254663 ) **or** it contains recommendations not required to be in the official specification. Thoughts? @robUx4 @mbunkus @dericed ?

### DIFF
--- a/order_guidelines.md
+++ b/order_guidelines.md
@@ -3,43 +3,43 @@
 
 # Matroska Element Ordering Guidelines
 
-Except for the EBML Header and the CRC-32 Element, the EBML specification does not require any particular storage order for Elements. The Matroska specification however defines mandates and recommendations for ordering certain Elements in order to facilitate better playback, seeking, and editing efficiency. This section describes and offers rationale for ordering requirements and recommendations for Matroska.
+Except for the `EBML Header` and the `CRC-32 Element`, the EBML specification does not require any particular storage order for `Elements`. The Matroska specification however defines mandates and recommendations for ordering certain `Elements` in order to facilitate better playback, seeking, and editing efficiency. This section describes and offers rationale for ordering requirements and recommendations for Matroska.
 
 ## Top-Level Elements
 
-A valid Matroska file requires only one Top-Level Element, the `Info` Element; however, to be playable Matroska MUST also contain at least one `Tracks` and `Cluster` Element. The first `Info` Element and the first `Tracks` Element MUST either be stored before the first `Cluster` Element or both be referenced by a `SeekHead` Element which occurs before the first `Cluster` Element.
+The `Info Element` is the only REQUIRED `Top-Level Element` in a Matroska file. To be playable, Matroska MUST also contain at least one `Tracks Element` and `Cluster Element`. The first `Info Element` and the first `Tracks Element` MUST either be stored before the first `Cluster Element` or both SHALL be referenced by a `SeekHead Element` occurring before the first `Cluster Element`.
 
-After a Matroska file has been created it could still be edited. For example chapters, tags or attachments can be added. When new Top-Level Elements are added to a Matroska file the `SeekHead` Element(s) MUST be updated so that the `SeekHead` Element(s) itemize the identity and position of all Top-Level Elements. Editing, removing, or adding Elements to a Matroska file often requires that some existing Elements be voided or extended; therefore, it is RECOMMENDED to use Void Elements as padding in between Top-Level Elements.
+It is possible to edit a Matroska file after it has been created. For example, chapters, tags or attachments can be added. When new `Top-Level Elements` are added to a Matroska file, the `SeekHead` Element(s) MUST be updated so that the `SeekHead` Element(s) itemize the identity and position of all `Top-Level Elements`. Editing, removing, or adding `Elements` to a Matroska file often requires that some existing `Elements` be voided or extended; therefore, it is RECOMMENDED to use `Void Elements` as padding in between `Top-Level Elements`.
 
 ## CRC-32
 
-As noted by the EBML specification, if a `CRC-32` Element is used then the `CRC-32` Element MUST be the first ordered Element within its Parent Element. The Matroska specification recommends that `CRC-32` Elements SHOULD NOT be used as an immediate Child Element of the `Segment` Element; however all Top-Level Elements of an EBML Document SHOULD include a CRC-32 Element as a Child Element.
+As noted by the EBML specification, if a `CRC-32 Element` is used then the `CRC-32 Element` MUST be the first ordered `Element` within its `Parent Element`. The Matroska specification recommends that `CRC-32 Elements` SHOULD NOT be used as an immediate `Child Element` of the `Segment Element`; however all `Top-Level Elements` of an `EBML Document` SHOULD include a `CRC-32 Element` as a `Child Element`.
 
 ## SeekHead
 
-If used, the first `SeekHead` Element SHOULD be the first non-`CRC-32` Child Element of the `Segment` Element. If a second `SeekHead` Element is used then the first `SeekHead` MUST reference the identity and position of the second `SeekHead`, the second `SeekHead` MUST only reference `Cluster` Elements and not any other Top-Level Element already contained within the first `SeekHead`, and the second `SeekHead` MAY be stored in any order relative to the other Top-Level Elements. Whether one or two `SeekHead` Element(s) are used, the `SeekHead` Element(s) MUST collectively reference the identity and position of all Top-Level Elements except for the first `SeekHead` itself.
+If used, the first `SeekHead Element` SHOULD be the first non-`CRC-32 Child Element` of the `Segment Element`. If a second `SeekHead Element` is used, then the first `SeekHead Element` MUST reference the identity and position of the second `SeekHead`. Additionally, the second `SeekHead Element` MUST only reference `Cluster` Elements and not any other `Top-Level Element` already contained within the first `SeekHead Element`. The second `SeekHead Element` MAY be stored in any order relative to the other `Top-Level Elements.` Whether one or two `SeekHead Element(s)` are used, the `SeekHead Element(s)` MUST collectively reference the identity and position of all `Top-Level Elements` except for the first `SeekHead Element`.
 
-It is RECOMMENDED that the first `SeekHead` Element be followed by some padding (a `Void` Element) to allow for the `SeekHead` Element to be expanded to cover new Top-Level Elements that could be added to the Matroska file, such as `Tags`, `Chapters` and `Attachments` Elements.
+It is RECOMMENDED that the first `SeekHead Element` be followed by a `Void Element` to allow for the `SeekHead Element` to be expanded to cover new `Top-Level Elements` that could be added to the Matroska file, such as `Tags`, `Chapters` and `Attachments Elements`.
 
 ## Cues (index)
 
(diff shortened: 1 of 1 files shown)


mbunkus:
This is yet another remainder of the website conversion. In specs duplication should be avoided as much as possible. If the content is present in other places, then I'm in favor of removing this file. However, we should use that opportunity to improve upon the existing content or at least check if  `ordered_guidelines.md` contains good parts that could and should be copied to those other places.

mbunkus:
Oh, and therefore I won't review this just yet as it doesn't make sense to review if the file's going away anyway.

ablwr:
I propose the ordering guidelines be folded into the Matroska Structure section ("diagram.md") of the specification under each category and the "Optimum layout" section be dropped from the specification. Or possibly moved to diagram.md as well, under the Top Level Elements section.

order_guidelines.md
```diff
@@ -3,43 +3,43 @@
...
 
 ## Top-Level Elements
 
-A valid Matroska file requires only one Top-Level Element, the `Info` Element; however, to be playable Matroska MUST also contain at least one `Tracks` and `Cluster` Element. The first `Info` Element and the first `Tracks` Element MUST either be stored before the first `Cluster` Element or both be referenced by a `SeekHead` Element which occurs before the first `Cluster` Element.
+The `Info Element` is the only REQUIRED `Top-Level Element` in a Matroska file. To be playable, Matroska MUST also contain at least one `Tracks Element` and `Cluster Element`. The first `Info Element` and the first `Tracks Element` MUST either be stored before the first `Cluster Element` or both SHALL be referenced by a `SeekHead Element` occurring before the first `Cluster Element`.
```
robUx4: I'm confused with the backticking. For elements for which we use the EBML name, shouldn't we just backtick the EBML name? Anyway it doesn't appear in the XML, TXT or HTML output so it's not a big deal.

order_guidelines.md
```diff
@@ -3,43 +3,43 @@
...
 
-After a Matroska file has been created it could still be edited. For example chapters, tags or attachments can be added. When new Top-Level Elements are added to a Matroska file the `SeekHead` Element(s) MUST be updated so that the `SeekHead` Element(s) itemize the identity and position of all Top-Level Elements. Editing, removing, or adding Elements to a Matroska file often requires that some existing Elements be voided or extended; therefore, it is RECOMMENDED to use Void Elements as padding in between Top-Level Elements.
+It is possible to edit a Matroska file after it has been created. Chapters, tags or attachments can be added. When new `Top-Level Elements` are added to a Matroska file, the `SeekHead` Element(s) MUST be updated so that the `SeekHead` Element(s) itemize the identity and position of all `Top-Level Elements`. Editing, removing, or adding Elements to a Matroska file often requires that some existing Elements be voided or extended; therefore, it is RECOMMENDED to use `Void Elements` as padding in between `Top-Level Elements`.
```
robUx4: IMO the "For example" was important. For example I want to edit the aspect ratio of a video track without having to remux it.

order_guidelines.md
```diff
@@ -3,43 +3,43 @@
...
 
 ## Attachments
 
-The `Attachments` Element is not meant to use by default when playing the file, but could contain the cover art and/or fonts. Cover art is useful even before the file is played and fonts could be needed before playback starts for initialization of subtitles that could use them. The `Attachments` Element MAY be placed before the first `Cluster` Element; however if the `Attachments` Element is likely to be edited, then it SHOULD be placed after the last `Cluster` Element.
+The `Attachments Element` is not intended to used by default when playing the file, but could contain information relevant to the content, such as cover art or fonts. Cover art is useful even before the file is played and fonts could be needed before playback starts for initialization of subtitles. The `Attachments Element` MAY be placed before the first `Cluster Element`; however if the `Attachments Element` is likely to be edited, then it SHOULD be placed after the last `Cluster Element`.
```
robUx4: to /be/ used

order_guidelines.md
mbunkus: No, see the [EBML spec](https://github.com/Matroska-Org/ebml-specification/blob/master/specification.markdown). We not only include `Element` in the backticks, but we also backtick several important terms such as `Top-Level Element`. See "Notations and Conventions". To me this change looks like the syntax used in the EBML spec.

robUx4: OK

ablwr: My instinct and preference is also to just backtick the name and not add `Element` to the end of everything, but adding it outside of the quotes when we are using it, but, like @mbunkus, I went with what was already established in the EBML spec after asking @dericed about it.

order_guidelines.md
ablwr: I just added it back!

order_guidelines.md
ablwr: 👍

ablwr:
Maybe best to conclude review on this, and then open an issue about my proposal to integrate into the diagram section of the spec? I want to keep a separation of concerns (even though I brought it up in the first place! Slowly learning to not scope-creep.)

order_guidelines.md
```diff
@@ -3,43 +3,43 @@
...
 
 ## CRC-32
 
-As noted by the EBML specification, if a `CRC-32` Element is used then the `CRC-32` Element MUST be the first ordered Element within its Parent Element. The Matroska specification recommends that `CRC-32` Elements SHOULD NOT be used as an immediate Child Element of the `Segment` Element; however all Top-Level Elements of an EBML Document SHOULD include a CRC-32 Element as a Child Element.
+As noted by the EBML specification, if a `CRC-32 Element` is used then the `CRC-32 Element` MUST be the first ordered Element within its `Parent Element`. The Matroska specification recommends that `CRC-32 Elements` SHOULD NOT be used as an immediate `Child Element` of the `Segment Element`; however all `Top-Level Elements` of an EBML Document SHOULD include a `CRC-32 Element` as a `Child Element`.
```
dericed: `EBML Document` could be backticked.

order_guidelines.md
```diff
@@ -3,43 +3,43 @@
...
 
-It is RECOMMENDED that the first `SeekHead` Element be followed by some padding (a `Void` Element) to allow for the `SeekHead` Element to be expanded to cover new Top-Level Elements that could be added to the Matroska file, such as `Tags`, `Chapters` and `Attachments` Elements.
+It is RECOMMENDED that the first `SeekHead Element` be followed by a `Void Element` to allow for the `SeekHead` Element to be expanded to cover new `Top-Level Elements` that could be added to the Matroska file, such as `Tags`, `Chapters` and `Attachments Elements`.
```
dericed: one instance of `SeekHead Element` and one of `SeekHead` Element

order_guidelines.md
```diff
@@ -3,43 +3,43 @@
...
+The `Cues Element` is RECOMMENDED to optimize seeking access in Matroska. It is programmatically simpler to add the `Cues Element` after all `Cluster Elements` have been written because this does not require a prediction of how much space to reserve before writing the `Cluster Elements`. However, storing the `Cues Element` before the `Cluster Elements` can provide some seeking advantages. If the `Cues Element` is present, then it SHOULD either be stored before the first `Cluster Element` or be referenced by a `SeekHead Element`.
 
 ## Info
 
-The first `Info` Element SHOULD occur before the first `Tracks` and first `Cluster` Element.
+The first `Info Element` SHOULD occur before the first `Tracks Element` and first `Cluster Element`.
```
dericed: Redundant to line 10, but is missing the exception for when the Info Element is referenced by SeekHead.

order_guidelines.md
```diff
@@ -77,5 +77,4 @@ The `Tags` Element is the one that is most subject to changes after the file was
 
 ## Cluster Timecode
 
-As each `BlockGroup` and `SimpleBlock` of a `Cluster` Element needs the Cluster `Timecode`, the `Timecode` Element MUST occur as the first Child Element within the `Cluster` Element.
-
+As each `BlockGroup` and `SimpleBlock` of a `Cluster Element` needs the Cluster `Timecode Element`, the `Timecode Element` MUST occur as the first `Child Element` within the `Cluster Element`.
```
dericed: Could simplify Cluster `Timecode Element` to just `Timecode Element`.
Also there is a conflict here, when `Cluster` has a `CRC-32` I would expect it to be first, but this mandates that `Timecode` is first, thus with rules as written it is not possible to have a valid `Cluster Element` with a CRC.

order_guidelines.md
```diff
@@ -3,43 +3,43 @@
...
+The `Tags Element` is most subject to changes after the file was originally created. For easier editing, the `Tags Element` SHOULD be placed at the end of the `Segment Element`, even after the `Attachments Element`. On the other hand, it is inconvenient to have to seek in the `Segment` for tags, especially for network streams. So it's better if the `Tags Element` is found early in the stream. When editing the `Tags Element`, the original `Tags Element` at the beginning can be voided and a new one written at the end of the `Segment Element`. The file size will only marginally change.
```
dericed: `can be voided` may be presumes too much. Could say `overwritten with a Void Element` and `new one` to `new Tags Element`

ablwr:
Thanks, Dave. I updated (along with some other stray Elements needing backticks) except for this last bit about the conflicting issue here on Line 80.

ablwr:
Better to open up issue about conflict and cover it in a subsequent PR?

dericed:
Maybe open an issue or a cellar email or maybe @robUx4 or @mbunkus know the answer: In a `Cluster` with a `CRC-32` both `Timecode` and `CRC-32` are mandated to occur first, but what if both are there?

mbunkus:
I think that `Timecode` being the first element is just a recommendation, not a requirement. The intention is that the `Timecode` is found before the first `Element` that needs it (`BlockGroup` or `SimpleBlock`). Therefore placing the `CRC32 Element` first wouldn't do any harm.

dericed:
seems like a mandate:
> the `Timecode` Element MUST occur as the first Child Element within the `Cluster` Element.

Maybe should be:
> the `Timecode Element` MUST occur as in storage order before any `SimpleBlock`, `BlockGroup`, or `EncryptedBlock` within the `Cluster` Element.

Also perhaps we need a short name for: `SimpleBlock`, `BlockGroup`, or `EncryptedBlock`

ablwr:
Hmmmm... I feel that still reads weird, too. Trying to think of a better one though...

mbunkus:
> Maybe should be:

> > the Timecode Element MUST occur as in storage order before any SimpleBlock, BlockGroup, or EncryptedBlock within the Cluster Element.

I like it!

> Also perhaps we need a short name for: `SimpleBlock`, `BlockGroup`, or `EncryptedBlock`

You mean in order to address those three as a group without having to name each? Something like `Frame-Containing Elements` or… `Whale Elements` (because they're big and contain a lot of volume)?

ablwr:
I am going to update to @dericed's recc and force-push squashed commits and maybe we can get this merged in too!

dericed:
`🐳 Element`

ablwr:
Maybe we can continue this conversation on the CELLAR list? I can draft a post about it.

ablwr:
I think this is OK to merge!

PR #158, opened 2017-07-30
Adjust definition of "forced" and "default" flags to better reflect usage

rcombs:
This avoids implying user-hostile behavior that doesn't match actual
implementations (that I'm aware of), and is more in-line with the common
usage of "forced" subtitles for translations of parts of an audio or video
track that are in different languages than their primary language tag.

It also avoids requiring an overlay of multiple subtitle tracks. While this
was a good idea in theory (to avoid duplication of events between a "forced"
track and a normal one), it doesn't seem to have been implemented in players
or authoring software.

### DIFF
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -215,10 +215,10 @@
     <documentation lang="en">Set if the track is usable. (1 bit)</documentation>
   </element>
   <element name="FlagDefault" path="1*1(\Segment\Tracks\TrackEntry\FlagDefault)" cppname="TrackFlagDefault" id="0x88" type="uinteger" minOccurs="1" maxOccurs="1" minver="1" default="1" range="0-1">
-    <documentation lang="en">Set if that track (audio, video or subs) SHOULD be active if no language found matches the user preference. (1 bit)</documentation>
+    <documentation lang="en">For audio or video, set if that track SHOULD be active if no language found matches the user's preference; this can be used to specify the original content language. For subtitles, set if that track SHOULD be chosen if there are multiple tracks which match the user's language preferences. (1 bit)</documentation>
   </element>
   <element name="FlagForced" path="1*1(\Segment\Tracks\TrackEntry\FlagForced)" cppname="TrackFlagForced" id="0x55AA" type="uinteger" minOccurs="1" maxOccurs="1" minver="1" default="0" range="0-1">
-    <documentation lang="en">Set if that track MUST be active during playback. There can be many forced track for a kind (audio, video or subs), the player SHOULD select the one which language matches the user preference or the default + forced track. Overlay MAY happen between a forced and non-forced track of the same kind. (1 bit)</documentation>
+    <documentation lang="en">Applies only to subtitles. Set if that track SHOULD be active during playback by default if it matches the user's language preference, even if the user's preferences would normally not enable subtitles with the selected audio track; this can be used for tracks containing only translations of foreign-language audio or onscreen text. (1 bit)</documentation>
   </element>
   <element name="FlagLacing" path="1*1(\Segment\Tracks\TrackEntry\FlagLacing)" cppname="TrackFlagLacing" id="0x9C" type="uinteger" minOccurs="1" maxOccurs="1" minver="1" default="1" range="0-1">
     <documentation lang="en">Set if the track MAY contain blocks using lacing. (1 bit)</documentation>
--- a/notes.md
+++ b/notes.md
@@ -191,13 +191,13 @@ Some general notes for a program:
 
 ## Default flag
 
-The "default track" flag is a hint for the playback application and SHOULD always be changeable by the user. If the user wants to see or hear a track of a certain kind (audio, video, subtitles) and she hasn't chosen a specific track then the player SHOULD use the first track of that kind whose "default track" flag is set to "1". If no such track is found then the first track of this kind SHOULD be chosen.
+The "default track" flag is a hint for the playback application and SHOULD always be changeable by the user. If the user wants to see or hear a track of a certain kind (audio, video, subtitles) and she hasn't chosen a specific track (either explicitly, or via a language preference), then the player SHOULD use the first track of that kind whose "default track" flag is set to "1". If no such track is found then the first track of this kind SHOULD be chosen.
 
 Only one track of a kind MAY have its "default track" flag set in a segment. If a track entry does not contain the "default track" flag element then its default value "1" is to be used.
 
 ## Forced flag
 
-The "forced" flag tells the playback application that it MUST display/play this track or another track of the same kind that also has its "forced" flag set. When there are multiple "forced" tracks, the player SHOULD determined based upon the language of the forced flag or use the default flag if no track matches the use languages. Another track of the same kind without the "forced" flag may be use simultaneously with the "forced" track (like DVD subtitles for example).
+If a track with the "forced" flag whose language matches the user's preference exists, and the user's preferences would normally indicate that subtitles would not be displayed with the selected audio track, the "forced" track SHOULD be displayed by default.
 
 # TrackTimecodeScale
 


ablwr:
+1 for friendlier docs.

dericed:
I feel hesitant about the change in the default value, since many existing files would suddenly change in behavior. That change could be reduced with a statement to say that only the first FlagDefault (or whatever order is preferred) should be used. I suspect that it closer to most common behaviors, that the first track is defaulted while the subsequent default tracks aren't.

rcombs:
The default value change is pretty tangential; I did that because reading this section:

```
Only one track of a kind MAY have its "default track" flag set in a segment. If a track entry does not contain the "default track" flag element then its default value "1" is to be used.
```

Made the default of "1" seem erroneous. If no tracks are set as "default", player behavior is the same as if they all are, so if I was designing a new format, I'd set "0" as the default value and allow writing implementations to only emit the flag when explicitly requested.

But yeah, I see your point re: compatibility; re-pushing without that change.

mbunkus:
There are several elements whose default values would have been chosen differently if we were to redesign the format with today's knowledge. `FlagDefault` is one of those, as is `Language` (English as the default language doesn't make much sense when there's `und` = "undefined" in the specs which makes much more sense). But alas, we cannot screw existing files retroactively by changing default values.

robUx4:
IMO this should be 2 separate Pull Request. One for "default" and one for "forced". One could be apply while the other is not.

ebml_matroska.xml
```diff
@@ -215,10 +215,10 @@
...
   </element>
   <element name="FlagForced" path="1*1(\Segment\Tracks\TrackEntry\FlagForced)" cppname="TrackFlagForced" id="0x55AA" type="uinteger" minOccurs="1" maxOccurs="1" minver="1" default="0" range="0-1">
-    <documentation lang="en">Set if that track MUST be active during playback. There can be many forced track for a kind (audio, video or subs), the player SHOULD select the one which language matches the user preference or the default + forced track. Overlay MAY happen between a forced and non-forced track of the same kind. (1 bit)</documentation>
+    <documentation lang="en">Applies only to subtitles. Set if that track SHOULD be active during playback by default if it matches the user's language preference, even if the user's preferences would normally not enable subtitles with the selected audio track; this can be used for tracks containing only translations of foreign-language audio or onscreen text. (1 bit)</documentation>
```
robUx4: No, this can't be subtitle only. That's the most common case. But we don't want to limit the format. For example you could have a forced video track. Some parts of the main video would be overlayed to show signs in the local language or censor parts in a specific country.

ebml_matroska.xml
```diff
@@ -215,10 +215,10 @@
     <documentation lang="en">Set if the track is usable. (1 bit)</documentation>
   </element>
   <element name="FlagDefault" path="1*1(\Segment\Tracks\TrackEntry\FlagDefault)" cppname="TrackFlagDefault" id="0x88" type="uinteger" minOccurs="1" maxOccurs="1" minver="1" default="1" range="0-1">
-    <documentation lang="en">Set if that track (audio, video or subs) SHOULD be active if no language found matches the user preference. (1 bit)</documentation>
+    <documentation lang="en">For audio or video, set if that track SHOULD be active if no language found matches the user's preference; this can be used to specify the original content language. For subtitles, set if that track SHOULD be chosen if there are multiple tracks which match the user's language preferences. (1 bit)</documentation>
```
robUx4: We can't say it could mean something else too (original content), That would be another flag and you can't have data that are left to interpretation depending on what you think the file means.
I don't see why subtitles should be treated differently than audio or video.

notes.md
```diff
@@ -191,13 +191,13 @@ Some general notes for a program:
...
 
 Only one track of a kind MAY have its "default track" flag set in a segment. If a track entry does not contain the "default track" flag element then its default value "1" is to be used.
 
 ## Forced flag
 
-The "forced" flag tells the playback application that it MUST display/play this track or another track of the same kind that also has its "forced" flag set. When there are multiple "forced" tracks, the player SHOULD determined based upon the language of the forced flag or use the default flag if no track matches the use languages. Another track of the same kind without the "forced" flag may be use simultaneously with the "forced" track (like DVD subtitles for example).
+If a track with the "forced" flag whose language matches the user's preference exists, and the user's preferences would normally indicate that subtitles would not be displayed with the selected audio track, the "forced" track SHOULD be displayed by default.
```
robUx4: A forced flag is really forced. It can't be overridden. Only a different flavor of a forced flag could replace it. And if there are many forced flags, then the one with the default flag takes precedence.

ebml_matroska.xml
rcombs: This sounds nice in theory, but in practice there are no implementations, and I heavily doubt there ever will be, even if this were clarified in the spec. There's some work to build such a system within subtitle tracks, which fits much better into player pipelines.

notes.md
rcombs: This is both user-hostile and inconsistent with real implementations and existing file intent. What's the goal?

ebml_matroska.xml
rcombs: An alternative I'd previously suggested in an email was for both track types to follow the definition given for subtitles, and separately to recommend that the first track of each type be the content's original language. I can switch this PR to that if you prefer.

notes.md
rcombs: Ping on this; I'd like to get this PR updated, but I'm not sure how to address this comment. No real players treat "forced" as not being user-overridable (including with a non-forced track), and I don't think there's any appetite to change that.

mcr:
Not happy with the "forced" flag definition, but were happy with the rest.
If this could be rebased, it would be adopted.

rcombs:
Rebased (after a long delay, sorry!); hoping for a bit more discussion. I've just been paging this all back in, and I'm going to dump some thoughts in here:

There are really 4 questions the spec has to answer here; 2 for each of the 2 flags:
- How should players make use of them, for optimal compatibility with the intent of both existing files and new ones?
- How should file authors set them, in order to achieve optimal behavior for users?

And there are a few important caveats to keep in mind:
- These flags are handled in players at a level well above the Matroska demuxer itself. They generally don't implement Matroska-specific behavior for them, and aren't likely to start after a spec change. Anything we recommend here has to be applicable to any format with default or forced flags, and has to provide as good or better a user experience than the ad-hoc behaviors players already have, or it's just going to be ignored.
  - As a corollary, these flags can't invent entirely new types of behavior (like the current "overlay two tracks together" suggestion). Anything unusual and complex like that would need to be defined as a distinct new field.
- Any advice we give file authors on how to use these flags is only going to be followed if doing so results in the behavior those authors want to achieve in the players their target audience uses. If following the spec's advice doesn't do what they want, they'll ignore it and come up with some method that does.

Now, musings on the individual flags:

## The default flag
The "default" flag has always been a little strange, in that if there's only one track with the flag set, it doesn't really provide much value that you can't achieve simply by putting a track first in the list. There are a few possible uses for it (and in practice, usage is inconsistent):
- Indicating which track should be used if there's no language setting
  - Essentially the same as putting the track first
  - Usually would be either a track in the original language, or one in the file author's/primary target audience's
- Indicating which track, out of multiple for a given language, should be used
  - Could be used to indicate a "normal" track, as opposed to a commentary, descriptive audio, forced subtitle, SDH, etc. track; this can have actual value beyond what you get from simple ordering, as it could be used on multiple tracks (e.g. a stereo and a 5.1) to indicate "by default, when selecting a track of this language, pick the one with the default flag set that best matches the output channel layout"
  - Sometimes might refer to a track that the file author considers preferable to others that seem, to a machine, equivalent (e.g. multiple different translations or dubs in the same language); this again isn't much different than putting the recommended one first

In practice both of these uses exist, plenty of files simply don't set a default (or set it on all tracks), and it's somewhat impractical to depend on in player software.

So, I think the best algorithm to recommend for players (for compat with all reasonable cases in existing files as well as new ones), written out in full, is:
- If no audio track matching the user's language preferences is present, select the language of the first audio track with the default flag set, or if none is present, the language of the first audio track.
- If multiple audio tracks matching the chosen language (set either by the user or in the previous step) are present, autoselect between the ones that have the default flag set. If none, use the first of the group.
- If multiple subtitle tracks matching the chosen language (either from the user's preferences or an optional fallback of undefined) are present, select the first one with the default flag set. If none, use the first of the group.

And for advice to file authors:
- Place tracks in the source language before those of the same type in other languages
- Within each group of tracks of the same language and type, place recommended tracks (e.g. preferred translation, dub, source, etc) first.
- Set the default flag on all tracks that you want to be available to autoselection without special circumstances; i.e. non-commentary, non-SDH, non-descriptive, non-forced (all of which have their own additional flags, though SDH doesn't have one in Matroska yet as far as I'm aware)

## The forced flag
In practice, the forced flag has clear meaning, use, and implementation:
- It signals a track that should be played, even if the user's settings normally wouldn't enable subtitles with the selected audio track.
- It's used on tracks containing translations of onscreen text, foreign-language dialogue, music left in the original language in an otherwise-dubbed audio track, or other text that needs to be displayed onscreen even if the user can hear and understand a corresponding audio track.
  - Note that in practice, file authors duplicate this data in both a forced and a regular track, rather than rely on the player to display both the forced track and the regular track.
  - And in practice, players don't implement support for selecting both a forced and a regular track simultaneously and displaying both.
- It's implemented by something along the lines of "when an audio track matching the user's language preferences is selected, autoselect a forced track even if the user doesn't have subtitles turned on for all content by default".

In practice the flag is also used very inconsistently, often on tracks where setting "default" would've made more sense, but usually in ways that aren't especially harmful in practice. My suggestion is that we simply codify that existing player behavior, and recommend that existing usage to file authors.

ebml_matroska.xml
robUx4: But if the specs says now *only* subtitles can be forced. It will never ever be possible to force other types of tracks because it will break backward compatibility. That's why the specs are kept open on purpose.

ebml_matroska.xml
robUx4: That sounds better, yes.

notes.md
robUx4: Sorry for the huge delay.
"forced" tracks are user hostile, yes. It can happen on some DVDs where they force you to see a certain subtitle. I also noticed on iTunes some movies have forced subtitles, you can't turn them off.

It can also be nice in some cases. For example if in the movie a person talks an alien language that will be shown as a subtitle even though you don't have subtitles one. But that raises another subject. In what language should it be displayed. In that case there could be multiple variants of the same (klingon) subtitle track. Each of them should have the forced flag. That's where this section comes from to pick the best one.

robUx4:
@rcombs can you rebase your branch on current master and push force in your PR branch ? I can't seem to be able to reopen the issue: The forced-tracks branch was force-pushed or recreated.
Rebasing works without any edit needed (tried here).

If we can't reopen the issue we might as well continue the discussion in #374 and do a new PR when there's a consensus on what do to.

notes.md
rcombs: > It can happen on some DVDs where they force you to see a certain subtitle.

Sure, but Matroska isn't DVD and I don't think there's any value in trying to imitate it here.

> I also noticed on iTunes some movies have forced subtitles, you can't turn them off.

This is probably implemented using the "forced" flag in HLS, which is also somewhat poorly-defined; even within Apple there are multiple different implementations with different behaviors (iTunes vs Safari vs QuickTime Player). But there's nothing stopping an end-to-end system like iTunes from preventing the user from disabling forced tracks, even if the spec didn't explicitly require it.

> It can also be nice in some cases. For example if in the movie a person talks an alien language that will be shown as a subtitle even though you don't have subtitles one. 

Right, and that's how the flag is used in practice (when it's used for anything meaningful at all, anyway).

> In what language should it be displayed. In that case there could be multiple variants of the same (klingon) subtitle track. Each of them should have the forced flag.

Yeah, that's what I'm trying to specify; the appropriate forced track for the user's selected language should be selected _by default_ by the player, unless the user's subtitle preferences cause a non-forced track to be used instead. I take issue with (a) making it a requirement that they can't be turned off, which no player implements and I expect no mainstream player ever will, and (b) specifying that multiple tracks should be superimposed, which no player implements and no player _should_ implement (since it would break essentially all existing files that use forced streams in this way).

I'd really like the spec to actually reflect reality here, and this "forced means literally forced" thing doesn't. A "MUST" in the spec is meaningless if every implementation decides to ignore it, and I can tell you that mpv and Plex certainly aren't going to implement that behavior. Along with how it'd result in unhappy users for no benefit, it would break a decent number of files that already exist in the wild and use the "forced" flag on regular subtitle streams.

rcombs:
Rebased, but, huh, apparently if you force-push a branch while a PR for it is closed, it can't be reopened. Funky; I guess I'll make a new one once we have consensus, yeah.

ebml_matroska.xml
rcombs: My issue is that the spec doesn't define anything useful that should happen when there's a forced audio or video track (I know it currently says "overlay MAY happen" but that clause might as well not exist seeing as nothing implements it).
Maybe we could say that authoring applications currently MUST not set this flag to 1 for video and audio tracks, and that player applications SHOULD ignore it, while leaving open the possibility that some meaning could be assigned to it in the future? (I'm just not sure what you would use it for that wouldn't be better-served by adding a new flag that doesn't have a the historical baggage.)

robUx4:
It seems [possible to reopen the issue](https://github.com/isaacs/github/issues/361#issuecomment-561158372), but you have to put your branch in the state it was when the branch was created. So if you rebase your branch on top of 868f024cbd41559e0ef35fb42aa7343706bc3724 and force push it, it might be possible...

rcombs:
Hmmmm, I've reverted the branch to its original state (on top of 868f024); can you reopen this now? (I'll force-push the latest rebased version afterwards)

robUx4:
No :(

mbunkus:
Just open a new PR and reference this one for those who'd like to read the original discussion.

PR #161, opened 2017-08-09
fix libMatroska class names of certain elements

mbunkus:
The new script "tools/verify_libmatroska_class_names.rb" has turned up
several <element>s in the spec where the class name given in the
spec (either via the 'cppname' attribute or, if that isn't present,
the 'name' attribute) dose not appear in any of libMatroska's header
files.

This commit fixes all six of them.

### DIFF
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -7,10 +7,10 @@
   <element name="Segment" path="1*1(\Segment)" id="0x18538067" type="master" unknownsizeallowed="1" minOccurs="1" maxOccurs="1" minver="1">
     <documentation lang="en">The Root Element that contains all other Top-Level Elements (Elements defined only at Level 1). A Matroska file is composed of 1 Segment.</documentation>
   </element>
-  <element name="SeekHead" path="0*2(\Segment\SeekHead)" cppname="SeekHeader" id="0x114D9B74" type="master" maxOccurs="2" minver="1">
+  <element name="SeekHead" path="0*2(\Segment\SeekHead)" id="0x114D9B74" type="master" maxOccurs="2" minver="1">
     <documentation lang="en">Contains the Segment Position of other Top-Level Elements.</documentation>
   </element>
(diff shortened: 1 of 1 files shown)


ebml_matroska.xml
```diff
@@ -944,7 +944,7 @@
   <element name="TagName" path="1*1(\Segment\Tags\Tag\SimpleTag\TagName)" id="0x45A3" type="utf-8" minOccurs="1" maxOccurs="1" minver="1" webm="1">
     <documentation lang="en">The name of the Tag that is going to be stored.</documentation>
   </element>
-  <element name="TagLanguage" path="1*1(\Segment\Tags\Tag\SimpleTag\TagLanguage)" id="0x447A" type="string" minOccurs="1" maxOccurs="1" minver="1" webm="1" default="und">
+  <element name="TagLanguage" path="1*1(\Segment\Tags\Tag\SimpleTag\TagLanguage)" cppname="TagLangue" id="0x447A" type="string" minOccurs="1" maxOccurs="1" minver="1" webm="1" default="und">
```
dericed: typo?

mbunkus: In a matter of sense, yes — it was a typo back when the original libMatroska implementation was written. The specs (even on the web page) came later and had to incorporate the existing class names.

Changing it now would break existing implementations.

We could consider using C++ type aliases for such cases; in C++11 syntax: `using KaxTagLanguage = KaxTagLangue` (or the other way around and rename the class), but I haven't fully thought about how that'll affect existing code, or if at all.

For now I'd just like to make the specs match existing source code out there.

mbunkus:
I've extended the new [libmatroska verification script](https://github.com/Matroska-Org/matroska-specification/blob/master/tools/verify_libmatroska_class_names.rb) to check the mapping between element IDs ↔ libMatroska classes, too. The good news is that everything matches up nicely.

ebml_matroska.xml
dericed: in that case +1 for `Langue`

robUx4: +1 for Frenchism. I had to leave my mark somewhere.

robUx4:
LGTM

PR #162, opened 2017-08-12
correct FieldOrder definitions

dericed:
the definitions were originally based on QuickTime’s specification
which is acknowledged by an Apple engineer to be wrong. This corrects
the definitions to fit current practice with ffmpeg. See
https://github.com/amiaopensource/vrecord/issues/170 for the discussion
that started this.

### DIFF
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -318,17 +318,17 @@
     <restriction>
       <enum value="0" label="progressive"/>
       <enum value="1" label="tff">
-        <documentation lang="en" type="definition">Interlaced with top field display first and top field stored first.</documentation>
+        <documentation lang="en" type="definition">Top field displayed first. Top field stored first.</documentation>
       </enum>
       <enum value="2" label="undetermined"/>
       <enum value="6" label="bff">
-        <documentation lang="en" type="definition">Interlaced with bottom field displayed first and bottom field stored first.</documentation>
+        <documentation lang="en" type="definition">Bottom field displayed first. Bottom field stored first.</documentation>
       </enum>
       <enum value="9" label="bff(swapped)">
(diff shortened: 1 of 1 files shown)


ablwr:
Thanks for your work!

mbunkus:
FYI: I'd prefer to have feedback from @robUx4 before merging.

JeromeMartinez:
IMHO it is better to wait a couple of days for comments from anyone before merging.

dericed:
ping to @aconverse who IIUC originally added the BT / TB notations in https://github.com/FFmpeg/FFmpeg/commit/4bf3c8f226252e18de8051fd0d417c1d39857b67.

aconverse:
It's been forever since I've dealt with interlacing. IIRC I was using [icefloe019 ](http://mirror.informatimago.com/next/developer.apple.com/quicktime/icefloe/dispatch019.html#fiel)as my reference. 

Your prose seems to better agree with icefloe019 than my comments do.

robUx4:
I honestly don't remember where these values come from. Given they are oddly precise and not continuous it's probably borrowed from somewhere else. So the Quicktime source is likely here.
I'll follow @mbunkus take on this because mkvmerge is probably the first one who output interlaced content. So I think it's the reference.

mbunkus:
mkvmerge supports the field. Its man page documents the allowed values which is basically just a copy & paste job from the specs. However, I support changing the wording in the specs and risking old Matroska files having their meaning altered slightly, simply because I consider a system that's clearly copied from QuickTime but that doesn't match how QuickTime works to be insane. It makes it very confusing for people to use (the people who will be using the field order are the ones that actually do have a lot of domain knowledge and probably know QuickTime's system).

PR #163, opened 2017-08-13
bumps subtitle-related information under subtitles section

ablwr:
Proposal to congregate all subtitles information under one main Subtitles section, instead of having different subtitles as their own top-level section.

Right now, the built specification has each taking up a line, like:

```
32. Subtitles . . . . . . . . . . . . . . . . . . . . . . . . . . 174
   33. Images Subtitles  . . . . . . . . . . . . . . . . . . . . . . 174
   34. SRT Subtitles . . . . . . . . . . . . . . . . . . . . . . . . 177
   35. SSA/ASS Subtitles . . . . . . . . . . . . . . . . . . . . . . 178
   36. USF Subtitles . . . . . . . . . . . . . . . . . . . . . . . . 181
   37. WebVTT  . . . . . . . . . . . . . . . . . . . . . . . . . . . 181
```
..etc. 

This would nest them all under section 32.

PR #164, opened 2017-08-13
Menu and streaming nesting

ablwr:
Similar to #163 but for Menu and Streaming sections, although there seems to be a bug present in master while building the Menu section, causing the final paragraph to be considered one of the headers. I've messed around with trying to sort this out, but no luck yet. Anyway, shouldn't block this small structural update.

### DIFF
--- a/menu.md
+++ b/menu.md
@@ -3,19 +3,19 @@
 
 # Menu Specifications
 
-# Introduction
+## Introduction
 
 This document is a _draft of the Menu system_ that will be the default one in Matroska. As it will just be composed of a Control Track, it will be seen as a "codec" and could be replaced later by something else if needed.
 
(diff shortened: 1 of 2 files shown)


ablwr:
What's the justification in squashing? Each commit covers each section!

mbunkus:
They're both trivial and do the same thing, just for different sections. In my opinion "next menu and streaming items" would suffice as a description.

I'm just not a friend of having too many commits. It can make working with the history more difficult than necessary. Yes, one commit for one logical/semantic change, and yes, this is debatable in this case.

I'd be fine with keeping both commits, though. I'm not that dogmatic.

ablwr:
I squashed! And I will return to tidy up both of these sections.

mbunkus:
I didn't mean to imply that you should squash it. It was intended as a reminder for whoever merges it to use github's "squash and merge" functionality. But thanks anyway 😄

Issue #165, opened 2017-08-16
Makefile not building due to `<` error


ablwr:
I noticed I couldn't locally build the RFC. I narrowed it down to the merging in of this commit https://github.com/Matroska-Org/matroska-specification/commit/0250d601d8967e713671c2b2486accae75435ee9 refactoring ordered_guidelines. There's nothing wrong with the update in theory, but rather the Makefile is determined to occasionally turn the last paragraph of each file into a section heading. WHY!?

So the last sentence or ordered_guidelines.md was set up to quote the elements like so:
```
The `Timecode Element` MUST occur as in storage order before any `SimpleBlock`, `BlockGroup`, or `EncryptedBlock` within the `Cluster Element`.
```

Which causes the break. The same occurs if modifying the last line in notes.md, which precedes it.

Working on sorting out this bug since I unintentionally caused it to rise up!

ablwr:
I don't know why it is doing this, but I found out by adding the same as the empty front-loaded YAML, the dashes, like so:
```
---
---
```

it resolves the issue.

ablwr:
(Streaming.md does this too)

ablwr:
This problem doesn't appear in the latest draft on CELLAR though, so I'm not sure what changed to cause it to pop up in the meantime.

ablwr:
I think this started when @robUx4 took the default: layout out of the files. This was good, but took a little too much out of the Makefile.

If I return `	cat $^ | grep -v '^---' > $@` to line 13, it gets rid of this problem by clearing out these pointless lines.

Ah! It just made sense to me, because there are two ways to make header tags in Markdown: Using `# This` syntax or using
```
This
---
```
syntax! Mystery solved.

ablwr:
The --- are still annoying but I am going to ask the Makefile to take them out for the time being so we have a clean directory structure.

robUx4:
Fix merged.

PR #166, opened 2017-08-16
clears yaml lines causing a distruption in RFC building

ablwr:
You can watch me basically live-debug the reason behind this in #165 😊


robUx4:
LGTM

PR #168, opened 2017-08-19
removes line of advice for player implementation

ablwr:
This is good advice, but I don't think it fits in the spec. If it does fit in the spec, I don't think it fits here, not in a category, at the beginning of the notes section (which immediately follows the schema structure details).

### DIFF
--- a/notes.md
+++ b/notes.md
@@ -1,8 +1,6 @@
 ---
 ---
 
-If you intend to implement a Matroska player, make sure you can handle all the files in [our test suite](http://www.matroska.org/downloads/test_w1.html), or at least the features presented there, not necessarily the same codecs.
-
 # Beginning of File
 
 An EBML file always starts with 0x1A. The 0x1A makes the DOS command "type" ends display. That way you can include ASCII text before the EBML data and it can be displayed. The EBML parser is safe from false-alarm with these ASCII only codes.


retokromer:
Will there be a document on player implementation?

ablwr:
There aren't plans to make one as part of standardization. There could be an independent guide... but the spec itself should be clear enough that it isn't necessary.

retokromer:
While working on FLAC people showed up and give us useful hints, in order to avoid the mistakes they made. I was thinking about something like that, in addition to the specs.

ablwr:
@mbunkus is the expert!

retokromer:
Therefore he is an excellent candidate for writing such a guide ;-)

mbunkus:
Not really. I've never written Matroska for a media player or a media player in general myself. I'm much more knowledgeable regarding creating Matroska files.

The only real advice I have is… don't forget about default values.

JeromeMartinez:
+1 for removing this part from spec, not the right place.

I see the same kind of sentence on https://matroska.org/technical/specs/notes.html (the first paragraph) , isn't it what is expected?

mbunkus:
The thing with expectations is that everyone has different ones. Stating that there is a test suite and that people should use it is a good thing. However, such advice has no place in an RFC.

retokromer:
It has definitely nothing to do in the specifications. I guess we all agree on this.

However I would personally like to see it as an additional document. Not just to put it into the rubbish bin. That‘s my only point.

ablwr:
Perhaps you want to bring this up on the CELLAR list instead of this not-well-viewed pull request?

retokromer:
@ablwr Done.

PR #173, opened 2017-08-25
arranges notes section

ablwr:
Okay, this is kind of a big PR.

This seeks to...

- Re-arrange Notes.md to give a more cohesive structure: Creating top-level Segment Position, Linked Segments, Track Flags, and Timecodes sections. Some of these were scattered around and I put them all under the same semantically relevant grouping. Other notes that don't fit into the subcategories remain at the top-level. Unfortunately this makes the diff a little hard to read.

- Update grammar and syntax, as I've been doing.

- Raise a few questions that I'll comment in-line

### DIFF
--- a/notes.md
+++ b/notes.md
@@ -3,27 +3,33 @@
 
 # Beginning of File
 
-An EBML file always starts with 0x1A. The 0x1A makes the DOS command "type" ends display. That way you can include ASCII text before the EBML data and it can be displayed. The EBML parser is safe from false-alarm with these ASCII only codes.
+An EBML file MUST start with `0x1A`. The 0x1A makes the DOS command "type" ends display. Therefore, ASCII text can be included before the EBML data and it can be displayed. After `0x1A`, the EBML header is stored. This allows the the parser to know what type of EBML file it is parsing.
 
-Next the EBML header is stored. This allows the the parser to know what type of EBML file it is parsing.
+# Mime Types
 
-# Block Timecodes
+There is no IETF-endorsed MIME type for Matroska files. But you can use the ones we have defined on our web server:
 
-The Block's timecode is signed integer that represents the Raw Timecode relative to the Cluster's Timecode, multiplied by the TimecodeScale (see the TimecodeScale notes.
+* .mka : Matroska audio `audio/x-matroska`
+* .mkv : Matroska video `video/x-matroska`
+* .mk3d : Matroska 3D video `video/x-matroska-3d`
 
-The Block's timecode is represented by a 16bit signed integer (sint16). This means that the Block's timecode has a range of -32768 to +32767 units. When using the default value of TimecodeScale, each integer represents 1ms. So, the maximum time span of Blocks in a Cluster using the default TimecodeScale of 1ms is 65536ms.
+# Unknown elements
 
-If a Cluster's Timecode is set to zero, it is possible to have Blocks with a negative Raw Timecode. Blocks with a negative Raw Timecode are not valid.
+Matroska is based upon the principle that a reading application does not have to support 100% of the specifications in order to be able to play the file. A Matroska file therefore contains version indicators that tell a reading application what to expect.
 
-# Default decoded field duration
+It is possible and valid to have the version fields indicate that the file contains Matroska `Elements` from a higher specification version number while signaling that a reading application MUST only support a lower version number properly in order to play it back (possibly with a reduced feature set). For example, a reading application supporting at least Matroska version `V` reading a file whose `DocTypeReadVersion` field is equal to or lower than `V` MUST skip Matroska/EBML `Elements` it encounters but does not know about if that unknown element fits into the size constraints set by the current `Parent Element`.
 
-The `DefaultDecodedFieldDuration` Element can signal to the displaying application how often fields of a video sequence will be available for displaying. It can be used for both interlaced and progressive content.
+# Default Values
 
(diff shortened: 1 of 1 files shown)


notes.md
```diff
@@ -3,27 +3,33 @@
 
 # Beginning of File
 
-An EBML file always starts with 0x1A. The 0x1A makes the DOS command "type" ends display. That way you can include ASCII text before the EBML data and it can be displayed. The EBML parser is safe from false-alarm with these ASCII only codes.
+An EBML file MUST start with `0x1A`. The 0x1A makes the DOS command "type" ends display. Therefore, ASCII text can be included before the EBML data and it can be displayed. After `0x1A`, the EBML header is stored. This allows the the parser to know what type of EBML file it is parsing.
```
ablwr: Is this section necessary for the specification? Can it be remove as "out of scope"?

notes.md
```diff
@@ -3,27 +3,33 @@
...
 
-Next the EBML header is stored. This allows the the parser to know what type of EBML file it is parsing.
+# Mime Types
```
ablwr: What will the status of this section be? Will it stay as it is, or should we update it?

notes.md
retokromer: At least
>An EBML file MUST start with `0x1A`.

is necessary here, in my opinion. Yet I think more does not hurt.

ablwr: To follow this up, if it should be here, I think it should move somewhere, possibly the top of the Ordered Guidelines section.

dericed: This should be reviewed carefully as removing it would invalidate potential files in the wild. Personally I am not supportive of the ASCII before EBML header idea. From this sentence it seems like having 10 gigabytes of 0x20 bytes followed by a Matroska file would be a valid Matroska file. If this section is kept I'd suggest it be limited a reasonable number of bytes. I think MXF is a bit similar where it can start with junk before the header but I hope they apply a limit to it.

[after some testing]. I tried appending a 0x20 to the beginning of Matroska and then it fails to play in VLC or ffmpeg. @robUx4 is there a particular use case planned with this ASCII feature or could we drop this?

ablwr: My interest in removal has to do with it being a rule for EBML (and moved to that spec, if not already written or otherwise stood there), and not Matroska specifically.

ablwr: If kept (or moved), maybe language should change from "can be included" to SHOULD NOT -- dictating that it would not technically be invalid if done, but is not recommended.

dericed: I'd vote no on moving it to EBML. This feature was historically only listed in Matroska specs and not referenced in EBML. I'd prefer not to allow this feature to infect other EBML based formats.

ablwr:
I will resolve the above conflict. What else can I do to nudge this PR along?

notes.md
mbunkus: Definitely not move it to EBML, please. I'd rather just remove that section. I've never seen such a file in the wild (by which I don't mean that they don't exist, of course). Right now seems a good opportunity to remove such cruft.

notes.md
mbunkus: As we're talking about [that mail extension thingy](https://en.wikipedia.org/wiki/MIME) and not about different types of actors, we should write it all-uppercase.

notes.md
```diff
@@ -3,27 +3,33 @@
...
+# Default Values
 
-# Default decoded field duration
+The default value of an `Element` is assumed when not present in the data stream. It is assumed only in the scope of its `Parent Element`. For example, the `Language Element` is in the scope of the `Track Element`. If the `Parent Element` is not present or assumed, then the `Child Element` cannot be assumed.
 
-The `DefaultDecodedFieldDuration` Element can signal to the displaying application how often fields of a video sequence will be available for displaying. It can be used for both interlaced and progressive content.
+## DefaultDecodedFieldDuration
 
-If the video sequence is signaled as interlaced, then the period between two successive fields at the output of the decoding process equals DefaultDecodedFieldDuration.
+The `DefaultDecodedFieldDuration Element` can signal to the displaying application how often fields of a video sequence will be available for displaying. It can be used for both interlaced and progressive content. If the video sequence is signaled as interlaced, then the period between two successive fields at the output of the decoding process MUST equal `DefaultDecodedFieldDuration`.
```
mbunkus: Adding a `MUST` here would invalidate any file that includes this header element and has some kind of gap in the video. I don't think this is wise. It's called a "default" for a reason, meaning that it applies to the (vast) majority of frames but potentially not to all of them.

ablwr:
Great, thanks @mbunkus! Will do! :D I will close and reference this PR for the convo/decisions made here.

mbunkus:
I just realized I forgot to substantiate my preference: it's simply easier to reason about. If I have one PR that only includes moving around stuff, then I can focus on the new order of headings, and I don't have to check each and every sentence with a keen eye (simply because I trust you not to mess up moving around stuff).

A subsequent PR only containing grammatical and semantic changes on the other hand allows me to focus on each and every change (as every little detail counts) as those changes are much fewer in number and smaller in scope.

Overall it makes reviewing much easier and less error-prone.

ablwr:
Oh I agree with you for the same reasons! PRs should follow good commit-message guidelines and coding guidelines, only fixing one thing.

notes.md
ablwr: Resolved in https://github.com/Matroska-Org/matroska-specification/pull/177

notes.md
ablwr: Resolved (and conversation/question moved to) https://github.com/Matroska-Org/matroska-specification/pull/178

PR #175, opened 2017-09-01
Make omitting audio CuePoint Elements optional

t-rapp:
When video bitrate is large (FFV1 1080p) and seeking latency is high (tape-based storage system) the additional round-trips to find the accurate audio track position can be in the range of minutes. So for this scenario adding audio tracks to the cluster index makes sense and index size overhead is still small compared to the total file size.

Thus change the recommendation text from "SHOULD" into "MAY".


mbunkus:
Looks like agreement across the board. Merging.

PR #178, opened 2017-09-10
capitalize MIME, rephrase usage

ablwr:
Capitalizes MIME and rephrases the usage recommendation. I didn't change it to "These are RECOMMENDED." In general, I am not sure the status of this section, if there will be an IETF-endorsed MIME type, if that falls into the scope of this standardization effort... etc. Advice?


retokromer:
>if there will be an IETF-endorsed MIME type

WebM can be used from HTML5, therefore no additional MIME type is needed. Yet we should indeed advocate for Matroska in HTML5!

PR #179, opened 2017-09-10
removes octet definition section

ablwr:
It is superfluous to have a section dedicated to this definition and not within standardization scope IMO.

### DIFF
--- a/notes.md
+++ b/notes.md
@@ -77,10 +77,6 @@ There is no IETF endorsed MIME type for Matroska files. But you can use the ones
 * .mkv : Matroska video `video/x-matroska`
 * .mk3d : Matroska 3D video `video/x-matroska-3d`
 
-# Octet
-
-An Octet refers to a byte made of 8 bits.
-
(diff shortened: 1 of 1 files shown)


mbunkus:
What's the reason for this removal?

ablwr:
I think this is a commonly-known computer term and not specific to Matroska

mbunkus:
You're right :grin:

retokromer:
BTW: octet is byte in French.

robUx4:
Actually a byte is not an octet. It can have various sizes, whereas an octet is always 8 bits. That's why we use the name `octet` in the specifications.
https://en.wikipedia.org/wiki/Octet_(computing)

ablwr:
Ceci n'est pas une BYTE!

retokromer:
Yes, historically, it was the size of the accumulator (with the students I prefer to use «word»). I had in mind that ISO/IEC 80000 did standardise that which SI, but I am probably wrong, sorry.

PR #183, opened 2017-09-11
rephrase versioning section

ablwr:


notes.md
```diff
@@ -53,13 +53,13 @@ Thanks to the PixelCropXXX elements, it's possible to crop the image before bein
 
 PixelXXX (size of the coded image) -> PixelCropXXX (size of the image to keep) -> DisplayXXX (resized cropped image)
 
-# Matroska version indicators
+# Matroska versioning
 
-The EBML Header each Matroska file starts with contains two version number fields that inform a reading application about what to expect. These are `DocTypeVersion` and `DocTypeReadVersion`.
+The `EBML Header` of each Matroska file MUST contain the two version number fields that inform a reading application on what to expect. These are `DocTypeVersion` and `DocTypeReadVersion`.
```
dericed: This update maintains the original meaning which is fine, but it isn't true that the header MUST contain the version fields, as if not then the default values would be used. Would prefer to say something like "The EBML Header of each Matroska document informs the reading application on what version of Matroska to expect...."

notes.md
```diff
@@ -53,13 +53,13 @@ Thanks to the PixelCropXXX elements, it's possible to crop the image before bein
...
 
-`DocTypeVersion` MUST contain the highest Matroska version number of any Element present in the Matroska file. For example, a file using the SimpleBlock Element MUST have a `DocTypeVersion` of at least 2 while a file containing `CueRelativePosition` Elements MUST have a `DocTypeVersion` of at least 4.
+`DocTypeVersion` MUST contain the highest Matroska version number of any `Element` present in the Matroska file. For example, a file using the `SimpleBlock Element` MUST have a `DocTypeVersion` of at least 2. A file containing `CueRelativePosition` Elements MUST have a `DocTypeVersion` of at least 4.
```
dericed: nitpick on original language rather than @ablwr's update, but there's a mismatch between:
> `DocTypeVersion` MUST contain the highest Matroska version number of any `Element` present

and 
> A file containing `CueRelativePosition` Elements MUST have a `DocTypeVersion` of at least 4.

Is it `must equal` or `must be equal or greater than`? I'd vote for equal or greater than.

mbunkus: Nothing in the two partial sentences quoted says that DTV must equal anything in particular. In the examples there might be other elements present with a higher version number requirement. The examples are just the lower bounds for the version number (e.g. if a file contains a `SimpleBlock` element, the version number cannot be lower than 2, but it can be higher, e.g. if the file contains a `CueRelativePosition`, too).

Granted, one could have chosen better examples that explain how DTV must actually equal a certain value. But that's not easy as you'd have to say that a certain element is present but all other elements with higher version number requirements are absent.

ablwr: Hmm... Would it be suitable to clarify as the following...

`DocTypeVersion` MUST contain the highest Matroska version number of any `Element` present in the Matroska file. For example, a file using the `SimpleBlock Element` MUST have a `DocTypeVersion` equal or greater than 2. A file containing `CueRelativePosition` Elements MUST have a `DocTypeVersion` equal or greater than 4.

I also don't see the mismatch.

ablwr: Overall I think these paragraphs could be better worded in relation to each other...

notes.md
```diff
@@ -53,13 +53,13 @@ Thanks to the PixelCropXXX elements, it's possible to crop the image before bein
...
 
-The `DocTypeReadVersion` MUST contain the minimum version number a reading application MUST at least support properly in order to play the file back (optionally with a reduced feature set). For example, if a file contains only Elements of version 2 or lower except for `CueRelativePosition` (which is a version 4 Matroska Element) then `DocTypeReadVersion` SHOULD still be set to 2 and not 4 because evaluating `CueRelativePosition` is not REQUIRED for standard playback -- it only makes seeking more precise if used.
+The `DocTypeReadVersion` MUST contain the minimum version number that a reading application can minimally support in order to play the file back (optionally with a reduced feature set). For example, if a file contains only `Elements` of version 2 or lower except for `CueRelativePosition` (which is a version 4 Matroska `Element`), then `DocTypeReadVersion` SHOULD still be set to 2 and not 4 because evaluating `CueRelativePosition` is not necessary for standard playback -- it only makes seeking more precise if used.
```
retokromer: >-- it only makes seeking more precise if used.

The `only` here sounds strange to me.

And I don’t understand why parenthesis are used for
> optionally with a reduced feature set

but `--` here. I suggest to unify in a way ot the other.

notes.md
dericed: Regarding the mismatch, for instance the output of `ffmpeg -f lavfi -i testsrc -t 1 test.mkv`. This file is DocTypeVersion = 4, but contains no Element defined as minver=4, so this file would be invalid to this mandate:
> `DocTypeVersion` MUST contain the highest Matroska version number of any `Element` present in the Matroska file.

The DocTypeVersion here contains `4` which is not the highest Matroska version number of any Element present in the Matroska file.

mbunkus: Ah, now I see your point. Thanks.

Proposal:

> `DocTypeVersion` MUST be equal to or greater than the highest Matroska version number of any Element present in the Matroska file.

ablwr:
Updated based on @dericed @mbunkus and @retokromer comments and squashed/force-pushed as @mbunkus prefers :)

ablwr:
I think it'd be fine either way, but since we use that phrasing earlier, I updated them as you suggested.

PR #184, opened 2017-09-11
rephrase image cropping section

ablwr:


notes.md
```diff
@@ -49,9 +49,9 @@ Encryption information is stored in the `ContentEncodings Element` under the `Co
 
 # Image cropping
 
-Thanks to the PixelCropXXX elements, it's possible to crop the image before being resized. That means the image size follows this path:
+Thanks to the `PixelCrop Elements` (`PixelCropTop`, `PixelCropBottom`, `PixelCropRight` and `PixelCropLeft`), it's possible to crop the image before being resized. The image size follows this pattern, using `PixelCropTop` as an example:
```
mbunkus: > …it's possible to crop the image before being resized.

This sounds strange to me (not only in your modification but in the original already). I'd prefer either "it's possible to crop the image before resizing it" or "it's possible to crop the image before it's resized".

notes.md
```diff
@@ -49,9 +49,9 @@ Encryption information is stored in the `ContentEncodings Element` under the `Co
...
 
-PixelXXX (size of the coded image) -> PixelCropXXX (size of the image to keep) -> DisplayXXX (resized cropped image)
+`PixelHeight` (size of the coded image) -> `PixelCropTop` (size of the image to keep) -> `DisplayHeight` (resized cropped image)
```
mbunkus: We should be precise here and state that the value of `PixelCropTop` must be subtracted from `PixelHeight` in order to calculate the height of the cropped image. Or maybe make it even more explicit including all four values:

> `PixelWidth` (width of the coded image) → remove `PixelCropLeft` pixels from the left and `PixelCropRight` pixels from the right (size of the cropped image to keep) → resize to `DisplayWidth` (width of cropped & resized image)
> `PixelHeight` (size of the coded image) → remove `PixelCropTop` pixels from the top and `PixelCropBottom` pixels from the bottom (size of the cropped image to keep) → resize to `DisplayHeight` (height of cropped & resized image)

Doesn't sound that nice either, but I'd like to have state what a player must do a bit more explicitly than your change proposes.

ablwr:
@mbunkus I included your comments and rephrased the section quite a bit, let me know what you think!

notes.md
```diff
@@ -49,9 +49,12 @@ Encryption information is stored in the `ContentEncodings Element` under the `Co
 
 # Image cropping
 
-Thanks to the PixelCropXXX elements, it's possible to crop the image before being resized. That means the image size follows this path:
+The `PixelCrop Elements` (`PixelCropTop`, `PixelCropBottom`, `PixelCropRight` and `PixelCropLeft`) allow an image to be cropped before it is resized. For `PixelHeight`, the value of `PixelCropTop` and/or `PixelCropBottom` MUST be subtracted from `PixelHeight` in order to calculate the height of the cropped image. For `PixelWidth`, the value of `PixelCropLeft` and/or `PixelCropRight` MUST be subtracted from `PixelWidth` in order to calculate the height of the cropped image.
```
dericed: > For `PixelHeight`, the value of `PixelCropTop` and/or `PixelCropBottom` MUST be subtracted from `PixelHeight` in order to calculate the height of the cropped image.

The conjunction `and/or` is confusing. When would it be `and` and when `or`?

mbunkus: I guess `and` if both are present and `or` if only one of them is present. Changing it to `and` would be fine by me, though.

ablwr: Should it be phrased in a way that doesn't use RFC2119 statements?

dericed: sorry for nitpicks, I reviewed the corresponding documentation from the quicktime clap atom, and then rewrote this section. A proposal:

The \`PixelCrop Elements\` (\`PixelCropTop\`, \`PixelCropBottom\`, \`PixelCropRight\` and \`PixelCropLeft\`) indicate when and by how much encoded videos frames SHOULD be cropped for display. These Elements allow edges of the frame that are not intended for display, such as the sprockets of a full-frame film scan or the VANC area of a digitized analog videotape, to be stored but hidden. \`PixelCropTop\` and \`PixelCropBottom\` store an integer of how many rows of pixels SHOULD be cropped from the top and bottom of the image (respectively). \`PixelCropLeft\` and \`PixelCropRight\` store an integer of how many columns of pixels SHOULD be cropped from the left and right of the image (respectively). For example, a pillar-boxed video that stores a 1440x1080 visual image within the center of a padded 1920x1080 encoded image MAY set both \`PixelCropLeft\` and \`PixelCropRight\` to \`240\`, so that a Matroska Player SHOULD crop off 240 columns of pixels from the left and right of the encoded image to present the image with the pillar-boxes hidden.

mbunkus: :+1: I like this a lot.

retokromer: Makes sense and LGTM.

ablwr: I updated with Dave's proposal and credited him accordingly. I got rid of the examples too since this does a good job at explaining and they were a bit confusing anyway. Let me know if you didn't want that to happen.

PR #188, opened 2017-09-11
moves DefaultDecodedFieldDuration under Default Values

ablwr:
Note that the requested change from PR https://github.com/Matroska-Org/matroska-specification/pull/173/files has been adjusted here.


ablwr:
How about if DefaultDecodedFieldDuration becomes its own top-level element underneath Default Values? I made the change.

PR #191, opened 2017-09-16
moves timecodes to one section

ablwr:
only moves, no grammar/syntax updates! Still doesn't look too clean..


ablwr:
That edit will immediately follow this one! I think the move from # to ## confused it

PR #192, opened 2017-09-17
Clarify pcm signedness

dericed:
addresses
https://github.com/Matroska-Org/matroska-specification/issues/142

### DIFF
--- a/codec_specs.md
+++ b/codec_specs.md
@@ -301,7 +301,7 @@ Codec ID: A_PCM/INT/BIG
 
 Codec Name: PCM Integer Big Endian
 
-Description: The audio bit depth MUST be read and set from the `BitDepth Element`. Corresponding ACM wFormatTag : ???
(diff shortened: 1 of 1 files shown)


dericed:
WIP, see questions in https://github.com/Matroska-Org/matroska-specification/issues/142#issuecomment-330004950

codec_specs.md
```diff
@@ -321,7 +321,7 @@ Codec ID: A_PCM/FLOAT/IEEE
 
 Codec Name: Floating Point, IEEE compatible
 
-Description: The bitdepth has to be read and set from KaxAudioBitDepth element (32 bit in most cases). The float are stored in little endian order (most common float format). Corresponding ACM wFormatTag : 0x0003
+Description: The audio bit depth MUST be read and set from the `BitDepth Element` (32 bit in most cases). The floats are stored in little endian order (most common float format). Corresponding ACM wFormatTag : 0x0003
```
mbunkus: Hmm, one usually only talks about endianess in the context of integer data. In the context of floating point numbers one only states that the format is (e.g.) IEEE floating-point numbers with 32 or 64 bits precision. Those only come with one type of layout.

dericed: I suggest that this is not blocking the PR. I can add an issue that the source spec of IEEE floats could be more clear.

Issue #194, opened 2017-09-24
define Matroska Player and Matroska Reader


dericed:
Not sure if we have an IETF style notations section yet, but both `Matroska Player` and `Matroska Reader` should be defined. Related comment: https://github.com/Matroska-Org/matroska-specification/pull/190#discussion_r139317047

ablwr:
Thanks, I'll take a look at this based on the convo, etc., and propose changes / sync across documentation

ablwr:
This seems to have been resolved. Re-open if I am mistaken.

PR #198, opened 2017-09-25
defines matroska, reader, and player; update typo for Notation

ablwr:
Please comment on these definitions (and preferably define them better). #194

### DIFF
--- a/index_matroska.md
+++ b/index_matroska.md
@@ -84,9 +84,17 @@ Attacks on a `Matroska Reader` could include:
 
 To be determined.
 
-# Notations and Conventions
+# Notation and Conventions
 
 The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
+
+This document defines specific terms in order to define the format and application of `Matroska`.  Specific terms are defined below:
+
+`Matroska`: a multimedia container format based on EBML (Extensible Binary Meta Language)
+
+`Matroska Player`: A `Matroska Player` is a data parser that interprets the semantics of a Matroska document and creates a way for programs to play `Matroska` documents as audiovisual files.
+
+`Matroska Reader`: A `Matroska Reader` is a data parser that interprets the semantics of a Matroska document and creates a way for programs to use `Matroska`.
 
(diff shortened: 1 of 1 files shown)


robUx4:
LGTM

hubblec4:
Well done @ablwr !

(What means LGTM?)

robUx4:
Looks Good To Me

retokromer:
>For similar reasoning I think we could say that a `Matroska Player` is a `Matroska Reader` that creates a way…

I second, I think this makes things clearer.

index_matroska.md
```diff
@@ -84,10 +84,18 @@ Attacks on a `Matroska Reader` could include:
...
 
 The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
 
+This document defines specific terms in order to define the format and application of `Matroska`.  Specific terms are defined below:
+
+`Matroska`: a multimedia container format based on EBML (Extensible Binary Meta Language)
+
+`Matroska Player`: A `Matroska Player` is a data parser that interprets the semantics of a Matroska document and creates a way for programs to play `Matroska` documents as audiovisual files.
+
+`Matroska Reader`: A `Matroska Reader` is a data parser that interprets the semantics of a Matroska document and creates a way for programs to use `Matroska`.
```
JeromeMartinez: > to use `Matroska`

IMO it is too generic, I don't understand the idea behind having 2 names, what is the purpose of the `Matroska Reader`.

retokromer: See the discussion at https://github.com/Matroska-Org/matroska-specification/pull/190#discussion_r139317047

JeromeMartinez: sure, but for the moment it is not in the PR, and IMO it should be expanded with some content from the comment you point out.
(but I don't really know wich wording to use :( )

robUx4: Not sure we should repeat the name in the definition. It would be lighter to read like this:
`Matroska Player`: A data parser that interprets...

retokromer: Or simply:
A `Matroska Player` is a data parser…

ablwr: Reason for this was mimicking how it's set up in the EBML spec https://datatracker.ietf.org/doc/draft-ietf-cellar-ebml/?include_text=1

PR #202, opened 2017-09-25
fix to Default Edition logic

dericed:
fixes the error that @hubblec4 found in https://github.com/Matroska-Org/matroska-specification/pull/190#issuecomment-331874813

### DIFF
--- a/chapters.md
+++ b/chapters.md
@@ -25,7 +25,7 @@ The `EditionFlagHidden Flag` behaves similarly to the `ChapterFlagHidden Flag`: 
 
 #### EditionFlagDefault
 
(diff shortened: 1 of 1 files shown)


hubblec4:
Now it is ok.

dericed:
@hubblec4 I am still curious of:

> When all `EditionFlagDefault Flags` are set to `false`, then the first `Edition` is the `Default Edition`.

So:
```
edition 1
 - default false
 - hidden true
edition 2
 - default false
 - hidden false
```

The Default Edition is `1`, right? Since all Editions have default set to false.

hubblec4:
Yes you are right. The description is not perfect.

..., then the first `Edition` with `EditionFlagHidden Flag` set to `false` is the `Default Edition`.

"... is the Default Edition" is also not really right, better is: 

... is the `Edition` which have to use. 

A Default Edition Flag is not set, so we can't call it Default Edition.

hubblec4:
Maybe it make sense to define the term  `Default Edition`:

The `Default Edition` is the `Edition` which MUST played by the `Matroska Player`.

hubblec4:
When all `EditionFlagDefault Flags` are set to `false`, then the first `Edition` with the `EditionFlagHidden Flag` is set to `false` MUST played by the `Matroska Player`.

PR #205, opened 2017-10-30
All EditionFlagDefault Flags set to false

hubblec4:


chapters.md
```diff
@@ -25,7 +25,7 @@ The `EditionFlagHidden Flag` behaves similarly to the `ChapterFlagHidden Flag`:
 
 #### EditionFlagDefault
 
-It is RECOMMENDED that no more than one `Edition` have an `EditionFlagDefault Flag` set to `true`. The first `Edition` with both the `EditionFlagDefault Flag` set to `true` and the `EditionFlagHidden Flag` set to `false` is the `Default Edition`. When all `EditionFlagDefault Flags` are set to `false`, then the first `Edition` is the `Default Edition`.
+It is RECOMMENDED that no more than one `Edition` have an `EditionFlagDefault Flag` set to `true`. The first `Edition` with both the `EditionFlagDefault Flag` set to `true` and the `EditionFlagHidden Flag` set to `false` is the `Default Edition`. When all `EditionFlagDefault Flags` are set to `false`, then the first `Edition` with the `EditionFlagHidden Flag` is set to `false`, MUST be played by the Matroska Player.
```
dericed: for grammar I think:
> the first `Edition` with the `EditionFlagHidden Flag` is set to `false`

should be:
> the first `Edition` with the `EditionFlagHidden Flag` set to `false`

After this proposed change, the term `Default Edition` is defined but not reused. If intended, then maybe the term `Default Edition` should be removed since after definition it is unused.

IMHO
> When all `EditionFlagDefault Flags` are set to `false`, then the first `Edition` with the `EditionFlagHidden Flag` is set to `false`

could be more simply expressed as:
When all `EditionFlagDefault Flags` are set to `false`, then the first `Edition`

Or am I missing a semantic difference?

hubblec4:
Your first change is OK.

But not the second one. Cause the first(1.) Edition could be hidden, so "a next" Edition (2 or 3 or 4), which is not hidden is used. 
And this Edition is not the Default Edition, it is a normal Edition and so we don't should use the term `Default Edition`.
You started here with the discussion
https://github.com/Matroska-Org/matroska-specification/pull/202

The term Default Edition is needed, because more than one Edition can have the Default Flag set to true.

hubblec4:
The title was wrong, arrgh (All EditionFlagHidden Flags set to false)

All EditionFlagDefault Flags set to false

hubblec4:
This PR is obsolete.
New PR is https://github.com/cellar-wg/matroska-specification/pull/336

PR #207, opened 2017-11-16
Remove bogus claim a 0 BlockDuration value means it's not a keyframe

robUx4:


JeromeMartinez:
[mkvvalidator](https://ffmpeg.org/pipermail/ffmpeg-devel/2017-November/219642.html) should be modified too, I guess.

PR #208, opened 2017-11-18
Add the Contributor Covenant code of conduct

robUx4:
The email is the same general contact email found on the website.


JeromeMartinez:
I prefer my version of the added text but I am not objective :).
Fine except that you use different word for same thing "added language vs "added text, I would use same terminology in both places.

PR #214, opened 2018-01-02
remove incorrect/confusing size calculation for Block/SimpleBlock ele…

### LINKED ISSUE #213: Block/SimpleBlock size equations are wrong
The [specs](https://github.com/Matroska-Org/matroska-specification/blob/master/index_matroska.md) say the following in the section "Block Structure" and "SimpleBlock structure":

> Size = 1 + (1-8) + 4 + (4 + (4)) octets. So from 6 to 21 octets.

This is, at beast, highly misleading and unclear, and at worst, just wrong.

Better numbers would be:

  * 1 byte: element ID
  * 1–8 bytes: element size (variable-sized integer)
  * 1–8 bytes: track number (variable-sized integer)
  * 2 bytes: block's timestamp relative to the cluster's timestamp (signed 16-bit integer)
  * 1 byte: flags
  * 1 byte: number of laced frames; only present if lacing is signaled in the flags
  * n bytes: lacing information; only present if lacing is signaled in the flags; total size depends on lacing method and number/size of laced frames; no upper limit (e.g. the Xiph method is rather wasteful for big frame sizes)
  * n bytes: content of laced frames

My guess is that the `1` in the equation refers to the element ID, the first `4` is supposed to be the combination of track number, relative timestamp & flags (however, that is wrong if the track number is >= 128), and I have no idea what the `(4 + (4))` is supposed to convey.

It's true that the minimum number of bytes is 6, but it's completely false that the maximum number of octets is 21. Due to how lacing works, especially the incredibly bad Xiph lacing, the number of bytes doesn't really have an upper bound.

The same applies to the `SimpleBlock` structure & equation.

mbunkus:
…ments

Fixes #213.

### DIFF
--- a/index_matroska.md
+++ b/index_matroska.md
@@ -137,8 +137,6 @@ Each level can have different meanings for audio and video. The ORIGINAL_MEDIUM 
 
 ### Block Structure
 
-Size = 1 + (1-8) + 4 + (4 + (4)) octets. So from 6 to 21 octets.
-
 Bit 0 is the most significant bit.
 
 Frames using references SHOULD be stored in "coding order". That means the references first and then the frames referencing them. A consequence is that timecodes MAY NOT be consecutive. But a frame with a past timecode MUST reference a frame already known, otherwise it's considered bad/void.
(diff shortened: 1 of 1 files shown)


mbunkus:
I've intentionally removed the calculation without a replacement as other parts of the specs don't talk about how big elements can be either. My bullet point list from #213 would just rehash what's written further below in the specs in the first place.

PR #215, opened 2018-01-03
bump to version 04

dericed:


dericed:
This bumps matroska to version 04 and sets codec and tags to version 00 (from a mistakenly set 01) for their first release in the cellar document tracker. Matroska version 3 expires on January 4th.

PR #217, opened 2018-01-09
removes ideas/cruft from menu section #201

### LINKED ISSUE #201: Cruft/ideas in Menu section
The Menu sections gets a little ill-formed at the bottom, see excerpt:

```

## Ideas

!!!! KNOW Where the main/audio/subs menu starts wherever we are (use chapters) !!!!

!!!! Keep in mind the state of the selected tracks of each kind (more than 1 for each possible) !!!!
!!!! Order of blending !!!!
!!!! What if a command is not supported by the player ? !!!!
!!!! Track selection issue, only applies when 'quitting' the menu (but still possible to change live too) !!!!
!!!! Allow to hide (not render) some parts of a movie for certain editions !!!!
!!!! Get the parental level of the player (can be changed live) !!!!


## Data Structure

As a Matroska side project, the obvious choice for storing binary data is EBML.
```

I suggest removal but seek approval.

ablwr:
resolves #201

### DIFF
--- a/menu.md
+++ b/menu.md
@@ -55,20 +55,3 @@ Some `Matroska Players` might not support the control track. That mean they will
 Matroska Source file -> Control Track <-> Player.
                      -> other tracks   -> rendered
 ```
-
-## Ideas
-
-```
-!!!! KNOW Where the main/audio/subs menu starts wherever we are (use chapters) !!!!
-
-!!!! Keep in mind the state of the selected tracks of each kind (more than 1 for each possible) !!!!
-!!!! Order of blending !!!!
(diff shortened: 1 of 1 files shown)


dericed:
OK with me, pinging @hubblec4 if there's any ideas in here worth incorporating elsewhere.

hubblec4:
Like Steve said we could move the Menu ideas  to a normal issue and so we have stored this info outside of the Matroska specs. 

It makes total sense to me, to separate the Matroska Menu from the other specs.

ablwr:
I just made a general issue, as auto-linked above.

PR #218, opened 2018-01-09
makes section names unique; removes dupe Laced Data section

### LINKED ISSUE #113: use unique section names
Currently a few section names are repeated in different context which breaks the TOC in the IETF html output, such as:

- Matroska Element Ordering Guidelines / Chapters
- Chapters

The section should be renamed to be unique regardless of context.

ablwr:
Seeks to resolve #113.

### DIFF
--- a/attachments.md
+++ b/attachments.md
@@ -3,8 +3,6 @@
 
 # Attachments
 
-## Introduction
-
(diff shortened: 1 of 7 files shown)


dericed:
thx this is much cleaner

mbunkus:
I like! Thanks.

PR #220, opened 2018-01-13
move matroska tag list into xml

dericed:


dericed:
This is a work in progress to move the Matroska tag list into an xml document. During the make process the contents of the xml can be formatting and then inserted into the Matroska Tags RFC documents. I think this structure will make it a bit easier to maintain the tags and likely easier for developers to integrate these vocabularies into systems. I also need to add the process to 'make' the tag list, but am sending this PR with the xml for any comments about the xml or the idea in general.

matroska_tags.xml
```diff
@@ -0,0 +1,306 @@
...
+    <description>Nested Information includes tags contained in other tags.</description>
+  </class>
+  <tag tagname="ORIGINAL" class="Nesting Information" tagtype="Container">
+    <description>A special tag that is meant to have other tags inside (using nested tags) to describe the original work of art that this item is based on. All tags in this list can be used "under" the ORIGINAL tag like LYRICIST, PERFORMER, etc.</description>
+  </tag>
+  <tag tagname="SAMPLE" class="Nesting Information" tagtype="Container">
+    <description>A tag that contains other tags to describe a sample used in the targeted item taken from another work of art. All tags in this list can be used "under" the SAMPLE tag like TITLE, ARTIST, DATE_RELEASED, etc.</description>
+  </tag>
+  <tag tagname="COUNTRY" class="Nesting Information" tagtype="UTF-8">
+    <description>The name of the country ([biblio ISO-639-2](http://lcweb.loc.gov/standards/iso639-2/englangn.html#two)) that is meant to have other tags inside (using nested tags) to country specific information about the item. All tags in this list can be used "under" the COUNTRY_SPECIFIC tag like LABEL, PUBLISH_RATING, etc.</description>
```
retokromer: `https`

matroska_tags.xml
```diff
@@ -0,0 +1,306 @@
...
+    <description>The name of the sound engineer or sound recordist.</description>
+  </tag>
+  <tag tagname="ART_DIRECTOR" class="Entities" tagtype="UTF-8">
+    <description>The person who oversees the artists and craftspeople who build the sets.</description>
+  </tag>
+  <tag tagname="PRODUCTION_DESIGNER" class="Entities" tagtype="UTF-8">
+    <description>Artist responsible for designing the overall visual appearance of a movie.</description>
+  </tag>
+  <tag tagname="CHOREGRAPHER" class="Entities" tagtype="UTF-8">
+    <description>The name of the choregrapher</description>
```
retokromer: `choregrapher` or `choreographer `?

matroska_tags.xml
dericed: The pr regards the make process rather than any semantic change. I suggest to leave this for subsequent changes.

dericed:
The PR is now updated with the make process. The rfc txt output is identical after the commits.

matroska_tags.xml
retokromer: OK, I just saw it while browsing through.

tagging_end.md
```diff
@@ -0,0 +1,3 @@
+## Notes
+
+* In the Target list, a logical OR is applied on all tracks, a logical OR is applied on all chapters. Then a logical AND is applied between the Tracks list and the Chapters list to know if an element belongs to this Target.
```
ablwr: Wouldn't it be better to use the word "and" instead of a comma in this first sentence?

dericed: This isn't new text but it the part that used to follow the metadata list in tagging.md. In the Make process now tagging.md the-xml-translation and tagging_end.md are concatenated. The pr overall does not intend any semantic change, so I suggest this is addressed in a subsequent pr.

retokromer:
Can this be merged? Then I can open another PR with the link fixes. Otherwise I will forgot that.

PR #223, opened 2018-03-24
RAWcooked reserved IDs

JeromeMartinez:
As there is no private IDs in Matroska AFAIK, we would like to publicly reserve some IDs for [RAWcooked](https://mediaarea.net/RAWcooked) project.
We need IDs at different levels:
- TrackEntry level: we need to store private data per track (0 to 1 occurrence) e.g. file header/footer content of the file corresponding to the whole track.
- BlockGroup level: we need to store private data per Block (0 to 1 occurrence) e.g. file header/footer content of the file corresponding to the single block.
- Segment level: we need to store private data across the file (0 to multiple occurrences) e.g. forward error correction data for several Matroska clusters.

Data inside such elements is intended to be opaque from Matroska parsers (binary type) but specifications will be public when fully ready.

I set minver to 4 because I am not sure we could add officially new elements to older versions, but older versions could also have it if it is possible to add elements to older versions (parsers are expected to just ignore these elements).

I don't use to work with this XML, so please double check.

### DIFF
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -184,6 +184,9 @@
       <a href="http://labs.divx.com/node/16601">DivX trick track extensions</a>
     </documentation>
   </element>
+  <element name="RawcookedBlockGroup" path="0*1(\Segment\Cluster\BlockGroup\RawcookedBlockGroup)" id="0x207262" type="binary" maxOccurs="1" minver="4">
+    <documentation lang="en">Data needed by RAWcooked features at the BlockGroup level, see <a href="https://mediaarea.net/RAWcooked">RAWcooked project page</a> for more information</documentation>
+  </element>
   <element name="EncryptedBlock" path="0*(\Segment\Cluster\EncryptedBlock)" id="0xAF" type="binary" minver="0" maxver="0" webm="0">
     <documentation lang="en">Similar to <a href="https://www.matroska.org/technical/specs/index.html#SimpleBlock">SimpleBlock</a> but the data inside the Block are Transformed (encrypt and/or signed). (see <a href="https://www.matroska.org/technical/specs/index.html#encryptedblock_structure">EncryptedBlock Structure</a>)</documentation>
   </element>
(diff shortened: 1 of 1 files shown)


robUx4:
See https://github.com/Matroska-Org/ebml-specification/pull/181

dericed:
Could https://github.com/Matroska-Org/matroska-specification/pull/276 supersede this?

robUx4:
I think #276 should be rewritten to use either use `BlockAdditionMapping` (which ties it to `BlockGroup`) but that means it's incompatible with files you may already have created.
Otherwise use an EBML extension that add element pathes (the #181) seems to be the way to go. But this extension is also missing from files you created (I assume).

JeromeMartinez:
>  but that means it's incompatible with files you may already have created.

That's fine, I didn't created yet such files :).
I was waiting for a well accepted method (for the moment I use an attachment).

> BlockAdditionMapping 

For the moment, with the discussion from CELLAR, it would fit the track part and the block part, but the issue remains for RAWcooked part linked to a segment (I am thinking to use a "Tags" elements for that, it is not limited to 1 instance in opposition to "Attachments", and my need is to have an element I can add even if another muxer already added such element at the beginning of the file).

robUx4:
> > but that means it's incompatible with files you may already have created.
> 
> That's fine, I didn't created yet such files :).
> I was waiting for a well accepted method (for the moment I use an attachment).

Great. The proposal is coming.
 
> > BlockAdditionMapping
> 
> For the moment, with the discussion from CELLAR, it would fit the track part and the block part, but the issue remains for RAWcooked part linked to a segment (I am thinking to use a "Tags" elements for that, it is not limited to 1 instance in opposition to "Attachments", and my need is to have an element I can add even if another muxer already added such element at the beginning of the file).

Yes `Tags` are probably fine for that. This way you need to depend on parser udpates or defining an extension (and also wait for parsers to support that).

JeromeMartinez:
At least the Track and Block part, in favor of BlockAdditional stuff.
For segment part, I may find something else (Tags?).
So I close for the moment.

PR #226, opened 2018-05-23
Remove restriction from VideoPixelCropBottom

mkver:
Currently VideoPixelCropBottom has the restriction that it only applies to HDTV content; the other VideoPixelCrop elements don't have said restriction. Besides this restriction being nonsensical in itself it is also unclear because it isn't defined what HDTV means exactly (or why it should be restricted to HDTV and not to HD more general). This commit brings VideoPixelCropBottom in line with the specification of the other elements.


mbunkus:
I interpreted this as more like an example of when the element might be used. But I still agree that it is both superfluous and ambiguous and therefore harmful.

robUx4:
This is from the time MPEG-2 would use 1088 lines for 1080 content. But it may apply to a lot of other things.

PR #235, opened 2018-06-26
rename EditionFlagHidden to EditionFlagDisabled and rework its definition

robUx4:
eventually Control Tracks will go away or be defined in a future version

### DIFF
--- a/chapters.md
+++ b/chapters.md
@@ -15,12 +15,12 @@ As an example, consider a `Parent ChapterAtom Element` that has its `ChapterFlag
 
 ### Edition Flags
 
-Three `Edition Flags` are defined to describe the behavior of the `EditionEntry Element`: `EditionFlagHidden`, `EditionFlagDefault` and `EditionFlagOrdered`.
+Three `Edition Flags` are defined to describe the behavior of the `EditionEntry Element`: `EditionFlagDisabled`, `EditionFlagDefault` and `EditionFlagOrdered`.
 
 
-#### EditionFlagHidden
+#### EditionFlagDisabled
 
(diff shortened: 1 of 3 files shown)


robUx4:
> With your change it now reads "it indicates all the data inside this edition MUST be discarded and not presented to the user".

That's a wording issue. It means all EBML elements inside this Edition MUST be discarded. Not the rest of the EBML Segment.

robUx4:
I don't see how people would be using it when the specs say a hidden edition can be unhidden by a Control Track. Something that doesn't exist anywhere.

robUx4:
In VLC if an edition is hidden the chapters are not shown to the user. And there's no way to flip that flag. It's exactly like a disabled flag.

mbunkus:
> > With your change it now reads "it indicates all the data inside this edition MUST be discarded and not presented to the user".

> That's a wording issue. It means all EBML elements inside this Edition MUST be discarded. Not the rest of the EBML Segment.

Well, the wording is really bad in that case. Can you clarify the wording, please? Say something like "all chapters atoms in the edition…". "All the data" can easily be interpreted to mean "all the track content covered by the chapter atoms inside this edition", which I obviously did.

Thanks.

> I don't see how people would be using it when the specs say a hidden edition can be unhidden by a Control Track. Something that doesn't exist anywhere.

MKVToolNix's programs have read and written the flag since ages. Hubblec4's Chapter Editor reads and writes that flag.

hubblec4:
Yo, I thought me that could be problematic with the existing tools that handle this Flag.
In the Matroska specs it is only a "renaming" and the description has to be a bit adjusted.

But MKVToolNix's chapter.xml is a well used format and to change there the Flag name is a bit work, for Mosu and also for me.

But I'm fine with that, more important is that all is correct.

PR #236, opened 2018-06-27
Allow defining the default values with a proper path

robUx4:
Tools using this XML will need to be updated accordingly
The DisplayWidth/DisplayHeight should also use this

### DIFF
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -130,7 +130,7 @@
   <element name="BlockAdditional" path="1*1(\Segment\Cluster\BlockGroup\BlockAdditions\BlockMore\BlockAdditional)" id="0xA5" type="binary" minOccurs="1" maxOccurs="1" minver="1" webm="0">
     <documentation lang="en">Interpreted by the codec as it wishes (using the BlockAddID).</documentation>
   </element>
-  <element name="BlockDuration" path="0*1(\Segment\Cluster\BlockGroup\BlockDuration)" id="0x9B" type="uinteger" maxOccurs="1" minver="1" default="DefaultDuration">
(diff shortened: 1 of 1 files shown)


ebml_matroska.xml
```diff
@@ -130,7 +130,7 @@
   <element name="BlockAdditional" path="1*1(\Segment\Cluster\BlockGroup\BlockAdditions\BlockMore\BlockAdditional)" id="0xA5" type="binary" minOccurs="1" maxOccurs="1" minver="1" webm="0">
     <documentation lang="en">Interpreted by the codec as it wishes (using the BlockAddID).</documentation>
   </element>
-  <element name="BlockDuration" path="0*1(\Segment\Cluster\BlockGroup\BlockDuration)" id="0x9B" type="uinteger" maxOccurs="1" minver="1" default="DefaultDuration">
+  <element name="BlockDuration" path="0*1(\Segment\Cluster\BlockGroup\BlockDuration)" id="0x9B" type="uinteger" maxOccurs="1" minver="1" default="\Segment\Tracks\TrackEntry\DefaultDuration">
```
dericed: This is an improvement, but still it's unclear *which* TrackEntry is referred to.
With an xpath style reference would be like:
`\Segment\Tracks\TrackEntry[TrackNumber=./Block/TrackNumber]\DefaultDuration`
to link the intended TrackEntry to the TrackNumber in the Block Header; unfortunately this style of xpath linking isn't feasible since the Track Number of the Block Header is not an element but a byte range of binary data.

dericed:
@robUx4, close without merge? I think this is superseded by https://github.com/cellar-wg/matroska-specification/pull/272

PR #237, opened 2018-06-28
Use ITU H.273 values for the transfer characteristics and colour primaries

robUx4:


ebml_matroska.xml
```diff
@@ -493,26 +493,28 @@
       <enum value="13" label="IEC 61966-2-1"/>
       <enum value="14" label="ITU-R BT.2020 10 bit"/>
       <enum value="15" label="ITU-R BT.2020 12 bit"/>
-      <enum value="16" label="SMPTE ST 2084"/>
+      <enum value="16" label="ITU-R BT.2100 Perceptual Quantization"/>
```
dericed: Does this break reverse compatibility?

robUx4: So it's the same thing, but the ITU one is accessible for free.

dericed:
LGTM, I suggest merging.

PR #239, opened 2018-06-30
AV1: Resolve conflicting recommendations & typos

mkver:
This commit resolves conflicting recommendations regarding the presence of `Sequence Header OBU` in keyframes in case the `Sequence Header OBU` doesn't change.
It also fixes some typos.


robUx4:
Thanks for this but please split the changes in different commits. Typos are one thing, removing a whole part of a sentence that changes the meaning of the spec shouldn't be in here.

PR #241, opened 2018-07-17
bump to wg doc version 00

dericed:


dericed:
to whoever merges this, please mark a release just after https://github.com/Matroska-Org/matroska-specification/releases/new

dericed:
merging since these docs are already at https://datatracker.ietf.org/wg/cellar/documents/ and otherwise the repo is out of sync

PR #244, opened 2018-07-18
New chapters.md

hubblec4:
This is a first big change for the chapters.md. It is not complete because I want to know is it the right way?


chapters.md
```diff
@@ -3,31 +3,82 @@
 
 # Chapters
 
-## Edition and Chapter Flags
+The Matroska Chapters system can consist of`Simple Chapters` where a chapter start time is used as marker in the timeline only. The system can be more complex with `Ordered Chapters` where a chapter end time stamp is additionally used or much more complex with `Linked Chapters`. The Matroska Chapters system can also have a menu structure, borrowed from the DVD menu system, or have it's own Native Matroska menu structure.
```
dericed: - missing space in "of\`Simple Chapters\`"
- is `Simple Chapters` intended as a formal term? Does it mean non `Ordered Chapters`?
- the qualities of the chapters should be reference within an edition rather than related to the `Matroska Chapters system` since a Matroska may include all of these types of chapters simultaneously.

chapters.md
```diff
@@ -3,31 +3,82 @@
...
 
-### Chapter Flags
+## EditionEntry
 
-Two `Chapter Flags` are defined to describe the behavior of the `ChapterAtom Element`: `ChapterFlagHidden` and `ChapterFlagEnabled`.
+The `EditionEntry` is also called an `Edition`.  
+An `Edition` is a set of `ChapterAtoms` and there MUST be at least one `ChapterAtom`.  
+Multiple `Editions` are allowed.
 
-If a `ChapterAtom Element` is the `Child Element` of another `ChapterAtom Element` with a `Chapter Flag` set to `true`, then the `Child ChapterAtom Element` MUST be interpreted as having its same `Chapter Flag` set to `true`. If a `ChapterAtom Element` is the `Child Element` of another `ChapterAtom Element` with a `Chapter Flag` set to `false` or if the `ChapterAtom Element` does not have a `ChapterAtom Element` as its `Parent Element`, then it MUST be interpreted according to its own `Chapter Flag`.
```
dericed: Does this section move somewhere else? It seems like this part is simply cut but I'm not sure why.

chapters.md
```diff
@@ -3,31 +3,82 @@
...
 
-## Edition and Chapter Flags
+The Matroska Chapters system can consist of`Simple Chapters` where a chapter start time is used as marker in the timeline only. The system can be more complex with `Ordered Chapters` where a chapter end time stamp is additionally used or much more complex with `Linked Chapters`. The Matroska Chapters system can also have a menu structure, borrowed from the DVD menu system, or have it's own Native Matroska menu structure.
 
-### Chapter Flags
+## EditionEntry
 
-Two `Chapter Flags` are defined to describe the behavior of the `ChapterAtom Element`: `ChapterFlagHidden` and `ChapterFlagEnabled`.
+The `EditionEntry` is also called an `Edition`.  
+An `Edition` is a set of `ChapterAtoms` and there MUST be at least one `ChapterAtom`.
```
dericed: An `Edition` (aka `EditionEntry`) is more than a set of `Chapter Atoms` since it also include Edition Flags. I suggest "An \`Edition\` contains a set of \`Edition Flags\` and MUST contain at least one \`ChapterAtom Element\`".

chapters.md
```diff
@@ -3,31 +3,82 @@
...
-Three `Edition Flags` are defined to describe the behavior of the `EditionEntry Element`: `EditionFlagHidden`, `EditionFlagDefault` and `EditionFlagOrdered`.
+### EditionFlagHidden
 
+When the `EditionFlagHidden Flag` is set to `false` means the `Edition` is visible and selectablly in a `Matroska Player`.
+All `ChapterAtoms Elements` MUST be interpreted of their own `ChapterFlagHidden Flags`.
 
-#### EditionFlagHidden
+ChapterAtom / ChapterFlagHidden | False | True | visible
+--------------------------------|-------|-------|--
+Chapter 1                       |   X   |  		|  yes
```
dericed: some alignment issues here and in all the tables

chapters.md
```diff
@@ -3,31 +3,82 @@
...
-The `EditionFlagHidden Flag` behaves similarly to the `ChapterFlagHidden Flag`: if `EditionFlagHidden` is set to `true`, its `Child ChapterAtoms Elements` MUST also be interpreted as if their `ChapterFlagHidden` is also set to `true`, regardless of their own `ChapterFlagHidden Flags`. If `EditionFlagHidden` is toggled by a `Control Track` to `false`, the `ChapterFlagHidden Flags` of the `Child ChapterAtoms Elements` SHALL determine whether the `ChapterAtom` is hidden or not.
+When the `EditionFlagHidden Flag` is set to `true` the `Edition` is hidden and SHOULD not be selectablly in a `Matroska Player`.
+It exists one case where a hidden `Edition` MUST be played:  
+All `Editions` `EditionFlagHidden Flags` are set to true, so there is no visible `Edition`.  
+In this case all `ChapterAtoms Elements` MUST also be interpreted as if their `ChapterFlagHidden Flag` is also set to `true`, regardless of their own `ChapterFlagHidden Flags`.
 
+ChapterAtom / ChapterFlagHidden | False | True | visible
+--------------------------------|-------|-------|--
+Chapter 1                       |   X   |  		|  no
+Chapter 2                       |       |  X	|  no
```
dericed: According to this table, both Chapter 1 and 2 are not visible despite having different settings for the hidden flag.

chapters.md
```diff
@@ -3,31 +3,82 @@
...
+All `ChapterAtoms Elements` MUST be interpreted of their own `ChapterFlagHidden Flags`.
 
-#### EditionFlagHidden
+ChapterAtom / ChapterFlagHidden | False | True | visible
+--------------------------------|-------|-------|--
+Chapter 1                       |   X   |  		|  yes
+Chapter 2                       |       |  X	|  no
 
-The `EditionFlagHidden Flag` behaves similarly to the `ChapterFlagHidden Flag`: if `EditionFlagHidden` is set to `true`, its `Child ChapterAtoms Elements` MUST also be interpreted as if their `ChapterFlagHidden` is also set to `true`, regardless of their own `ChapterFlagHidden Flags`. If `EditionFlagHidden` is toggled by a `Control Track` to `false`, the `ChapterFlagHidden Flags` of the `Child ChapterAtoms Elements` SHALL determine whether the `ChapterAtom` is hidden or not.
+When the `EditionFlagHidden Flag` is set to `true` the `Edition` is hidden and SHOULD not be selectablly in a `Matroska Player`.
```
dericed: "selectable"?

chapters.md
```diff
@@ -3,31 +3,82 @@
...
+When the `EditionFlagHidden Flag` is set to `false` means the `Edition` is visible and selectablly in a `Matroska Player`.
+All `ChapterAtoms Elements` MUST be interpreted of their own `ChapterFlagHidden Flags`.
 
-#### EditionFlagHidden
+ChapterAtom / ChapterFlagHidden | False | True | visible
+--------------------------------|-------|-------|--
+Chapter 1                       |   X   |  		|  yes
+Chapter 2                       |       |  X	|  no
 
-The `EditionFlagHidden Flag` behaves similarly to the `ChapterFlagHidden Flag`: if `EditionFlagHidden` is set to `true`, its `Child ChapterAtoms Elements` MUST also be interpreted as if their `ChapterFlagHidden` is also set to `true`, regardless of their own `ChapterFlagHidden Flags`. If `EditionFlagHidden` is toggled by a `Control Track` to `false`, the `ChapterFlagHidden Flags` of the `Child ChapterAtoms Elements` SHALL determine whether the `ChapterAtom` is hidden or not.
```
dericed: also curious where this section went or if it is superseded by other text here

chapters.md
```diff
@@ -3,31 +3,82 @@
...
+When the `EditionFlagHidden Flag` is set to `true` the `Edition` is hidden and SHOULD not be selectablly in a `Matroska Player`.
+It exists one case where a hidden `Edition` MUST be played:
```
dericed: I'm confused why this edge case is here. If all Editions are hidden then shouldn't the file operate as if there were no Editions at all?

chapters.md
```diff
@@ -3,31 +3,82 @@
...
+Chapter 2                       |       |  X	|  no
 
-#### EditionFlagDefault
+### EditionFlagDefault
 
-It is RECOMMENDED that no more than one `Edition` have an `EditionFlagDefault Flag` set to `true`. The first `Edition` with both the `EditionFlagDefault Flag` set to `true` and the `EditionFlagHidden Flag` set to `false` is the `Default Edition`. When all `EditionFlagDefault Flags` are set to `false`, then the first `Edition` is the `Default Edition`.
+It is RECOMMENDED that no more than one `Edition` have an `EditionFlagDefault Flag` set to `true`. The first `Edition` with both the `EditionFlagDefault Flag` set to `true` and the `EditionFlagHidden Flag` set to `false` is the `Default Edition`.
 
-#### EditionFlagOrdered
+Edition | FlagHidden | FlagDefault | used Edition
```
dericed: hmmm, maybe need a better term than `used Edition`

chapters.md
hubblec4: This section is deleted because it is wrong. FlagHidden works different as FlagEnabled.

chapters.md
hubblec4: This section is also deleted because it is wrong.

The same was above with the Chapter Flags.  
Control Track is a therm which I will use not here, I have planed a separate section.

chapters.md
hubblec4: Is the word "selectable" better as "selectablly" or is the meaning not clear?

chapters.md
hubblec4: Yes that is correct
> In this case all ChapterAtoms Elements MUST also be interpreted as if their ChapterFlagHidden Flag is also set to true, regardless of their own ChapterFlagHidden Flags.

When the edition FlagHidden = true then should be all chapters also hidden.

chapters.md
hubblec4: Done.
https://github.com/Matroska-Org/matroska-specification/pull/205/commits/f897de7c77cac103756e2fd3dccfc88ee8e7a1f7

PR #245, opened 2018-07-18
Hard-Linking change

hubblec4:
The current Hard-Linking describtion is not correct. I have added some examples.


dericed:
Is this PR related to https://github.com/Matroska-Org/matroska-specification/pull/244?

hubblec4:
No.

PR #246, opened 2018-07-20
ebml_matroska.xml - fixes typo in FlagInterlaced section

kieranjol:
BTW this is the first time I realised that ye are converting the XML here to all the various formats in the IETF spec - that's so cool!


robUx4:
Oops and thanks !

PR #249, opened 2018-07-21
Add avs2 codec mapping

hwrenx:


robUx4:
Thanks for your contribution. Are there anything capable of reading such files ?

Also in the future we should have a more detailed approach to how video formats are described. A very detailed example can be found [here](https://github.com/Matroska-Org/matroska-specification/blob/av1-mappin/codec/av1.md). It should at least explain how the PixelWidth/Height is extracted, what is inside the Block on the raw/original video format and when to set a KeyFrame.

hwrenx:
Introduction of AVS2 could be found in [What is AVS2](http://www.avs.org.cn/AVS2/en/index.asp) and [AVS2 Coding Tools](http://www.avs.org.cn/AVS2/en/tools.asp). Draft AVS2-P2 under IEEE 1857.4 has not been published yet, AVS2-P2 decoder  [davs2](https://github.com/pkuvcl/davs2) and encoder [xavs2](https://github.com/pkuvcl/xavs2) open source may be helpful.

dericed:
Seems fine to me. I suggest a small sample be submitted to https://github.com/Matroska-Org/matroska-test-files.

hwrenx:
Will be submitted if necessary.

JeromeMartinez:
@hwrenx I am interested in a sample file for adding support of this format in [MediaInfo](https://MediaArea.net/MediaInfo), would you mind to provide one? (I second dericed about matroska-test-files repo)

hwrenx:
Raw avs2 samples could be found [here](https://gitee.com/hwren/avs2_samples), avs2 for matroska-test-files will be added later.

codec_specs.md
```diff
@@ -263,6 +263,16 @@ Description: FFV1 is a lossless intra-frame video encoding format designed to ef
 
 Initialisation: For FFV1 versions 0 or 1, `Private Data` SHOULD NOT be written. For FFV1 version 3 or greater, the `Private Data` MUST contain the FFV1 Configuration Record structure, as defined in https://tools.ietf.org/html/draft-niedermayer-cellar-ffv1-01#section-4.1, and no other data.
 
+### V_AVS2_P2
+
+Codec ID: V_AVS2
```
mbunkus: For the other codecs list here the Codec ID and the heading are always identical. Therefore I don't understand what that `_P2` suffix is supposed to mean.

codec_specs.md
```diff
@@ -625,6 +635,16 @@ Description: The Wavpack packets consist of a stripped header followed by the fr
 
 Initialisation: none
 
+### A_AVS2_P3
+
+Codec ID: A_AVS2
```
mbunkus: Same as above, really. Why is there the `_P3` suffix in the heading?

codec_specs.md
hwrenx: The P2 and P3 suffix were added to distinguish two different part of AVS2 standard (the video part and audio part). Well, it's looks like there is no need to do that. I will modify and re-submit them soon.

robUx4:
LGTM if you just want to add the CodecID for now.

So far we haven't decided the proper way to define codec thoroughly although the AV1 doc should be a good starting point. Unless the codec has a properly defined "packetization" in which case it should be mentioned.

mcr:
Should be in an extension document.

PR #250, opened 2018-07-25
Add sections to mkv elements

dericed:
As suggested by @kieranjol at https://mailarchive.ietf.org/arch/msg/cellar/pyJxP5txAFOkhyVtvWJ6s04qa5g.


Makefile
```diff
@@ -25,25 +25,22 @@ $(OUTPUT_TAGS).md: index_tags.md tagging.md matroska_tagging_registry.md tagging
 	cat $^ > $@
 
 %.xml: %.md
-	mmark -xml2 -page $< > $@
+	mmark -xml2 -page $< | awk '/<?rfc toc=/ && !modif { printf("<?rfc tocdepth=\"12\"?>\n"); modif=1 } {print}' > $@
```
kieranjol: I see a value of 12 for tocdepth here, but the example I saw on gist goes six elements deep. Will this patch produce a tocdepth of 6? I can't test right now unfortunately as I'm about to leave work.

dericed: Ya it fails to use depths above 6. I think this may be related to html which only defines down to <h6>. h7 and so on are invalid. We could squash the other elements back to h6. Possibly we could elevate the whole section since it already starts at h3.

kieranjol:
I still find the depth of six to be very useful, I think that the extra
depth could end up looking a bit messy potentially.

dericed:
I updated the pr to move the elements up a level (with Segment starting at ## rather than ###). Also the sectional depth is now limited to 6.

kieranjol:
I just tested the update, it's really lovely. Having that extra layer
available in the ToC is so handy for scanning over the Video sub-elements,
for example.
I think it makes the document easier to understand anyhow.
Thanks so much for adding this so quick.

dericed:
yes the xml has the elements already sorted, so the transformation doesn't add any additional sorting.

PR #256, opened 2018-08-10
Use ITU BT.723 for matrix coefs

robUx4:
It's free !


ebml_matroska.xml
```diff
@@ -419,19 +419,23 @@
...
       <enum value="3" label="reserved"/>
       <enum value="4" label="FCC"/>
-      <enum value="5" label="BT470BG"/>
+      <enum value="5" label="ITU-R BT.470BG"/>
       <enum value="6" label="SMPTE 170M"/>
       <enum value="7" label="SMPTE 240M"/>
       <enum value="8" label="YCoCg"/>
       <enum value="9" label="BT2020 Non-constant Luminance"/>
       <enum value="10" label="BT2020 Constant Luminance"/>
+      <enum value="11" label="SMPTE 2085"/>
```
JeromeMartinez: SMPTE ST 2085 (SMPTE uses now "ST" for standards; see e.g. H.265 spec, it uses "ST" for this label).
Generally speaking, if you change old labels I suggest to fix all issues in one shot:
0 = Identity (may be used for XYZ too, which is not GBR)
4 = FCC 73.682
9&10 = ITU-R BT.2020 ...

and to have 2 commits (1 for fixing old labels and 1 for adding new labels)

PR #262, opened 2018-09-18
Sequence header same within sequence not bitstream

agrange:
Sequence headers must be identical within a sequence, but a bitstream may contain more than one sequence.


robUx4:
LGTM, thanks !

PR #263, opened 2018-09-19
Align encryption text with AV1 iso-bmff binding text

agrange:
This change is a suggestion to align the description of encryption with the text in the AV1 iso-bmff binding specification:
https://aomediacodec.github.io/av1-isobmff/


robUx4:
LGTM, thanks !

PR #266, opened 2018-09-26
Refine AV1 encryption details

robUx4:
* Use the WebM terminology
* define the parts that are not encrypted
* the first unprotected partition doesn't contain protected data


robUx4:
The non 16-bytes aligned parts at the beginning are now clearly define.

PR #271, opened 2018-10-02
add rotation element

dericed:
to address https://github.com/Matroska-Org/matroska-specification/issues/269

### DIFF
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -382,6 +382,9 @@
   <element name="PixelCropRight" path="0*1(\Segment\Tracks\TrackEntry\Video\PixelCropRight)" cppname="VideoPixelCropRight" id="0x54DD" type="uinteger" maxOccurs="1" minver="1" default="0">
     <documentation lang="en">The number of video pixels to remove on the right of the image.</documentation>
   </element>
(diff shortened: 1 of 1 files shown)


JeromeMartinez:
integer and in radians?
how do you indicate e.g. π radians (value for 180° rotation) with an integer?

dericed:
@JeromeMartinez rotate = `180*(PI/180)` .... ok i see your point. Prefer in degrees?

JeromeMartinez:
I usually prefer SI units (so radians with floating point number?) but it makes the most classic values (π, π/2, -π/2) approximations (IEEE representation of π is not exact).
From the issue ticket using the matrix values as in MP4 is not appreciated.
So maybe degrees is the "less worse" solution, -90/0/90/180 being the most used values.
I imagine to prefer a floating number in case of weird need of e.g. 0.1°, but maybe too much (and 4 bytes used instead of 1 or 2 bytes.
Do others have better idea?

dericed:
closing as discussed in https://github.com/Matroska-Org/matroska-specification/issues/269#issuecomment-426390778

robUx4:
Technically for 360 videos there are 3 degrees of movement and so 3 different values.

This one is the rotation on the Z axis (flat videos being X,Y).

PR #272, opened 2018-10-02
add implementation_notes on conditional element attributes

dericed:
This PR is a proposal for an alternative to https://github.com/Matroska-Org/matroska-specification/pull/236. In some case Element attributes are conditional or have some meaning too complex to store in an Element attribute. In this case for certain attributes (such as minOccurs and default), I propose storing a reserved string such as 'note' and then storing a `<implementation_notes>` section of the Element definition that clarifies that attribute. In this example I tried this for BlockDuration which has conditional minOccurs and default attributes.

Another example would be:
```xml
<element name="ColourSpace" path="0*1(\Segment\Tracks\TrackEntry\Video\ColourSpace)" cppname="VideoColourSpace" id="0x2EB524" type="binary" minOccurs="note" maxOccurs="1" minver="1" webm="0" size="4">
  <documentation lang="en">Specify the pixel format used for the Track's data as a FourCC. This value is similar in scope to the biCompression value of AVI's BITMAPINFOHEADER.</documentation>
  <implementation_notes>
    <minOccurs>ColourSpace is MANDATORY in TrackEntry when the CodecID Element of the TrackEntry is set to "V_UNCOMPRESSED".</minOccurs>
  </implementation_notes>
</element>
```

Interesting in comments on this approach before testing it with the entire schema.

### DIFF
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -1118,7 +1118,7 @@
   </element>
   <element name="ChapterSegmentUID" path="0*1(\Segment\Chapters\EditionEntry\ChapterAtom\ChapterSegmentUID)" id="0x6E67" type="binary" range="&gt;0" length="16" maxOccurs="1">
     <documentation lang="en" purpose="definition">The SegmentUID of another Segment to play during this chapter.</documentation>
-    <documentation lang="en" purpose="usage notes">ChapterSegmentUID is mandatory if ChapterSegmentEditionUID is used.</documentation>
+    <implementation_note note_type="minOccurs">ChapterSegmentUID is MANDATORY if ChapterSegmentEditionUID is used.</implementation_note>
     <extension webm="0"/>
   </element>
   <element name="ChapterSegmentEditionUID" path="0*1(\Segment\Chapters\EditionEntry\ChapterAtom\ChapterSegmentEditionUID)" id="0x6EBC" type="uinteger" range="not 0" maxOccurs="1">
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -963,7 +963,8 @@
     <extension webm="0"/>
   </element>
   <element name="Cues" path="0*1(\Segment\Cues)" id="0x1C53BB6B" type="master" maxOccurs="1">
-    <documentation lang="en" purpose="definition">A Top-Level Element to speed seeking access. All entries are local to the Segment. This Element SHOULD be mandatory for non <a href="https://www.matroska.org/technical/streaming/index.hmtl">"live" streams</a>.</documentation>
+    <documentation lang="en" purpose="definition">A Top-Level Element to speed seeking access. All entries are local to the Segment.</documentation>
+    <implementation_note note_type="minOccurs">This Element SHOULD be mandatory when the Segment is not transmitted as a Livestreaming (see #livestreaming).</implementation_note>
   </element>
   <element name="CuePoint" path="1*(\Segment\Cues\CuePoint)" id="0xBB" type="master" minOccurs="1">
     <documentation lang="en" purpose="definition">Contains all information relative to a seek point in the Segment.</documentation>
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -517,8 +517,9 @@
     </restriction>
     <extension cppname="VideoAspectRatio"/>
   </element>
-  <element name="ColourSpace" path="0*1(\Segment\Tracks\TrackEntry\Video\ColourSpace)" id="0x2EB524" type="binary" length="4" maxOccurs="1">
-    <documentation lang="en" purpose="definition">Specify the pixel format used for the Track's data as a FourCC. This value is similar in scope to the biCompression value of AVI's BITMAPINFOHEADER. This Element is MANDATORY in TrackEntry when the CodecID Element of the TrackEntry is set to "V_UNCOMPRESSED".</documentation>
+  <element name="ColourSpace" path="0*1(\Segment\Tracks\TrackEntry\Video\ColourSpace)" id="0x2EB524" type="binary" length="4" minOccurs="see note" maxOccurs="1">
+    <documentation lang="en" purpose="definition">Specify the pixel format used for the Track's data as a FourCC. This value is similar in scope to the biCompression value of AVI's BITMAPINFOHEADER.</documentation>
+    <implementation_note note_type="minOccurs">ColourSpace is MANDATORY in TrackEntry when the CodecID Element of the TrackEntry is set to "V_UNCOMPRESSED".</implementation_note>
     <extension webm="0"/>
     <extension cppname="VideoColourSpace"/>
   </element>
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -158,8 +158,10 @@
     <documentation lang="en" purpose="definition">Interpreted by the codec as it wishes (using the BlockAddID).</documentation>
     <extension webm="1"/>
   </element>
-  <element name="BlockDuration" path="0*1(\Segment\Cluster\BlockGroup\BlockDuration)" id="0x9B" type="uinteger" default="DefaultDuration" maxOccurs="1">
-    <documentation lang="en" purpose="definition">The duration of the Block (based on TimestampScale). This Element is mandatory when DefaultDuration is set for the track (but can be omitted as other default values). When not written and with no DefaultDuration, the value is assumed to be the difference between the timestamp of this Block and the timestamp of the next Block in "display" order (not coding order). This Element can be useful at the end of a Track (as there is no other Block available), or when there is a break in a track like for subtitle tracks.</documentation>
+  <element name="BlockDuration" path="0*1(\Segment\Cluster\BlockGroup\BlockDuration)" id="0x9B" type="uinteger" default="see note" minOccurs="see note" maxOccurs="1">
+    <documentation lang="en" purpose="definition">The duration of the Block (based on TimestampScale). The BlockDuration Element can be useful at the end of a Track to define the duration of the last frame (as there is no subsequent Block available), or when there is a break in a track like for subtitle tracks.</documentation>
+    <implementation_note note_type="minOccurs">BlockDuration is a MANDATORY Element if the associated TrackEntry stores a DefaultDuration value.</implementation_note>
+    <implementation_note note_type="default">When not written and with no DefaultDuration, the value is assumed to be the difference between the timestamp of this Block and the timestamp of the next Block in "display" order (not coding order).</implementation_note>
   </element>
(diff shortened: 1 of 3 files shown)


mbunkus:
I kind of like this approach better than trying to shoehorn default values via paths. The "via paths" approach would lend itself to automatic processing, true, but I can think of a lot of situations in which rather complex conditions would have to be described (e.g. "if field X of header structure Y present in the avcC in CodecPrivate is set to value A…").

In the context of a specification: expressing the relationship in prose is easier, and all the other specs do it, too.

dericed:
@robUx4 any comment on this approach? Sufficient to use this approach to supersede #236. If ok, I'll adjust EBML first to add the new aspects of an Element Definition and then apply it here.

dericed:
I updated this PR to use the <implementation_note>, now supported via EBML, for elements with complex or conditional default and minOccurs. So far adding in BlockDuration, DisplayWidth, DisplayHeight, and OutputSamplingFrequency. I also updated the transforms to pass this data along to the compiled RFC.

In EBML it says that any implementation_note supersedes the attribute that it describes, however we don't say if anything should be done with that attribute, for now I added the text `see note` to those attributes. Open to other ideas.

transforms/ebml_schema2markdown4rfc.xsl
```diff
@@ -99,6 +99,19 @@
       <xsl:value-of select="."/>
       <xsl:text>&#xa;&#xa;</xsl:text>
     </xsl:for-each>
+    <xsl:if test="ebml:implementation_note">
+      <xsl:text>implementation notes:&#xa;&#xa;</xsl:text>
+      <xsl:text>|type|note|&#xa;</xsl:text>
```
robUx4: rather than `type` I'd prefer `attribute` as that's what the implementation note refers to. (I had to go back to the EBML spec to remember what the `note_type` was for)

ebml_matroska.xml
```diff
@@ -485,12 +487,14 @@
...
     <extension cppname="VideoPixelCropRight"/>
   </element>
-  <element name="DisplayWidth" path="0*1(\Segment\Tracks\TrackEntry\Video\DisplayWidth)" id="0x54B0" type="uinteger" range="not 0" default="PixelWidth - PixelCropLeft - PixelCropRight" maxOccurs="1">
-    <documentation lang="en" purpose="definition">Width of the video frames to display. Applies to the video frame after cropping (PixelCrop* Elements). The default value is only valid when <a href="https://www.matroska.org/technical/specs/index.html#DisplayUnit">DisplayUnit</a> is 0.</documentation>
+  <element name="DisplayWidth" path="0*1(\Segment\Tracks\TrackEntry\Video\DisplayWidth)" id="0x54B0" type="uinteger" range="not 0" default="see note" maxOccurs="1">
+    <documentation lang="en" purpose="definition">Width of the video frames to display. Applies to the video frame after cropping (PixelCrop* Elements).</documentation>
+    <implementation_note note_type="default">If the DisplayUnit is 0, then the default value for DisplayWidth is equal to PixelWidth - PixelCropLeft - PixelCropRight, else there is no default value.</implementation_note>
```
robUx4: We should either use the path of `DisplayUnit` or mention that's the `DisplayUnit` of the `TrackEntry` in which `DisplayWidth` belongs to. Same thing with the other attributes, so maybe rephrase to make it clear that's value from this `TrackEntry`.

ebml_matroska.xml
```diff
@@ -485,12 +487,14 @@
...
     <extension cppname="VideoDisplayWidth"/>
   </element>
-  <element name="DisplayHeight" path="0*1(\Segment\Tracks\TrackEntry\Video\DisplayHeight)" id="0x54BA" type="uinteger" range="not 0" default="PixelHeight - PixelCropTop - PixelCropBottom" maxOccurs="1">
-    <documentation lang="en" purpose="definition">Height of the video frames to display. Applies to the video frame after cropping (PixelCrop* Elements). The default value is only valid when <a href="https://www.matroska.org/technical/specs/index.html#DisplayUnit">DisplayUnit</a> is 0.</documentation>
+  <element name="DisplayHeight" path="0*1(\Segment\Tracks\TrackEntry\Video\DisplayHeight)" id="0x54BA" type="uinteger" range="not 0" default="see note" maxOccurs="1">
+    <documentation lang="en" purpose="definition">Height of the video frames to display. Applies to the video frame after cropping (PixelCrop* Elements).</documentation>
+    <implementation_note note_type="default">If the DisplayUnit is 0, then the default value for DisplayHeight is equal to PixelHeight - PixelCropTop - PixelCropBottom, else there is no default value.</implementation_note>
```
robUx4: same remark as for DisplayWidth

ebml_matroska.xml
```diff
@@ -513,8 +517,9 @@
     </restriction>
     <extension cppname="VideoAspectRatio"/>
   </element>
-  <element name="ColourSpace" path="0*1(\Segment\Tracks\TrackEntry\Video\ColourSpace)" id="0x2EB524" type="binary" length="4" maxOccurs="1">
-    <documentation lang="en" purpose="definition">Specify the pixel format used for the Track's data as a FourCC. This value is similar in scope to the biCompression value of AVI's BITMAPINFOHEADER. This Element is MANDATORY in TrackEntry when the CodecID Element of the TrackEntry is set to "V_UNCOMPRESSED".</documentation>
+  <element name="ColourSpace" path="0*1(\Segment\Tracks\TrackEntry\Video\ColourSpace)" id="0x2EB524" type="binary" length="4" minOccurs="see note" maxOccurs="1">
+    <documentation lang="en" purpose="definition">Specify the pixel format used for the Track's data as a FourCC. This value is similar in scope to the biCompression value of AVI's BITMAPINFOHEADER.</documentation>
+    <implementation_note note_type="minOccurs">ColourSpace is MANDATORY in TrackEntry when the CodecID Element of the TrackEntry is set to "V_UNCOMPRESSED".</implementation_note>
```
robUx4: same remark as BlockDuration regarding the value of `minOccurs` that becomes 1 in certain conditions.

ebml_matroska.xml
```diff
@@ -767,8 +772,9 @@
     <documentation lang="en" purpose="definition">Sampling frequency in Hz.</documentation>
     <extension cppname="AudioSamplingFreq"/>
   </element>
-  <element name="OutputSamplingFrequency" path="0*1(\Segment\Tracks\TrackEntry\Audio\OutputSamplingFrequency)" id="0x78B5" type="float" range="&gt; 0x0p+0" default="SamplingFrequency" maxOccurs="1">
+  <element name="OutputSamplingFrequency" path="0*1(\Segment\Tracks\TrackEntry\Audio\OutputSamplingFrequency)" id="0x78B5" type="float" range="&gt; 0x0p+0" default="see note" maxOccurs="1">
     <documentation lang="en" purpose="definition">Real output sampling frequency in Hz (used for SBR techniques).</documentation>
+    <implementation_note note_type="default">The default value for OutputSamplingFrequency is equal to the SampllingFrequency.</implementation_note>
```
robUx4: *SampllingFrequency* typo

robUx4: I think we should also mention that's the `SamplingFrequency` from the same `TrackEntry`.

ebml_matroska.xml
```diff
@@ -1111,7 +1118,7 @@
   </element>
   <element name="ChapterSegmentUID" path="0*1(\Segment\Chapters\EditionEntry\ChapterAtom\ChapterSegmentUID)" id="0x6E67" type="binary" range="&gt;0" length="16" maxOccurs="1">
     <documentation lang="en" purpose="definition">The SegmentUID of another Segment to play during this chapter.</documentation>
-    <documentation lang="en" purpose="usage notes">ChapterSegmentUID is mandatory if ChapterSegmentEditionUID is used.</documentation>
+    <implementation_note note_type="minOccurs">ChapterSegmentUID is MANDATORY if ChapterSegmentEditionUID is used.</implementation_note>
```
robUx4: same remark as BlockDuration regarding the value of minOccurs that becomes 1 in certain conditions.

robUx4:
This PR should be updated if https://github.com/cellar-wg/ebml-specification/pull/288 is merged.

dericed:
And https://github.com/cellar-wg/ebml-specification/pull/288 was merged, so yes it should be :)

ebml_matroska.xml
```diff
@@ -158,8 +158,10 @@
...
+  <element name="BlockDuration" path="0*1(\Segment\Cluster\BlockGroup\BlockDuration)" id="0x9B" type="uinteger" default="see note" minOccurs="see note" maxOccurs="1">
+    <documentation lang="en" purpose="definition">The duration of the Block (based on TimestampScale). The BlockDuration Element can be useful at the end of a Track to define the duration of the last frame (as there is no subsequent Block available), or when there is a break in a track like for subtitle tracks.</documentation>
+    <implementation_note note_type="minOccurs">BlockDuration is a MANDATORY Element if the associated TrackEntry stores a DefaultDuration value.</implementation_note>
```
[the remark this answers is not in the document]
dericed: Hmmm, part of the reason that I add `minOccurs="see note"` is because the EBML spec says:
> If the minOccurs attribute is not present then that EBML Element has a minOccurs value of 0.

So either we need to say:
> If the minOccurs attribute is not present and that Element has no implementation_note with a note_attribute set to "minOccurs", then that EBML Element has a minOccurs value of 0.

Or, the EBML specification could have a reserved value ("see note" or something else), to note that the implementation_note defines the minOccurs rather than this attribute.

ebml_matroska.xml
dericed: resolved in https://github.com/cellar-wg/matroska-specification/pull/272/commits/84bf545b0bf19b1fce1b143cbaf25d6f287b67d4

ebml_matroska.xml
dericed: resolved in https://github.com/cellar-wg/matroska-specification/pull/272/commits/84bf545b0bf19b1fce1b143cbaf25d6f287b67d4

transforms/ebml_schema2markdown4rfc.xsl
dericed: resolved in https://github.com/cellar-wg/matroska-specification/pull/272/commits/0dba7599f70d3c074395c383d4fb5c061afff7f7 since I merged your change to EBML

dericed:
I think all comments are resolved except:

- there is a conflict is meaning between the minOccurs attribute not being written which means 0 (per EBML spec), and the implementation_note saying otherwise. The implementation_note does say it supersedes minOccurs but I think this should be acknowledged in the minOccurs section of EBML.

robUx4:
>     * there is a conflict is meaning between the minOccurs attribute not being written which means 0 (per EBML spec), and the implementation_note saying otherwise. The implementation_note does say it supersedes minOccurs but I think this should be acknowledged in the minOccurs section of EBML.

I think we should clarify that first in the EBML specs, and then fix it here when the solution is approved.

Issue #257, opened 2018-08-28
Use `recurring` attribute on some level 1 elements


robUx4:
As discussed here https://github.com/Matroska-Org/ebml-specification/issues/176

dericed:
unless I'm missing something this is closed by https://github.com/cellar-wg/matroska-specification/pull/275.

PR #275, opened 2018-10-02
use recurring attribute

dericed:
resolves https://github.com/Matroska-Org/ebml-specification/issues/176


robUx4:
Also see #257

robUx4:
LGTM

dericed:
Forgot to use the recurring attribute in the transformation to make the element list in the rfc. updated now.

dericed:
going to merge today unless there's more comments otherwise

PR #276, opened 2018-10-02
add block metadata structure

dericed:
addresses https://github.com/Matroska-Org/matroska-specification/issues/270

### DIFF
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -142,6 +142,21 @@
   <element name="ReferenceVirtual" path="0*1(\Segment\Cluster\BlockGroup\ReferenceVirtual)" id="0xFD" type="integer" maxOccurs="1" minver="0" maxver="0" webm="0">
     <documentation lang="en">The Segment Position of the data that would otherwise be in position of the virtual block.</documentation>
   </element>
+  <element name="Metadata" path="0*(\Segment\Cluster\BlockGroup\Metadata)" id="0xD2" type="master" minver="4" webm="0">
+    <documentation lang="en">A named and stored Metadata value that references the BlockGroup.</documentation>
+  </element>
+  <element name="MetadataLaceNumber" path="1*1(\Segment\Cluster\BlockGroup\MetadataLaceNumber)" id="0x4F12" type="uinteger" minOccurs="1" minver="4" webm="0" default="0">
+    <documentation lang="en">Provide the number minus one of the laced frame that the following key/value Metadata pair applies to within the BlockGroup.</documentation>
+  </element>
+  <element name="MetadataName" path="1*1(\Segment\Cluster\BlockGroup\MetadataName)" id="0xD3" type="string" minOccurs="1" maxOccurs="1" minver="4" webm="0">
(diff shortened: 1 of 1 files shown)


dericed:
Updated to @mbunkus' suggestions, also removed some unneeded defaults.

mbunkus:
What about renaming them? :)

mbunkus:
And I meant giving the `Metadata` master a one-byte ID, too. It will be present for each metadata entry, just like `MetadataName` and one of `MetadataString`/`MetadataBinary`; hence it deserves a short ID, too.

dericed:
thx @mbunkus, updated

dericed:
@mbunkus, ok with BlockMetadata rather than FrameMetadata? Block* fits in context here better.

dericed:
For #3 perhaps I can add a section to tagging.md that documents official MetadataName values. Taking the same approach as official/unofficial TagNames may make sense here.

hubblec4:
I have read ever thing (email, issue, PR). It is very interesting and I understand a bit, because it is similar to the Tags.

I also think we need this Metadata system to preserve all info/data. +1

Where do I start best? I thought about that a little bit.

Mosu is right with the name and prefix: Only metadata would be too inaccurate.
Frame/BlockMetadata would be more accurate for the current project, because it could be, we need in the future
in another place in Matroska also metadata.

And I think of a job as well. For an edition. The edition is very often given a name nowadays.
Since there is no EditionName element you have to fall back on the Tags, which is a bit awkward when parsing.
Likewise it could happen that a user accidentally deletes / overwrites the Tags and already all edition names are gone.

I can well imagine that metadata in other places are very helpful.
At the Matroska meeting a woman told me that some audio data was lost while saving in Matroska.

Therefore, my consideration to make the Metadata element as a global element.

You would then only have a list for official/unofficial MetaNames like the TagNames.
Otherwise we would have to create a new list each time for a new Metadata system (for example AudioMetadata).

hubblec4:
The MKVToolNix Track Statistics Tags is a very helpful feature, and it could be also stored in a Metadata Element for each track. I can imagine that would be better for a parser.

mjbshaw:
Personally, I'd rather just use `BlockAdditions` for per-frame side metadata. After all, that's what it's for, isn't it?

dericed:
I agree with @mjbshaw and think we can extend the use of BlockAdditions to cover other types of side frame data. Originally I thought this wasn't possible since the meaning of BlockAdditionId was handled by the codec mapping, with no reservation for the container to declare it's own ids. However from seeing how BlockAdditions has been used, I think it's safe to define BlockAdditionId=1 as in the context of the codec mapping (lossless part for WAVPACK4, alpha for VP8/VP9/others?). And then from 2 and up can be used for new mappings.

I think I prefer to close this without merge and move the effort to https://github.com/Matroska-Org/matroska-specification/pull/287.

robUx4:
So we close the pull request ? I'm in favor of that.

dericed:
Yes, I'll close. The scope of work has moved to https://github.com/Matroska-Org/matroska-specification/pull/287.

PR #277, opened 2018-10-10
Av1 encryption sections

robUx4:
The WebM spec talks about partitions (the physical way of storing clear/encrypted parts) and sections which are part of the codec data.

This PR uses these 2 notions for a more refined and generic way of describing the 16-bytes end aligned encrypted buffers. The sections are merged to form one or more partition (clear or encrypted). An encrypted partition is not created for tiles that are too small. These are merged with the clear sections before them.

The Metadata OBU is also set to use the 16-bytes end aligned encrypted buffers system.


robUx4:
This branch was mistakenly pushed to master...

PR #221, opened 2018-01-17
Drop the word "Guidelines" from the element ordering section

mjbshaw:
Many of the things in that section are requirements, not guidelines.

PR #278, opened 2018-10-11
assume values not found in WebM have their default value

robUx4:
WebM doesn't use all our elements but sometimes they are implied even though
they don't define it. For example ContentEncodingOrder.

The elements must not be in WebM files (for compliance) but a parser may need
the actual value (of ContentEncodingOrder) to make any sense. So we assume
elements have their default values if they are not found.

If this is too loose, we may have to list the elements not in WebM but which
default value should be used.

### DIFF
--- a/matroska_schema_section_header.md
+++ b/matroska_schema_section_header.md
@@ -5,7 +5,7 @@ In addition to the EBML Schema definition provided by the EBML Specification, Ma
 
 | attribute name | required | definition |
 |:---------------|:---------|:-----------|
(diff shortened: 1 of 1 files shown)


robUx4:
For the record `ContentEncodingOrder` *is* defined for WebM. But the general principle stands. Even for new elements we might add.

mkver:
The problem that this issue is about not only exists between closely related `EBML Schemas` like Matroska and Webm, but also between different version of the same `EBML Schema`. See [EBML-issue #221](https://github.com/Matroska-Org/ebml-specification/issues/221).

robUx4:
Technically `Matroska` and `WebM` are 2 different DocTypes so there's not really direct compatibility at the EBML level. Especially the DocTypeVersion may not match at all (I don't even know what's the current value for WebM), neither minver and maxver. IMO a separate EBML Schema should be done for WebM. And then we can wonder if we actually want to define WebM at the IETF, this is not out responsibility.

So I think we should drop the "webm" attribute of the EBML Schema (which is not a valid attribute anyway) and keep on for WebM, which we may never use.

PR #281, opened 2018-10-11
always have a ContentEncAlgo in there is a ContentEncryption

robUx4:
It doesn't make sense if there isn't one. And it has a default value anyway.


mkver:
I am wondering whether the commit actually did what it was supposed to do or whether you changed the wrong element:
a) According to [your email](https://mailarchive.ietf.org/arch/msg/cellar/ui7i9YqCe8xZN6JSiCfOuOF8ljE) that referenced this pull request you actually wanted to make `ContentEncAlgo` and not `ContentEncryption` mandatory.
b) Your description (you meant "Always have a `ContentEncAlgo` if there is a `ContentEncryption`", didn't you?) said the same thing -- at least in my interpretation of it.
c) `ContentEncryption` doesn't have a default value, whereas `ContentEncAlgo` has one.
d) It makes perfect sense not to have a `ContentEncoding` without a `ContentEncryption`.

PS: The [corresponding commit for the in foundation-source](https://github.com/Matroska-Org/foundation-source/commit/727b0a98991bf2037e8720a2620573d39b6dcf10) makes mkclean add unnecessary `ContentEncryption` elements.

robUx4:
You're right !

PR #285, opened 2018-10-30
…LanguageIETF: adjust C++ names to match non-IETF variants

mbunkus:
This adjusts the name of the generated C++ classes so that they mirror
how the non-IETF variants of language elements are named.


mbunkus:
Meh. We released libMatroska with the current (inconsistent) names back in March. Even though I highly doubt there's any code out there that uses those names, I still don't like changing them retroactively. Closing the PR.

PR #287, opened 2018-11-18
add BlockAdditionMapping to describe each BlockAdditional data

robUx4:
This is similar to a TrackID + CodecID + CodecPrivate + TrackName but for extra frame data.

The `Timecode` and `RAWCooked` values may be removed and done in a separate spec. They need to define properly how to interpret the `BlockAddIDExtraData` binary data.

### DIFF
--- a/codec_specs.md
+++ b/codec_specs.md
@@ -655,7 +655,7 @@ Codec Name: [WavPack](http://www.wavpack.com/) lossless audio compressor
 
 Description: The Wavpack packets consist of a stripped header followed by the frame data. For multi-track (> 2 tracks) a frame consists of many packets. For more details, check the [WavPack muxing description](wavpack.html).
 
-Codec BlockAdditions: For hybrid A_WAVPACK4 encodings (that include a lossy encoding with a supplemental correction to produce a lossless encoding), the correction part is stored in BlockAdditional.
+Codec BlockAdditions: For hybrid `A_WAVPACK4` encodings (that include a lossy encoding with a supplemental correction to produce a lossless encoding), the correction part is stored in BlockAdditional.
 
 Initialization: none
 
--- a/codec_specs.md
+++ b/codec_specs.md
@@ -48,7 +48,7 @@ Each encoding supported for storage in Matroska MUST have a defined Initializati
 
 ### Codec BlockAdditions
 
-Additional data MAY be stored within a `BlockMore Element` of the `BlockAdditional Element` that is passed to the decoder along with the content of the `Block Element`. Each `BlockAdditional` is coupled with a `BlockAddID` that helps identify the kind of data it contains. The following table defines the meanings of `BlockAddID` values.
+Additional data that contextualizes or supplements a `Block` can be stored within the `BlockAdditional Element` of a `BlockMore Element`. This `BlockAdditional` data MAY be passed to the associated decoder along with the content of the `Block Element`. Each `BlockAdditional` is coupled with a `BlockAddID` that identifies the kind of data it contains. The following table defines the meanings of `BlockAddID` values.
 
 BlockAddID Value | Definition
 -----------------|:---------------
@@ -56,9 +56,9 @@ BlockAddID Value | Definition
 1                | Indicates that the context of the `BlockAdditional` data is defined by the corresponding `Codec Mapping`.
 2 or greater     | `BlockAddID` values of 2 and greater are mapped to the `BlockAddIDValue` of the `BlockAdditionMapping` of the associated Track.
 
-The values of `BlockAddID` that are 2 of greater have no semantic meaning, but simply associate the `BlockMore Element` with a `BlockAdditionMapping` of the associated Track. See [the section on Block Additional Mappings](#block-additional-mappings) for more information
+The values of `BlockAddID` that are 2 of greater have no semantic meaning, but simply associate the `BlockMore Element` with a `BlockAdditionMapping` of the associated Track. See [the section on Block Additional Mappings](#block-additional-mappings) for more information.
 
-The following XML depicted the nested Elements of a `BlockGroup Element` depicting an example of Codec BlockAdditions:
+The following XML depicts the nested Elements of a `BlockGroup Element` with an example of BlockAdditions:
 
 ```xml
 <BlockGroup>
@@ -66,7 +66,7 @@ The following XML depicted the nested Elements of a `BlockGroup Element` depicti
   <BlockAdditions>
     <BlockMore>
       <BlockAddID>1</BlockAddID>
-      <BlockAdditional>{alpha channel enconding to supplement the VP9 frame}</BlockAdditional>
+      <BlockAdditional>{alpha channel encoding to supplement the VP9 frame}</BlockAdditional>
     </BlockMore>
   </BlockAdditions>
 </BlockGroup>
--- a/codec_specs.md
+++ b/codec_specs.md
@@ -48,7 +48,15 @@ Each encoding supported for storage in Matroska MUST have a defined Initializati
 
 ### Codec BlockAdditions
 
-Additional data MAY be stored within a `BlockMore Element` of the `BlockAdditional Element` that is passed to the decoder along with the content of the `Block Element`. The `BlockAddID` value of `1` is reserved to indicate that the content of the `BlockAdditional Element` is defined by the corresponding `Codec Mapping`.
+Additional data MAY be stored within a `BlockMore Element` of the `BlockAdditional Element` that is passed to the decoder along with the content of the `Block Element`. Each `BlockAdditional` is coupled with a `BlockAddID` that helps identify the kind of data it contains. The following table defines the meanings of `BlockAddID` values.
+
+BlockAddID Value | Definition
+-----------------|:---------------
+0                | Invalid.
+1                | Indicates that the context of the `BlockAdditional` data is defined by the corresponding `Codec Mapping`.
+2 or greater     | `BlockAddID` values of 2 and greater are mapped to the `BlockAddIDValue` of the `BlockAdditionMapping` of the associated Track.
+
+The values of `BlockAddID` that are 2 of greater have no semantic meaning, but simply associate the `BlockMore Element` with a `BlockAdditionMapping` of the associated Track. See [the section on Block Additional Mappings](#block-additional-mappings) for more information
 
 The following XML depicted the nested Elements of a `BlockGroup Element` depicting an example of Codec BlockAdditions:
 
--- a/codec_specs.md
+++ b/codec_specs.md
@@ -56,7 +56,7 @@ BlockAddID Value | Definition
 1                | Indicates that the context of the `BlockAdditional` data is defined by the corresponding `Codec Mapping`.
 2 or greater     | `BlockAddID` values of 2 and greater are mapped to the `BlockAddIDValue` of the `BlockAdditionMapping` of the associated Track.
 
-The values of `BlockAddID` that are 2 of greater have no semantic meaning, but simply associate the `BlockMore Element` with a `BlockAdditionMapping` of the associated Track. See [the section on Block Additional Mappings](#block-additional-mappings) for more information.
+The values of `BlockAddID` that are 2 of greater have no semantic meaning, but simply associate the `BlockMore Element` with a `BlockAdditionMapping` of the associated Track. See [the section on Block Additional Mappings](#block-additional-mapping) for more information.
 
 The following XML depicts the nested Elements of a `BlockGroup Element` with an example of BlockAdditions:
 
--- a/codec_specs.md
+++ b/codec_specs.md
@@ -269,7 +269,7 @@ Codec Name: VP8 Codec format
 
 Description: VP8 is an open and royalty free video compression format developed by Google and created by On2 Technologies as a successor to VP7. [@!RFC6386]
 
-Codec BlockAdditions: A single-channel encoding of an alpha channel MAY be stored in `BlockAdditions`.
+Codec BlockAdditions: A single-channel encoding of an alpha channel MAY be stored in `BlockAdditions`. The `BlockAddId` of the `BlockMore` containing these data MUST be 1.
 
 Initialization: none
 
@@ -281,7 +281,7 @@ Codec Name: VP9 Codec format
 
 Description: VP9 is an open and royalty free video compression format developed by Google as a successor to VP8. [Draft VP9 Bitstream and Decoding Process Specification](https://www.webmproject.org/vp9/)
 
-Codec BlockAdditions: A single-channel encoding of an alpha channel MAY be stored in `BlockAdditions`.
+Codec BlockAdditions: A single-channel encoding of an alpha channel MAY be stored in `BlockAdditions`. The `BlockAddId` of the `BlockMore` containing these data MUST be 1.
 
 Initialization: none
 
@@ -655,7 +655,7 @@ Codec Name: [WavPack](http://www.wavpack.com/) lossless audio compressor
 
 Description: The Wavpack packets consist of a stripped header followed by the frame data. For multi-track (> 2 tracks) a frame consists of many packets. For more details, check the [WavPack muxing description](wavpack.html).
 
-Codec BlockAdditions: For hybrid `A_WAVPACK4` encodings (that include a lossy encoding with a supplemental correction to produce a lossless encoding), the correction part is stored in BlockAdditional.
+Codec BlockAdditions: For hybrid `A_WAVPACK4` encodings (that include a lossy encoding with a supplemental correction to produce a lossless encoding), the correction part is stored in BlockAdditional. The `BlockAddId` of the `BlockMore` containing these data MUST be 1.
 
 Initialization: none
 
--- a/codec_specs.md
+++ b/codec_specs.md
@@ -46,6 +46,24 @@ An optional description for the encoding. This value is only intended for human 
 
 Each encoding supported for storage in Matroska MUST have a defined Initialization. The Initialization MUST describe the storage of data necessary to initialize the decoder, which MUST be stored within the `CodecPrivate Element`. When the Initialization is updated within a track then that updated Initialization data MUST be written into the `CodecState Element` of the first `Cluster` to require it. If the encoding does not require any form of Initialization then `none` MUST be used to define the Initialization and the `CodecPrivate Element` SHOULD NOT be written and MUST be ignored. Data that is defined Initialization to be stored in the `CodecPrivate Element` is known as `Private Data`.
 
+### Codec BlockAdditions
+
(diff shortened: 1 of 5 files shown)


JeromeMartinez:
IMO it should be in https://github.com/Matroska-Org/matroska-specification/pull/276/files directly.

enum restriction: I am currently not in favor to restrict the list, as we don't limits e.g. tags.

robUx4:
> IMO it should be in https://github.com/Matroska-Org/matroska-specification/pull/276/files directly.

What do you mean by "it" ? As discussed on CELLAR putting a name and differentiate string/binary for something that can already be done with BlockAdditional doesn't seem like the right way to go. The lace number may be an issue, but it can certainly be added to `BlockMore`.
 
> enum restriction: I am currently not in favor to restrict the list, as we don't limits e.g. tags.

The enum is open-ended, there can be other values. But if I'm going to mention that, then I'll probably remove `Timecode` and `RAWCooked` to mention that currently only value 0 is defined.

robUx4:
This PR solves the current situation where there are `BlockAddID` and `BlockAdditional` but it doesn't say what to do with the data. Now we say the current use is a complement of data to give to the codec. But there can be other cases of metadata attached to each frame that don't fall in this category.

The PRs are separated. #276 should be updated to use this new system (thus it comes after).

t-rapp:
>The enum is open-ended, there can be other values. But if I'm going to mention that, then I'll probably remove Timecode and RAWCooked to mention that currently only value 0 is defined.

I would appreciate it if Timecode would not be mentioned here. It looks like a premature standardization of timecode storage format to me. Rather allow users to make their own BlockAddIDType definition possibly with some reserved range for later standardization.

mjbshaw:
It would be nice to clarify how new `BlockAddIDType`s can be registered. We at Google are considering adding some new side metadata for frames within WebM (using `BlockAdditions`). I don't have any concrete details I can share about this metadata right now, but it would be nice to know what the procedure is for standardizing it so we (and others) can utilize that procedure.

dericed:
>> enum restriction: I am currently not in favor to restrict the list, as we don't limits e.g. tags.

> The enum is open-ended, there can be other values. But if I'm going to mention that, then I'll probably remove Timecode and RAWCooked to mention that currently only value 0 is defined.

The EBML specification implies that enum lists are restrictive, not open. https://github.com/Matroska-Org/ebml-specification/blob/master/specification.markdown#restriction-element

notes.md
```diff
@@ -241,3 +241,9 @@ When playing back a track using the `TrackTimestampScale`, if the track is being
...
 
 While the above example deals specifically with audio tracks, this element can be used to align video, audio, subtitles, or any other type of track contained in a Matroska file.
+
+## BlockAdditionMapping
+
+Extra data or metadata can be added to each frame independenly using `BlockAdditional` data. Each `BlockAdditional` is coupled with a `BlockAddID` that helps identify the kind of data it contains. The description of such data is done with the elements in `BlockAdditionMapping`.
+
+So far only the `BlockAddIDType` value of 0 is defined. It means the content of `BlockAdditional` is passed to the track decoder in addition of the frame data. It is up to the decoder to use the "complement" data or not and how to use them. If the `AlphaMode` flag is set, that means the extra data define an alpha layer for the video, for example in WebM it's a secondary VP8/VP9 stream.
```
dericed: Per https://github.com/Matroska-Org/matroska-specification/blob/98c65bfccf38df4b1b2b7fdfeb9a0e2cba1ca979/ebml_matroska.xml#L164, the value `0` references the contents of the main block.

Also, IIUC, the codec mapping for [A_WAVPACK4](https://github.com/Matroska-Org/matroska-specification/blob/master/codec_specs.md#a_wavpack4), reserves the value of `1` for an id.

codec_specs.md
```diff
@@ -46,6 +46,24 @@ An optional description for the encoding. This value is only intended for human
 
 Each encoding supported for storage in Matroska MUST have a defined Initialisation. The Initialisation MUST describe the storage of data necessary to initialise the decoder, which MUST be stored within the `CodecPrivate Element`. When the Initialisation is updated within a track then that updated Initialisation data MUST be written into the `CodecState Element` of the first `Cluster` to require it. If the encoding does not require any form of Initialisation then `none` MUST be used to define the Initialisation and the `CodecPrivate Element` SHOULD NOT be written and MUST be ignored. Data that is defined Initialisation to be stored in the `CodecPrivate Element` is known as `Private Data`.
 
+### Codec BlockAdditions
+
+Additional data MAY be stored within a `BlockMore Element` of the `BlockAdditional Element` that is passed to the decoder along with the content of the `Block Element`. The `BlockAddID` value of `1` is reserved to indicate that the content of the `BlockAdditional Element` is defined by the corresponding `Codec Mapping`.
+
+The following XML depicted the nested Elements of a `BlockGroup Element` depicting an example of Codec BlockAdditions:
```
robUx4: Not sure about the wording here.

codec_specs.md
```diff
@@ -46,6 +46,24 @@ An optional description for the encoding. This value is only intended for human
...
+
+```xml
+<BlockGroup>
+  <Block>{Binary data of a VP9 video frame in YUV}</Block>
+  <BlockAdditions>
+    <BlockMore>
+      <BlockAddID>1</BlockAddID>
+      <BlockAdditional>{alpha channel enconding to supplement the VP9 frame}</BlockAdditional>
```
robUx4: `encoding`

codec_specs.md
```diff
@@ -623,7 +645,9 @@ Codec ID: A_WAVPACK4
 
 Codec Name: [WavPack](http://www.wavpack.com/) lossless audio compressor
 
-Description: The Wavpack packets consist of a stripped header followed by the frame data. For multi-track (> 2 tracks) a frame consists of many packets. For hybrid files (lossy part + correction part), the correction part is stored in an additional block (level 1). For more details, check the [WavPack muxing description](wavpack.html).
+Description: The Wavpack packets consist of a stripped header followed by the frame data. For multi-track (> 2 tracks) a frame consists of many packets. For more details, check the [WavPack muxing description](wavpack.html).
+
+Codec BlockAdditions: For hybrid A_WAVPACK4 encodings (that include a lossy encoding with a supplemental correction to produce a lossless encoding), the correction part is stored in BlockAdditional.
```
robUx4: backtick A_WAVPACK4

robUx4: Maybe we need to mention it's the `BlockAdditional` data marked with `BlockAddID`=1 ? (same for VP8 and VP9).

ebml_matroska.xml
```diff
@@ -244,6 +244,26 @@
...
+    <documentation lang="en">The <a href="https://www.matroska.org/technical/specs/index.html#BlockAddID">BlockAddID</a> value being described. To keep MaxBlockAdditionID as low as possible, small values SHOULD be used.</documentation>
+  </element>
+  <element name="BlockAddIDName" path="0*1(\Segment\Tracks\TrackEntry\BlockAdditionMapping\BlockAddIDName)" id="0x41A4" type="string" minOccurs="0" maxOccurs="1" minver="4" webm="0">
+    <documentation lang="en">A human-friendly name describing the type of BlockAdditional data.</documentation>
+  </element>
+  <element name="BlockAddIDType" path="1*1(\Segment\Tracks\TrackEntry\BlockAdditionMapping\BlockAddIDType)" id="0x41E7" type="uinteger" minOccurs="1" maxOccurs="1" minver="4" webm="0" default="0">
+    <documentation lang="en">How the BlockAdditional data should be handled.</documentation>
+    <restriction>
+      <enum value="0" label="Codec Complement data (passed to the codec handler with each Frame)"/>
+      <enum value="0x2122A9" label="Timecode data (format defined in BlockAddIDExtraData)"/>
```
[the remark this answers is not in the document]
robUx4: We don't need the enum given your commit on the `BlockAdditionMapping` text ?

dericed:
@robUx4 can you review the approach I'm testing in https://github.com/cellar-wg/matroska-specification/pull/287/commits/3d5e52d45b742e0be8ea7c1cc683b5d9ee1a6706. This moves the Block Additional Mappings to be handled similarly to Codec Mappings, rather than enumerating them in the schema.

dericed:
@robUx4 I updated this PR to start a block_additional_mapping (timecode as a string in this example) as a standalone document. I think the timecode document could be expanded to allow BlockAddIDExtraData to support different types of timecode such as expressions of SMPTE 12M or QuickTime's timecode descriptor of QuickTime Timecode Sample Data via https://github.com/cellar-wg/matroska-specification/pull/112/files.

However in this PR, timecode is mostly here as a first example while the PR more so focuses on the Block Additional Mapping section and how to define the mappings.

notes.md
```diff
@@ -253,3 +253,11 @@ When playing back a track using the `TrackTimestampScale`, if the track is being
 It would be possible for a `Matroska Player` to also adjust the audio's samplerate at the same time as adjusting the timestamps if you wanted to play the two audio streams synchronously. It would also be possible to adjust the video to match the audio's speed. However, for playback, the selected track(s) timestamps SHOULD be adjusted if they need to be scaled.
 
 While the above example deals specifically with audio tracks, this element can be used to align video, audio, subtitles, or any other type of track contained in a Matroska file.
+
+## Block Additional Mapping
+
+Extra data or metadata can be added to each frame independenly using `BlockAdditional` data. Each `BlockAdditional` is coupled with a `BlockAddID` that helps identify the kind of data it contains. The description of such data is done with the elements in `BlockAdditionMapping`.
```
robUx4: independently typo

notes.md
```diff
@@ -253,3 +253,11 @@ When playing back a track using the `TrackTimestampScale`, if the track is being
...
+
+A `BlockAddIDType` value of 0 is defined to mean that the content of `BlockAdditional` is passed to the track decoder in addition of the frame data. It is up to the decoder to use the "complement" data or not and how to use them. If the `AlphaMode` flag is set, that means the extra data define an alpha layer for the video, for example in WebM it's a secondary VP8/VP9 stream.
+
+Other BlockAddIDType values are defined and described by Block Additional Mappings. Block Additional Mappings detail how additional data MAY be stored in the BlockMore Element with a BlockAdditionMapping Element within the Track Element which identifies the BlockAdditional content. Block Additional Mappings define the reserved BlockAddIDType value reserved for that type of data as well as providing a label which MAY be stored within the BlockAddIDName Element. When the Block Additional Mapping is dependent on additional contextual information then the Mapping SHOULD describe that storage of how additional contextual information is stored within the BlockAddIDExtraData Element.
```
robUx4: Backtick on `BlockAddIDType` missing and many others

mjbshaw:
I can finally talk about [what I alluded to earlier](https://github.com/cellar-wg/matroska-specification/pull/287#issuecomment-440394625). YouTube now ingests and serves HDR10+ dynamic metadata. [`BlockAddID == 4` is used for VP9 to signal that the `BlockAdditional` payload is ITU T.35 metadata as defined by ITU-T T.35 terminal codes](https://www.webmproject.org/docs/container/#BlockAddID). I think it might make sense to have some kind of "registry" or something for `BlockAddIDType`. Without that I think various muxer implementations will diverge in the IDs they use for various types of metadata.

dericed:
Hi @robUx4, I did a little redesign of this PR so now there's a Block Additional Mapping introduction in its own document with an example in XML. In the Makefile this is add along with all markdown within the block_additional_mapping directory into the end of the Codec document. Please re-review.

dericed:
My recent updates don't accommodate @mjbshaw's comment at https://github.com/cellar-wg/matroska-specification/pull/287#issuecomment-534722873 as I think that would be better handled in a later PR as we'll have to discuss reserving BlockAddID=4.

ebml_matroska.xml
```diff
@@ -315,6 +315,21 @@
     <documentation lang="en" purpose="definition">The maximum value of <a href="https://www.matroska.org/technical/specs/index.html#BlockAddID">BlockAddID</a>. A value 0 means there is no <a href="https://www.matroska.org/technical/specs/index.html#BlockAdditions">BlockAdditions</a> for this track.</documentation>
     <extension webm="0"/>
   </element>
+  <element name="BlockAdditionMapping" path="0*(\Segment\Tracks\TrackEntry\BlockAdditionMapping)" id="0x41E4" type="master" minver="4" webm="0">
+    <documentation lang="en">Contains elements that describe each value of <a href="https://www.matroska.org/technical/specs/index.html#BlockAddID">BlockAddID</a> found in the Track.</documentation>
+  </element>
+  <element name="BlockAddIDValue" path="1*1(\Segment\Tracks\TrackEntry\BlockAdditionMapping\BlockAddIDValue)" id="0x41F0" type="uinteger" minOccurs="1" maxOccurs="1" minver="4" webm="0" default="1" range=">=2">
```
robUx4: If the range is `>=2`, the default value of  `1` is invalid.

block_additional_mappings/timecode.md
```diff
@@ -0,0 +1,19 @@
...
+
+### BlockAddIDName
+
+The BlockAddIDName value reserved for timecode is "timecode".
+
+### BlockAddIDExtraData
+
+#### Empty
+
+If the BlockAddIDExtraData Element is empty or unstored, the format of the timecode is a string expression comprised of only the follow ASCII characters `0`, `1`, `2`, `3`, `4`, `5`, `6`, `7`, `8`, `9`, `:`, `;`, and `.`.
```
robUx4: Is there any timecode specification that would tell how to interpret these numbers ?
Is `;:25645.:111` a valid timecode ?

ebml_matroska.xml
```diff
@@ -244,6 +244,26 @@
   <element name="MaxBlockAdditionID" path="1*1(\Segment\Tracks\TrackEntry\MaxBlockAdditionID)" id="0x55EE" type="uinteger" minOccurs="1" maxOccurs="1" minver="1" webm="0" default="0">
     <documentation lang="en">The maximum value of <a href="https://www.matroska.org/technical/specs/index.html#BlockAddID">BlockAddID</a>. A value 0 means there is no <a href="https://www.matroska.org/technical/specs/index.html#BlockAdditions">BlockAdditions</a> for this track.</documentation>
   </element>
+  <element name="BlockAdditionMapping" path="0*(\Segment\Tracks\TrackEntry\BlockAdditionMapping)" id="0x41E4" type="master" minver="4" webm="0">
+    <documentation lang="en">Contains elements that describe each value of <a href="https://www.matroska.org/technical/specs/index.html#BlockAddID">BlockAddID</a> found in the Track.</documentation>
+  </element>
+  <element name="BlockAddIDValue" path="1*1(\Segment\Tracks\TrackEntry\BlockAdditionMapping\BlockAddIDValue)" id="0x41F0" type="uinteger" minOccurs="1" maxOccurs="1" minver="4" webm="0" default="1" range="not 0">
```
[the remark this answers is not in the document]
robUx4: The value 1 for `BlockAddID` is valid and with a specific meaning. It should be allowed in `BlockAddIDValue` as it should cover all the values of `BlockAddID` (IMO). The fact this particular mapping is already defined is a side effect.

robUx4:
> I can finally talk about [what I alluded to earlier](https://github.com/cellar-wg/matroska-specification/pull/287#issuecomment-440394625). YouTube now ingests and serves HDR10+ dynamic metadata. [`BlockAddID == 4` is used for VP9 to signal that the `BlockAdditional` payload is ITU T.35 metadata as defined by ITU-T T.35 terminal codes](https://www.webmproject.org/docs/container/#BlockAddID). I think it might make sense to have some kind of "registry" or something for `BlockAddIDType`. Without that I think various muxer implementations will diverge in the IDs they use for various types of metadata.

Thanks for the tip @mjbshaw !
I think we should handle independant BlockAdditional types (ie not related as separate specs), just as it should be done for codecs (see the AV1 specs). At first we'll probably include a lot of common ones in the main codec specs. 

`BlockAddID` of 1 is already covered by these changes. For the value of 4 it should theoretically be combined with a `BlockAddIDValue` in the Track header. It seems unlikely YouTube will support that, since the spec is not even finalized yet. So if we want to be coherent we should treat value 4 as value 1, as a reserved/preexisting value. 1 is codec specific (ie data passed to the decoder with each frame), while 4 seems to be more metadata added to the frame.

`MaxBlockAdditionID` is not supported by WebM either, so we can't really rely on that to know in advance what `BlockAddID` will be found in the file or not.

On our side we should probably mention that BlockAddID value 4 is also a reserved value and explain what it does. I created #345 so we don't forget to do it once this PR is merged.

block_additional_mappings/timecode.md
dericed: I just meant this as an example and anticipated that we could use extra data to define various storage formats of timecode, but to start with one that doesn't necessarily imply rules about it. I'm thinking about how the decklink device input in ffmpeg will pass timecode as side data. The data is there as a string and there's certainly things that can be inferred about it but as side data, simply labelled as a timecode, there's no too much specific that could be stated about the storage of that data.

Perhaps we could provide a regex here though or cite SMPTE S12M, but in that case I'd suggest using the extra data to clarify that it is SMPTE S12M rather than letting the extra data be empty.

dericed:
I reviewed the recent changes and this seems ok to me. There is more work to do on timecode; however that isn't the focus on this PR. I suggest to merge and then we'll follow up with both timecode refinement and HDR.

ebml_matroska.xml
robUx4: @dericed ping, this is invalid IMO

block_additional_mappings/timecode.md
```diff
@@ -0,0 +1,19 @@
+## Block Additional Mapping - Timecode
+
+### Timecode Description
+
+Timecode values MAY be stored in the `BlockMore Element` to associate the content of a Matroska Block with a particular timecode value. The format of the timecode value is defined by which value is set in the BlockAddIDExtraData value.
+
+### BlockAddIDType
+
+The BlockAddIDType value reserved for timecode is 58.
```
robUx4: If the goal is to use SMPTE S12M in there, maybe use 12 as a hint ?

block_additional_mappings/timecode.md
robUx4: Just like a codec, you need different `BlockAddIDType` for each timecode type you want to store. You could have multiple types of timecodes set for the same frame. They are likely incompatible. So IMO this one should be SMPTE S12M only. If there are other formats, they define they own `Block Additional Mapping`.

Now looking at [Wikipedia](https://en.wikipedia.org/wiki/SMPTE_timecode) it seems that the norm is a 32 bits value, not a string. So there should at least be a variant with the untouched 32 bits value (maybe that's the value 12 to use for `BlockAddIDType`) and one with a string, which should have a normative definition (outside of the scope of Matroska ?).

dericed:
This PR is ready for a review. I removed the timecode block additional mapping (which was here as an example) to separate the discussions, but this focuses on the new elements for block additional mappings and how they are defined.

robUx4:
LGTM

dericed:
Merging then :)

PR #290, opened 2018-11-25
nit

retokromer:
- not sure `decomment` exists in English


robUx4:
Try to be more specific in the first line of commits, that's what most people will see in the git log.

retokromer:
OK

PR #292, opened 2018-11-25
all US English

retokromer:
- otherwise change `organizational`

### DIFF
--- a/tagging.md
+++ b/tagging.md
@@ -23,7 +23,7 @@ For authoring Tags outside of EBML, the [following XML syntax is proposed](http:
 
 ## Why official tags matter
 
-There is a debate between people who think all tags SHOULD be free and those who think all tags SHOULD be strict. If you look at this page you will realise we are in between.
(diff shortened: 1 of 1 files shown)


dericed:
There's some guidance in https://tools.ietf.org/html/rfc7322#section-3.1 in support of consistency. Perhaps worthwhile to query the working group for consensus before shifting the documentation to Americanized English as the majority of contributors and writings have been from non-American authors.

retokromer:
I personally prefer UK English, but in this case `organizational` is wrong.

JeromeMartinez:
The majority of contributors are also non-English ;-).
As section 3.1 slightly prioritize American English, I am in favor of American English.

retokromer:
That‘s exactly the reason why I changed to US English, despite the fact that I personally would prefer UK English.

robUx4:
Probably better to go with the newer version of English, so the US one.
There are many cases of the work `Initialisation` in the codec specs but in general the intent was to use US English.

retokromer:
OK, i’ll make a PR later on this.

PR #294, opened 2018-12-03
use US English

retokromer:
as discussed at https://github.com/Matroska-Org/matroska-specification/pull/292#issuecomment-443514478


robUx4:
LGTM thanks.

PR #297, opened 2019-01-02
Add media DB identifiers

Wampaa:
Given the difficulty for Media Center software to accurately identify a Movie or TV Show from its filename it is here suggested to provide a standardized way to link a Matroska file to those widely used (and relied upon) databases.

Note: A TV Show and a Movie can have the same ID number on TMDb. Using their API endpoint path as a way to distinguish them: `tv/` and `movie/` as a prefix to the value.

### DIFF
--- a/matroska_tags.xml
+++ b/matroska_tags.xml
@@ -290,6 +290,9 @@
     <tag name="LCCN" class="Identifiers" type="UTF-8">
       <description lang="en">[Library of Congress Control Number](https://www.loc.gov/marc/lccn.html)</description>
     </tag>
+    <tag name="IMDB" class="Identifiers" type="UTF-8">
+      <description lang="en">[Internet Movie Database (IMDb)](https://www.imdb.com/) identifier. 'tt' followed by at least 7 digits for Movies, TV Shows and Episodes.</description>
+    </tag>
     <tag name="PURCHASE_ITEM" class="Commercial" type="UTF-8">
       <description lang="en">URL to purchase this file. This is akin to the WPAY tag in ID3.</description>
     </tag>
--- a/matroska_tags.xml
+++ b/matroska_tags.xml
@@ -293,6 +293,9 @@
     <tag name="IMDB" class="Identifiers" type="UTF-8">
       <description lang="en">[Internet Movie Database (IMDb)](https://www.imdb.com/) identifier. 'tt' followed by at least 7 digits for Movies, TV Shows and Episodes.</description>
     </tag>
+    <tag name="TMDB" class="Identifiers" type="UTF-8">
+      <description lang="en">[The Movie Database (TMDb)](https://www.themoviedb.org/) identifier. The variable length digits string MUST be prefixed with either "movie/" or "tv/".</description>
+    </tag>
     <tag name="PURCHASE_ITEM" class="Commercial" type="UTF-8">
       <description lang="en">URL to purchase this file. This is akin to the WPAY tag in ID3.</description>
     </tag>
--- a/matroska_tags.xml
+++ b/matroska_tags.xml
@@ -296,6 +296,9 @@
     <tag name="TMDB" class="Identifiers" type="UTF-8">
       <description lang="en">[The Movie Database (TMDb)](https://www.themoviedb.org/) identifier. The variable length digits string MUST be prefixed with either "movie/" or "tv/".</description>
     </tag>
+    <tag name="TVDB" class="Identifiers" type="UTF-8">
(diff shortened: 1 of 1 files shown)


robUx4:
Good idea although it would be good to know the exact format to be used for each. It doesn't say if it's IDs or URLs. A link to a technical detail page describing each ID would be good as well.

Also you might use a commit for each (it may be 3 pull requests but that should be OK like that).

dericed:
With TMBD, what should a reader do if there's no prefix. If the id doesn't make sense without the prefix then use a "MUST" rather than a "should" (see https://www.ietf.org/rfc/rfc2119.txt).

In these examples if there a method to validate the ids or any external description that defines them. For instance with tv shows at TMDB I was a little confused as to how to parse the id from the url.

Wampaa:
The idea is to only include identifiers, not URLs that are prone to change.

I didn't find official pages formally describing those identifiers, so I described them.
Movie/TV Shows/Episode IMDb identifier starts with `tt`and is followed by at least 7 digits.
TMDb and TheTVDB identifiers are sequential numbers.

For IMDb they are to be found in the Movie/TV Show/Episode page's URL.
For TMDb we can find them on the URL path (only the digits) or with the API, and then add and the prefix.
For TheTVDB we can find them on the Shows page or with the API.

I split the commit and changed the should to MUST.

robUx4:
IMO it's still too vague. As the APIs suggest there can be more than one ID needed to describe an element. The URLs on the website may be one of them but I think the API descriptions are safer.

And to combine the IDs the way to do it in Matroska is to use multiple tags with a different `TargetTypeValue` (one for the episode, of for the show). It's less convenient to parse quickly something. But if episode IDs are unique globally they're still usable as such.
For IDs that don't target the default `TargetTypeValue` of [50 (Movie / Episode)](https://www.matroska.org/technical/specs/tagging/index.html#targettypes) it MUST be said what ID correspond to what. Probably with a link to the API.

As for the prefix, I don't know if it's worth adding. But if it is it's good to describe that it MUST be there of when it can be omitted. The API's don't seem to use it. I know for iTunes IDs on blu-ray.com they are omitted for example, even though they are in the URL.

You may also add an example for each to avoid confusion. Even if the URLs will not be usable in 10 years.

Wampaa:
For TMDb the ID is the same on the URL and in the API. The prefix is here because we can have this case:
https://tmdb.org/movie/100
https://tmdb.org/tv/100

Those two items both have and ID of 100, movie/ and tv/ is the only way to tell them apart.

The way it is proposed to use those tags for Movies is the following:
```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE Tags SYSTEM "matroskatags.dtd">
<Tags>
    <Tag> <!-- Movie -->
        <Targets>
            <TargetTypeValue>50</TargetTypeValue>
        </Targets>
        <Simple>
            <Name>IMDB</Name>
            <String>tt1234567</String>
        </Simple>
        <Simple>
            <Name>TMDB</Name>
            <String>movie/123456</String>
        </Simple>
    </Tag>
</Tags>
```

And for TV Shows we might come with something like this:
```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE Tags SYSTEM "matroskatags.dtd">
<Tags>
    <Tag> <!-- TV Show -->
        <Targets>
            <TargetTypeValue>70</TargetTypeValue>
        </Targets>
        <Simple>
            <Name>IMDB</Name>
            <String>tt1234567</String>
        </Simple>
        <Simple>
            <Name>TMDB</Name>
            <String>tv/123456</String>
        </Simple>
        <Simple>
            <Name>TVDB</Name>
            <String>123456</String>
        </Simple>
    </Tag>
  
    <Tag> <!-- Season -->
        <Targets>
            <TargetTypeValue>60</TargetTypeValue>
        </Targets>
        <Simple>
            <Name>PART_NUMBER</Name>
            <String>1</String> <!-- season number -->
        </Simple>
        <Simple>
            <Name>TOTAL_PARTS</Name>
            <String>10</String> <!-- number of episodes -->
        </Simple>
    </Tag>

    <Tag> <!-- Episode -->
        <Targets>
            <TargetTypeValue>50</TargetTypeValue>
        </Targets>
        <Simple>
            <Name>PART_NUMBER</Name>
            <String>10</String> <!-- episode number -->
        </Simple>
        <Simple>
            <Name>IMDB</Name>
            <String>tt1234568</String>
        </Simple>
    </Tag>
</Tags>
```

This way we can use the episode level identifier to retrieve information on IMDb, and use `PART_NUMBER`s to get them from TMDb and TheTVDB.

robUx4:
OK LGTM. It seems sufficient to define all the uses cases for each database.

PR #299, opened 2019-01-16
Correct ContentEncodingOrder description

mkver:
The current requirement of uniqueness among all ContentEncodingOrder elements of a Segment is not sensible and in contradiction to current practice.


dericed:
Noting that this change would cause a deviation with webm, see https://www.webmproject.org/docs/container/. So this PR should be correlated with a change in the webm specification.

mkver:
> 
> 
> Noting that this change would cause a deviation with webm, see https://www.webmproject.org/docs/container/. So this PR should be correlated with a change in the webm specification.

I agree that it would be nice to not create deviations with webm, but who's the webm guy we can talk to? And this problem is not unique to this PR; indeed, there are already lots of deviations with webm.

robUx4:
There's not really a person in charge of the WebM spec, nor a mailing list.

Anyway LGTM for this PR. It makes things clearer.

PR #302, opened 2019-01-22
Update the list wrt WebM-compability

mkver:
I noticed that there is quite a discrepancy between the [WebM specs](https://www.webmproject.org/docs/container/) and what Matroska says about WebM. This PR is my attempt to fix this. Notice that I read the absence of a "webm=" descriptor as "webm="1"" (i.e. elements are enabled in WebM by default). The website mostly seems to follow this logic, but for some elements (like `DefaultDecodedFieldDuration`) are declared as WebM incompliant although there is no "webm=" element in the spec table. Has this something to do with the fact that it is has a minimum version of 4?

The colour video element for which WebM support is preliminary has
been left out of this patch.


dericed:
Looks ok to me

PR #303, opened 2019-01-29
Revert "always have a ContentEncAlgo in there is a ContentEncryption"

mkver:
and do what was actually intended: Make `ContentEncAlgo` mandatory.

I have already explained this [here](https://github.com/Matroska-Org/matroska-specification/pull/281#issuecomment-455133872).


robUx4:
LGTM

PR #304, opened 2019-01-29
fix the ContentEncAlgo which may not be empty, unlike ContentEncryption

robUx4:
Fixes a bug introduced in #281


robUx4:
Yours is better.

mkver:
Ok, but could you nevertheless please explain what the numbers at the beginning of `path` are for?

robUx4:
The path is defined in EBML. Section 13.1.5.2 in https://datatracker.ietf.org/doc/draft-ietf-cellar-ebml/?include_text=1

More specifically this is the `EBMLElementOccurrence` part of the path with `EBMLMinOccurrence` and `EBMLMaxOccurrence`. They are duplicated as XML attributes for easier XML parsing.

PR #120, opened 2017-05-13
update to tag element definitions

dericed:

PR #308, opened 2019-02-11
Targets shouldn't be mandatory

mkver:
There is currently a mismatch in the definition of `Targets`: On the one hand, it is mandatory (minOccurs="1"), while on the other hand the description describes what happens when this element is not present. This conflict is resolved by keeping the description and changing the syntax, because this saves space for the not uncommon use-case in which the tag applies to the whole segment. Furthermore, changing it in this way doesn't invalidate any existing files, whereas changing the description to match the syntax would potentially invalidate files that abide by the current description and not the syntax.

The mismatch has been introduced with PR #120.

### DIFF
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -918,7 +918,7 @@
   <element name="Tag" path="1*(\Segment\Tags\Tag)" id="0x7373" type="master" minOccurs="1" minver="1" webm="1">
     <documentation lang="en">A single metadata descriptor.</documentation>
   </element>
-  <element name="Targets" path="1*1(\Segment\Tags\Tag\Targets)" cppname="TagTargets" id="0x63C0" type="master" minOccurs="1" maxOccurs="1" minver="1" webm="1">
+  <element name="Targets" path="0*1(\Segment\Tags\Tag\Targets)" cppname="TagTargets" id="0x63C0" type="master" maxOccurs="1" minver="1" webm="1">
     <documentation lang="en">Specifies which other elements the metadata represented by the Tag applies to. If empty or not present, then the Tag describes everything in the Segment.</documentation>
   </element>
   <element name="TargetTypeValue" path="0*1(\Segment\Tags\Tag\Targets\TargetTypeValue)" cppname="TagTargetTypeValue" id="0x68CA" type="uinteger" maxOccurs="1" minver="1" webm="1" default="50">


hubblec4:
Thanks for this PR, it is exactly what I have in my mind.
A real use case are the HAALI TRACKSETEX, there is no Targets element.

robUx4:
Actually I think it's the other way around. You should not have a tag without a target. The description `If empty or not present, then the Tag describes everything in the Segment` seems wrong to me as well (and I probably wrote it). For example if you join 2 Segments what happens to these tags ? Like taking many tracks of an album and making one file for the whole album. There's no way to know programatically.

IMO the `TargetTypeValue` should be mandatory and it has a default value. So by default the `Targets` don't have to be written anyway. And it also doesn't break any existing file. If the `Targets` was omitted, it meant it was for the whole Segment, which is likely for a movie or an episode or a track/album. Basically whatever the file contains. Because if it was for a part of the file it would have to be specified what part of the file.

What the example about joining tracks above shows, is that if `Targets` is mandatory the track tagging would have to be set explicitly (`TargetTypeValue`=30) (as it wouldn't be tags about the track if it was empty but about an album, the default is 50) rather than assuming it the Segment which happens to be a track. Noone would know.

So if anything it would force tags to be correct rather than hoping they are or guessing what they are.

It also shows setting ALBUM to 50 was a bad choice and it should have been TRACK as this level since that's what people usually have in one file.

robUx4:
As for Haali's tag. It's good they are documented but TrackSet seems like it should belong in Tracks and not in tags.

hubblec4:
Currently we have no mandatory element inside the `Targets element`.
A master element which is mandatory should have minimum one mandatory sub-element.

I'm not sure it is possible that a Matroska writer can write a master element without other elements.

HAALI TRACKSETEX
I know it is a workaround like the ROTATE feature. But again it shows how powerful Matroska Tags can be. A Matroska feature can be used/implemented without a "correct/right" place in the specs.

robUx4:
It's possible for an element to be mandatory and have no mandatory children. It can still be empty and present in the file or even missing but assumed to be present but with no children inside. In this case it makes no difference if it's written or not. The meaning in that case even specified in the text: it means the target is the whole segment, so all editions, all chapters and all tracks. This could be specified better. This is the same as having `TagTrackUID`, `TagEditionUID` and `TagChapterUID` written with no value, so their default value of 0 meaning all of "each of them".

robUx4:
The Tags are not meant to be used for playback. Only the Chapters are meant to alter playback. Otherwise the information must be in the Tracks. That's where rotate should be. I think that would be `ProjectionPoseYaw` that would hold this value for 2D videos.

hubblec4:
Hi Steve.
Sorry that I ask again, but could you describe here, why the Target element must be mandatory? (I didn't understand it right by our Matroska meeting).

hubblec4:
Like @dericed [wrote](https://github.com/cellar-wg/matroska-specification/issues/310#issuecomment-505944658), what is the issue if the `Targets` element isn't mandatory?

I guess in practice the `Targets` element is not written by a writer if there are no sub-elements. This means a reader can not trust that there is always a `Targets` element. This is for me the most important reason to say the `Targets` element should not be mandatory.

OK, there is a hint in the description that this element could not be present but it feels to me a bit complicated.

What should we do to finish this PR?
1. leave everything like it is
2. remove the mandatory flag
3. add more info to describe this edge case

I prefer point 2.

hubblec4:
a practice Matroska file
https://github.com/hubblec4/Matroska-Playback/blob/master/files/TRACKSETEX/TRACKSETEX.mkv

mcr:
If the sub-tag does not appear (the Target is empty), then the Target applies to the entire segment.
A master element of size zero is acceptable.
=> Targets element will be mandatory, but it may be zero sized, thus the target is the whole segment.

mcr:
The Target Element will have a default value of no-UIDs, so applies to the entire segment.
("But Master Elements do not have default values")

mcr:
use #334 to do this.
The Target remains mandatory.

hubblec4:
I have looked deeper into the practice Matroska file with TRACKSETEX.
The last Tag belongs to the TRACKSETEX and there is a Targets element with a TargetTypeValue element with a value of 2(I'm not sure, but 2 is not a valid value for TargetTypeValue).
It looks to me that mkvmerge.exe adds to an empty Targets element the TargetTypeValue element (in my mind is that not right/optimal).

An empty Targets element have to be written in this form [63][C0][80]. Is this correct?

robUx4:
@hubblec4 agrees to close this PR

PR #309, opened 2019-02-12
Fix typo

mkver:


robUx4:
How annoying that Github can spot such a simple difference and show it. I had to use my own merge tool.

LGTM

PR #311, opened 2019-02-26
rename the size element to length, as done in the EBML specs

robUx4:
Change required by https://github.com/Matroska-Org/ebml-specification/pull/217


robUx4:
This has been merged in EBML so we should merge it too.

PR #318, opened 2019-05-19
add a ChapterSkipType to mark chapters with a type of content to skip

robUx4:
It can be used to skip the content (credits) based on user preferences or show
a button to skip them during playback. It can also be used to start the next
content while the end credits are showing.

This could be done with tags but since it's a playback feature it's better if it's tied to the chapters. Just like `ChapterFlagEnabled`.

### DIFF
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -1405,7 +1405,7 @@ When disabled, the movie **SHOULD** skip all the content between the TimeStart a
 If a `ChapterAtom` is inside a `ChapterAtom` that has a `ChapterSkipType` set, it MUST NOT have a `ChapterSkipType` or have a `ChapterSkipType` with the same value as it's parent `ChapterAtom`.
 If the `ChapterAtom` doesn't contain a `ChapterTimeEnd`, the value of the `ChapterSkipType` is only valid until the next `ChapterAtom` with a `ChapterSkipType` value or the end of the file.
     </documentation>
-    <extension webm="0"/>
+    <extension type="webmproject.org" webm="0"/>
     <restriction>
       <enum value="0" label="No Skipping">
         <documentation lang="en" purpose="definition">Content which should not be skipped.</documentation>
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -1400,26 +1400,27 @@ When disabled, the movie **SHOULD** skip all the content between the TimeStart a
     <documentation lang="en" purpose="usage notes">The value **MUST NOT** be the `SegmentUID` value of the `Segment` it belongs to.</documentation>
     <implementation_note note_attribute="minOccurs">ChapterSegmentUID **MUST** be set (minOccurs=1) if ChapterSegmentEditionUID is used; see (#medium-linking) on medium-linking Segments.</implementation_note>
   </element>
-  <element name="ChapterContentType" path="0*1(\Segment\Chapters\EditionEntry\ChapterAtom\ChapterContentType)" id="0x4588" type="uinteger" maxOccurs="1" minver="4" webm="0" default="0">
+  <element name="ChapterContentType" path="0*1(\Segment\Chapters\EditionEntry\ChapterAtom\ChapterContentType)" id="0x4588" type="uinteger" maxOccurs="1" minver="4" default="0">
     <documentation lang="en">Indicate what type of content the ChapterAtom contains. It can be used to automatically skip content based on the type.</documentation>
-      <restriction>
+    <extension webm="0"/>
+    <restriction>
       <enum value="0" label="Regular Content"/>
       <enum value="1" label="Opening Credits">
         <documentation lang="en" purpose="definition">Credits usually found at the beginning of the content. The ChapterAtom MUST contain a value ChapterTimeEnd.</documentation>
       </enum>
-      <enum value="2" label="End Credits"/>
+      <enum value="2" label="End Credits">
         <documentation lang="en" purpose="definition">Credits usually found at the end of the content. The ChapterAtom MUST contain a value ChapterTimeEnd.</documentation>
       </enum>
-      <enum value="3" label="Recap"/>
+      <enum value="3" label="Recap">
         <documentation lang="en" purpose="definition">Recap of previous episodes of the content, usually found around the beginning. The ChapterAtom MUST contain a value ChapterTimeEnd.</documentation>
       </enum>
-      <enum value="4" label="Next Preview"/>
+      <enum value="4" label="Next Preview">
         <documentation lang="en" purpose="definition">Preview of the next episode of the content, usually found around the beginning. It may contain spoilers the user want to avoid. The ChapterAtom MUST contain a value ChapterTimeEnd.</documentation>
       </enum>
-      <enum value="5" label="Preview"/>
+      <enum value="5" label="Preview">
         <documentation lang="en" purpose="definition">Preview of the current episode of the content, usually found around the beginning. It may contain spoilers the user want to avoid. The ChapterAtom MUST contain a value ChapterTimeEnd.</documentation>
       </enum>
-      <enum value="6" label="Advertisement"/>
+      <enum value="6" label="Advertisement">
         <documentation lang="en" purpose="definition">Advertisement within the content, usually found around the beginning. The ChapterAtom MUST contain a value ChapterTimeEnd.</documentation>
       </enum>
     </restriction>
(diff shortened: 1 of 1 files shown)


robUx4:
This is similar to what Netflix and other OTT services do. It could be used for "local backups" of TV shows as well with proper player support.

hubblec4:
That's sound also interesting.
You said this could be done with Matroska Tags, do you have a working sample? Or could you describe how the user have to use the Tags?

You have specified 3 values, but a 4th value should be for a "Review", which I also skip(with ordered chapters).

hubblec4:
A question about the usage of the `ChapterContentType`. Should this element used in ordered chapters only?
For basic chapters there is no duration. A next basic chapter is required to have a play duration which can be skipped.

robUx4:
> You said this could be done with Matroska Tags, do you have a working sample? Or could you describe how the user have to use the Tags?

I'd rather remain as vague as possible as I don't think it should be done with tags (I didn't even thing very deep about this).

robUx4:
> You have specified 3 values, but a 4th value should be for a "Review", which I also skip(with ordered chapters).

If you mean skip back to the beginning of the chapter, then a skip button on the remote/player should handle that. You don't want a button on your screen all the time telling you you can go back. Which is really what this is about, a "popup" "button" to do simple special action.

If you mean giving the possibility to restart playback once an episode is finished, it might be done with the `End Credits` value. When you reach the end credit the player can give you the option to skip the credits, to restart from the beginning, or both. The values in the chapter don't force one way of handling this metadata.

robUx4:
> A question about the usage of the `ChapterContentType`. Should this element used in ordered chapters only?
> For basic chapters there is no duration. A next basic chapter is required to have a play duration which can be skipped.

Indeed, the chapters marked with `Start Credits` and `End Credits` MUST have a `ChapterTimeEnd` in order to be skipped.

It does not require to use ordered chapters.

hubblec4:
> I'd rather remain as vague as possible as I don't think it should be done with tags (I didn't even thing very deep about this).
Ah OK. I also think it makes not much sense to use Tags.

hubblec4:
>If you mean skip back to the beginning of the chapter, then a skip button on the remote/player should handle that. You don't want a button on your screen all the time telling you you can go back. Which is really what this is about, a "popup" "button" to do simple special action.

> If you mean giving the possibility to restart playback once an episode is finished, it might be done with the End Credits value. When you reach the end credit the player can give you the option to skip the credits, to restart from the beginning, or both. The values in the chapter don't force one way of handling this metadata.

No.
I mean with "Review": Last time on last episode. Some episodes starts with a review about the last episode(s).

The structure of such episodes is like this:
Review - FirstEpisodePart - Intro/Opening - SecondEpisodePart - Credits - Preview - StudioLogo

So we could add a 5th value for "Preview" content and a 6th value for "Logo".

hubblec4:
> Indeed, the chapters marked with Start Credits and End Credits MUST have a ChapterTimeEnd in order to be skipped.

> It does not require to use ordered chapters.

With ordered chapters it is clear. But with basic chapters? You need a next basic chapter with a higher start time. Or is it now possible to use the endtime of a basic chapter?

robUx4:
Ah, I think the word you're looking for is `Recap`. That's indeed a different kind of thing. In BattleStar Galactica there's also a `Preview` at the beginning of episodes, but right after the Start Credits IIRC.

One of the idea of the field is to skip the content. In the end the type itself doesn't matter too much. The difference between Start and End credits is mostly the location compared to the content. But usually after the End credits there's nothing else (I could the logos, translators, etc as end credits). The fact that the next episode Preview is in the credits might as well be skipped. But it's probably worth to differentiate with other kind of end content. Some people could set their player to never show the next episode spoilers but see the end credit and vice versa.

robUx4:
I'm not too sure about the `Logo`. I think it can be considered as an End Credit (Or Start Credit). If your player policy is to skip those kind it would make sense to skip the logo at the beginning or the end.

I think it's better to leave it open. That would makes things harder to configure for users or players (with more annoying popups).

hubblec4:
> Ah, I think the word you're looking for is Recap. 

In StarTrek episodes a speaker says: Last time on StarTrek.
Than 1min video plays as sum of the last episode. Now follows a short part of the episode, than the Opening part.
Recap is fine.

hubblec4:
OK, Start and End Credits are enough and could/should be used for Logo and other things as well.

robUx4:
I forgot the most annoying of all: advertisement.

dericed:
Maybe worth nesting the vocabulary. For instance so a user could say to skip all credits of any type (now or defined in the future), while our list is like "credits/opening", "credits/end", "credits/?middle?".

retokromer:
>I forgot the most annoying of all: advertisement.

You deserve a medal.

hubblec4:
I've been thinking about it for a while now.

The new function "Skip Video" should not be placed in a new element.
The whole thing is a chapter process for me and we already have the `ChapterProcess` element.

There one could set for example the `ChapProcessCodecID` to 0 (default is also 0).
That would give us the Native Matroska menu.
In `ChapProcessCommand->ChapProcessData` a new Matoska Script command would be used.
Then we would not have to worry about naming the video content.

It could then even information about the button be deposited as :. Position, view-duration, color, size, etc.

Furthermore, the question with the end times would be clarified,
because the `ChapProcess` element can only be used in ordered chapters (or should at least).

hubblec4:
What is with this feature? Should we add it or not?

I disagree with this feature like I wrote a ChapterProcess is better for this.

mcr:
Defer to an extension document.

robUx4:
For reference in the future document, this is not like ChapterProcessCommand which does one specific action. This is a Metadata that can be interpreted by the player as it wishes, for example for advertisement it could skip the content or just mute the audio (or play some other subversive content). This cannot be achieved with a hardcoded ChapterProcessCommand (or maybe a special chapter codec just to handle the state/kind of chapters being played.

The tagging of this information can also be used to remux content and remove the parts or encode with a different bitrate. That's information that's needed at the container level. What's inside of a codec (here a chapter codec) should not be needed at the container level.

hubblec4:
In principle, the idea is OK. It may help to know which chapter content is being used.

The only thing that is unclear is the use of basic chapters for it.
For this you had created an issue #322 in which you say that the end time should be used in basic chapters.
A basic chapter only uses a start time, according to the current Matroska specification.
This would then have to be changed and explained again.

hubblec4:
Can this PR reopen?

mcr:
Will need to be rebased.

mcr:
"Recap" -- "Preview", not sure whether or not the playback should continue after playing this part.
Should the player "keep watching" afterwards.

robUx4:
Rebased and updated.
Now we don't require ordered chapters and nested chapters are possible.

hubblec4:
> Now we don't require ordered chapters and nested chapters are possible.

LGTM.

Now there are a few enum values for that feature, but I guess there are more content which could be skipped.

- porn
- killing animals
- eating animals
- human puke

and for, me there are more content which should be skipped and I guess other humans want skip other content.
Is it maybe better to have a generic chapter skip content element? So we don't need to define all content types.

`ChapterSkipContent` element can be set to 1, means skip the content of the chapter. Default of this element is 0 -> don't skip.

robUx4:
This cannot be a `ChapterSkipContent` as it depends on what the user wants to skip or not. The best we can do is tag parts of content with specific content type. Then the user should setup it's player to select the kind of content s·he wants to skip. To do that there must be a fixed list known in advance (IMDB has some list of content people might not like).

The other option would be, when starting the file, to ask the user which content s·he wants to skip from the list found in the file. But as a convenience it's still better to have a convention on common values so that the user has the choice to make this selection in advance.

While credits/ads can be easily skipped in the content, as they don't blend in the rest of the content, it's more difficult to decide what part of a puke scene you'd want to skip. And if it's silently skipped then you may not understand what's going on (the player could display a message describing what actually happened, maybe keeping the sound as well).

robUx4:
Or maybe it *is* `ChapterSkipContent`. In the case of `ChapterContentType` you don't want to keep playing with only the sound on, you really want to skip the content.

In the kind of content you describe there may be more options for playback (even blurring the video). IMO it should be a different element because of this difference on how the player might handle the content. The player may opt to only support the element to "skip" the content and not the element to "hide" the content.

I'll edit this PR to reflect that.

hubblec4 (commented):
Ready to merge?

The new definition is fine to me and when we need more skip types we can easily add new values.

It is useful to set a default attribute with a default value of 0 (No skipping)?

robUx4:
I think the Next Preview should say it's usually found around the *end* of the content, not the beginning. Otherwise ready to merge.

We'll need an update to libmatroska to support it in VLC, poke @mbunkus

robUx4:
I'd like to merge this, but since it's a new feature it should be in Matroska v5. We would need to generate the RFC without the v5+ elements.
In general Pull Requests labeled "format_addition" should fall in this category.

robUx4:

PR #320, opened 2019-05-26
Normalize EBML schema

robUx4:
This should allow us to do some checks to validate the XML in Pull Request, have a unified way to check differences in EBML Schema.
This may also be put in the EBML repo without the Matroska specific elements.

It will also allow generating a clean/valid EBML Schema from the extended one we have now, generate one for WebM and one for DivX files. We can continue maintaing the one we have now, but to produce the RFC we should use a Schema without WebM or DivX.


transforms/normalize_ebml_schema.xsl
```diff
@@ -27,7 +27,7 @@
         <xsl:call-template name="parsePath">
             <xsl:with-param name="Path"><xsl:value-of select="@path" /></xsl:with-param>
         </xsl:call-template>
-        <xsl:if test="@minver">
+        <xsl:if test="@minver and minver!='1'">
```
robUx4: Should be @minver (and thus the produces files are wrong).

robUx4:
Closing as it's superseded by #325 which does the same thing and more and does it better.

PR #321, opened 2019-05-26
Use a valid EBML Schema to generate the Matroska RFC

robUx4:
And provide a way to generate a Schema for WebM


robUx4:
Later I'll try to unify these 2 XSLT and the one from #320 into a single one

robUx4:
Collides with #329 which has a cleaner way to generate the valid XML. It should be merged instead of this one for this feature.

The WebM handling not be what we want. So I'm closing this for now.

PR #323, opened 2019-06-01
the SeekID should not exceed 4 bytes, just like the max Matroska ID length

robUx4:
This check is done in libmatroska


dericed:
makes sense to me

PR #325, opened 2019-07-08
Sync ebml matroska with ebml schema definition

dericed:
This is an alternate pr to https://github.com/cellar-wg/matroska-specification/pull/320 and https://github.com/cellar-wg/matroska-specification/pull/321 but is dependent on a merge of https://github.com/cellar-wg/ebml-specification/pull/270.

td;lr it moves the cppname, divx and webm attributes from element to element/extension

### DIFF
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -22,48 +22,59 @@
   <element name="Info" path="1*(\Segment\Info)" id="0x1549A966" type="master" recurring="1" minOccurs="1" minver="1">
     <documentation lang="en" purpose="definition">Contains general information about the Segment.</documentation>
   </element>
-  <element name="SegmentUID" path="0*1(\Segment\Info\SegmentUID)" id="0x73A4" type="binary" maxOccurs="1" minver="1" webm="0" range="not 0" length="16">
+  <element name="SegmentUID" path="0*1(\Segment\Info\SegmentUID)" id="0x73A4" type="binary" maxOccurs="1" minver="1" range="not 0" length="16">
     <documentation lang="en" purpose="definition">A randomly generated unique ID to identify the Segment amongst many others (128 bits).</documentation>
     <documentation lang="en" purpose="usage notes">If the Segment is a part of a Linked Segment then this Element is REQUIRED.</documentation>
+    <extension webm="0"/>
   </element>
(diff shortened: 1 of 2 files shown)


robUx4:
Besides the dicussion about the way we can handle extra elements, I suppose you generate this using a tool (XSLT script ?) I would be good to put the script in `transform/` in the same PR so it can be reused, tweaked, validated.

ebml_matroska.xml
```diff
@@ -1,5 +1,5 @@
-<?xml version="1.0" encoding="utf-8"?>
-<EBMLSchema docType="matroska" version="4">
+ <?xml version="1.0" encoding="utf-8"?>
+<EBMLSchema docType="matroska" version="4" xmlns="https://ietf.org/cellar/ebml">
```
robUx4: Not sure it works for you but adding this makes the XML to MD/HTML not produce the right output (looking for a fix).
It is necessary for XSD validation so we should have it.

dericed: I'll fix that in the xsl

dericed: fixed and force pushed

dericed:
@robUx4 I did not use an xslt to generate this but careful editing.

ebml_matroska.xml
robUx4: arf, I saw this after I did #328. IMO it's cleaner to use `ebml:` rather than `e:`

robUx4:
> not use an xslt to generate this but careful editing

OK, I'll do a XSLT for that. I only trust machines 🤣 

IMO these should be different PRs. The type ->purpose patch can already be merged IMO. The namespace addition too (see #328).

The extra extension IMO needs more debate.

robUx4:
Rebased on top of #328 (`ebml:` instead of `e:`).
And I pushed `f6b04ba076aadab6c9bde9c20006caac4603049e` which is OK and unrelated to the `extension` change.

robUx4:
Added a XSLT script that transformed the old `ebml_matroska.xml` into the one with a single extension.
It will collide with #329

dericed:
LGTM, I suggest to merge soon. I'd like to start working on implementation_notes after this pr.

PR #328, opened 2019-07-13
Non empty namespace

robUx4:
Complement #325 and should probably be merged before


dericed:
Merging this and perhaps a few others before marking a release today.

PR #334, opened 2019-07-30
give an empty default to Targets

dericed:


dericed:
Updated to address conflict. Do we have consensus that ebml would need an update to address defaults with master elements or not blocking?

dericed:
Just noticed that the EBML spec presently says
> EBML Elements that are Master Elements MUST NOT declare a default value.

So this PR is invalid.

robUx4:
You should open an issue in the EBML specs to discuss if that's something we want or not (it sounds odd to me how to interpret this). Once this is solved we can get back to this.

mcr:
Need to apply solution in EBML to this Element.

dericed:
Closing as EBML will stay with the statement listed in https://github.com/cellar-wg/matroska-specification/pull/334#issuecomment-544293125

PR #335, opened 2019-08-01
Hard Linking provides a more option as was described.

hubblec4:
I have added some examples.

All this Hard Linkings working in VLC and MPC-HC and
in my chapterEditor is a tool which helps you to setup such Hard Linkings.

Hard-Linking example tables adjusted

### DIFF
--- a/notes.md
+++ b/notes.md
@@ -102,15 +102,39 @@ Matroska provides several methods to link two or many `Segment Elements` togethe
 
 ## Hard Linking
 
-Hard Linking (also called splitting) is the process of creating a `Linked Segment` by relating multiple `Segment Elements` using the `PrevUID` and `NextUID Elements`. Within a `Linked Segment`, the timestamps of each `Segment` MUST follow consecutively in linking order. With Hard Linking, the chapters of any `Segment` within the `Linked Segment` MUST only reference the current `Segment`. With Hard Linking, the `NextUID` and `PrevUID` MUST reference the respective `SegmentUID` values of the next and previous `Segments`. The first `Segment` of a `Linked Segment` MUST have a `NextUID Element` and MUST NOT have a `PrevUID Element`. The last `Segment` of a `Linked Segment` MUST have a `PrevUID Element` and MUST NOT have a `NextUID Element`. The middle `Segments` of a `Linked Segment` MUST have both a `NextUID Element` and a `PrevUID Element`.
+Hard Linking (also called splitting) is the process of creating a `Linked Segment` by relating multiple `Segment Elements` using the `PrevUID` and `NextUID Elements`. Within a `Linked Segment`, the timestamps of each `Segment` MUST follow consecutively in linking order. With Hard Linking, the chapters of any `Segment` within the `Linked Segment` MUST only reference the current `Segment`. With Hard Linking, the `NextUID` and `PrevUID` MUST reference the respective `SegmentUID` values of the next and previous `Segments`. The first `Segment` of a `Linked Segment` SHOULD have a `NextUID Element` and MUST NOT have a `PrevUID Element`. The last `Segment` of a `Linked Segment` SHOULD have a `PrevUID Element` and MUST NOT have a `NextUID Element`. The middle `Segments` of a `Linked Segment` SHOULD have both a `NextUID Element` and a `PrevUID Element`.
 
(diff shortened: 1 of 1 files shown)


notes.md
```diff
@@ -102,16 +102,40 @@ Matroska provides several methods to link two or many `Segment Elements` togethe
...
-As an example, four `Segments` can be Hard Linked as a `Linked Segment` through cross-referencing each other with `SegmentUID`, `PrevUID`, and `NextUID`, as in this table.
+As an example, three `Segments` can be Hard Linked as a `Linked Segment` through cross-referencing each other with `SegmentUID`, `PrevUID`, and `NextUID`, as in this table.
 
 file name   | `SegmentUID`                       | `PrevUID`                          | `NextUID`
 :-----------|:-----------------------------------|:-----------------------------------|:---------
 `start.mkv` | `71000c23cd31099853fbc94dd984a5dd` | n/a                                | `a77b3598941cb803eac0fcdafe44fac9`
 `middle.mkv`| `a77b3598941cb803eac0fcdafe44fac9` | `71000c23cd31099853fbc94dd984a5dd` | `6c92285fa6d3e827b198d120ea3ac674`
 `end.mkv`   | `6c92285fa6d3e827b198d120ea3ac674` | `a77b3598941cb803eac0fcdafe44fac9` | n/a
 
+An other example where the `NextUID Element` is used only.
```
robUx4: Should be
> where only the `NextUID Element` is used

notes.md
```diff
@@ -102,16 +102,40 @@ Matroska provides several methods to link two or many `Segment Elements` togethe
...
+
+file name   | `SegmentUID`                       | `PrevUID`                          | `NextUID`
+:-----------|:-----------------------------------|:-----------------------------------|:---------
+`start.mkv` | `71000c23cd31099853fbc94dd984a5dd` | n/a                                | `a77b3598941cb803eac0fcdafe44fac9`
+`middle.mkv`| `a77b3598941cb803eac0fcdafe44fac9` | n/a                                | `6c92285fa6d3e827b198d120ea3ac674`
+`end.mkv`   | `6c92285fa6d3e827b198d120ea3ac674` | n/a                                | n/a
+
+A next example where the `PrevUID Element` is used only.
```
robUx4: Should be
> where only the `PrevUID Element` is used

notes.md
```diff
@@ -102,16 +102,40 @@ Matroska provides several methods to link two or many `Segment Elements` togethe
...
+Hard Linking (also called splitting) is the process of creating a `Linked Segment` by relating multiple `Segment Elements` using the `PrevUID` and `NextUID Elements`. Within a `Linked Segment`, the timestamps of each `Segment` MUST follow consecutively in linking order. With Hard Linking, the chapters of any `Segment` within the `Linked Segment` MUST only reference the current `Segment`. With Hard Linking, the `NextUID` and `PrevUID` MUST reference the respective `SegmentUID` values of the next and previous `Segments`. The first `Segment` of a `Linked Segment` SHOULD have a `NextUID Element` and MUST NOT have a `PrevUID Element`. The last `Segment` of a `Linked Segment` SHOULD have a `PrevUID Element` and MUST NOT have a `NextUID Element`. The middle `Segments` of a `Linked Segment` SHOULD have both a `NextUID Element` and a `PrevUID Element`.
```
robUx4: OK with that. But we should also say that there has to be at least one link in either of the segments for this to work.

We should also mention what to do if both segments A has NextUID to segment B but segment B has a PrevUID not matching segment A. I think the `NextUID` should have priority over `PrevUID`. Or we can consider the link is broken and not use it at all.

hubblec4: Hard-Linking is a very simple construct, but there are a lot of linking options. I think it is enough to set the limits. So a `PrevUID` and/or `NextUID` must be there and then of course the file/segment. If these two conditions are not met, the linking ends at the current segment.

I have not really thought about a priority yet. But I think if one should set one, then for the `PrevUID` element. For the segment linking, one must first determine the "first" segment to build the timeline from there.

Your example of linking may be for a presentation, for example. Segment A is a long intro and segment B is the main part. If segment B has no `PrevUID`, it will be played alone. But if a `PrevUID` exists, to segment C, then this will be used. Segment C could be a short intro.

robUx4:
Replaced by #343

PR #336, opened 2019-08-01
Chapters changes

hubblec4:
Some more elements and other descriptions.

### DIFF
--- a/chapters.md
+++ b/chapters.md
@@ -3,31 +3,84 @@
 
 # Chapters
 
-## Edition and Chapter Flags
+The Matroska Chapters system can have multiple `Editions` and each `Edition` can consist of `Simple Chapters` where a chapter start time is used as marker in the timeline only. An `Edition` can be more complex with `Ordered Chapters` where a chapter end time stamp is additionally used or much more complex with `Linked Chapters`. The Matroska Chapters system can also have a menu structure, borrowed from the DVD menu system, or have it's own Native Matroska menu structure.
 
-### Chapter Flags
+## EditionEntry
 
-Two `Chapter Flags` are defined to describe the behavior of the `ChapterAtom Element`: `ChapterFlagHidden` and `ChapterFlagEnabled`.
+The `EditionEntry` is also called an `Edition`.  
+An `Edition` contains a set of `Edition Flags` and MUST contain at least one `ChapterAtom Element`.  
+Multiple `Editions` are allowed.
 
-If a `ChapterAtom Element` is the `Child Element` of another `ChapterAtom Element` with a `Chapter Flag` set to `true`, then the `Child ChapterAtom Element` MUST be interpreted as having its same `Chapter Flag` set to `true`. If a `ChapterAtom Element` is the `Child Element` of another `ChapterAtom Element` with a `Chapter Flag` set to `false` or if the `ChapterAtom Element` does not have a `ChapterAtom Element` as its `Parent Element`, then it MUST be interpreted according to its own `Chapter Flag`.
+### EditionUID
 
-As an example, consider a `Parent ChapterAtom Element` that has its `ChapterFlagHidden` set to `true` and also contains two child `ChapterAtoms`, the first with `ChapterFlagHidden` set to `true` and the second with `ChapterFlagHidden` either set to `false` or not present at all (in which case the default value of the Element applies, which is `false`). Since the parent `ChapterAtom` has its `ChapterFlagHidden` set to `true` then all of its children `ChapterAtoms` MUST also be interpreted as if their `ChapterFlagHidden` is also set to `true`. However, if a `Control Track` toggles the parent's `ChapterFlagHidden` flag to `false`, then only the parent `ChapterAtom` and its second child `ChapterAtom` MUST be interpreted as if `ChapterFlagHidden` is set to `false`. The first child `ChapterAtom` which has the `ChapterFlagHidden` flag set to `true` retains its value until its value is toggled to `false` by a `Control Track`.
-
-### Edition Flags
-
-Three `Edition Flags` are defined to describe the behavior of the `EditionEntry Element`: `EditionFlagHidden`, `EditionFlagDefault` and `EditionFlagOrdered`.
+A unique ID with a range from 1 to 18446744073709551615 can be used to identify the `Edition`. It's useful for tagging an `Edition`.
 
 
-#### EditionFlagHidden
+### EditionFlagHidden
 
-The `EditionFlagHidden Flag` behaves similarly to the `ChapterFlagHidden Flag`: if `EditionFlagHidden` is set to `true`, its `Child ChapterAtoms Elements` MUST also be interpreted as if their `ChapterFlagHidden` is also set to `true`, regardless of their own `ChapterFlagHidden Flags`. If `EditionFlagHidden` is toggled by a `Control Track` to `false`, the `ChapterFlagHidden Flags` of the `Child ChapterAtoms Elements` SHALL determine whether the `ChapterAtom` is hidden or not.
+When the `EditionFlagHidden Flag` is set to `false` means the `Edition` is visible and selectablly in a `Matroska Player`.
+All `ChapterAtoms Elements` MUST be interpreted of their own `ChapterFlagHidden Flags`.
+
+ChapterAtom / ChapterFlagHidden | False | True | visible
+--------------------------------|-------|------|--------
+Chapter 1                       |   X   |      | yes
+Chapter 2                       |       | X    | no
+
+When the `EditionFlagHidden Flag` is set to `true` the `Edition` is hidden and SHOULD not be selectablly in a `Matroska Player`.
+It exists one case where a hidden `Edition` MUST be played:  
+All `Editions` `EditionFlagHidden Flags` are set to true, so there is no visible `Edition`.  
+In this case all `ChapterAtoms Elements` MUST also be interpreted as if their `ChapterFlagHidden Flag` is also set to `true`, regardless of their own `ChapterFlagHidden Flags`.
+
+ChapterAtom / ChapterFlagHidden | False | True | visible
+--------------------------------|-------|------|--------
+Chapter 1                       |   X   |      | no
+Chapter 2                       |       | X    | no
+
+### EditionFlagDefault
+
+It is RECOMMENDED that no more than one `Edition` have an `EditionFlagDefault Flag` set to `true`. The first `Edition` with both the `EditionFlagDefault Flag` set to `true` and the `EditionFlagHidden Flag` set to `false` is the `Default Edition`.
+
+Edition   | FlagHidden | FlagDefault | Edition to play
+----------|------------|-------------|----------------
+Edition 1 | true       | true        |
+Edition 2 | true       | true        |
+Edition 3 | false      | true        | X
+
+If the `Default Edition's` `EditionFlagHidden Flag` is set to `true`, then a `Matroska Player` SHOULD play this `Edition` only if all other `Edition` `EditionFlagHidden Flags` are set to `true`. 
+
+Edition   | FlagHidden | FlagDefault | Edition to play
+----------|------------|-------------|----------------
+Edition 1 | true       | false       |
+Edition 2 | true       | true        | X
+Edition 3 | true       | false       |
+
+Exists an `Edition`  with `EditionFlagHidden Flag` set to `false`, the `Matroska Player` MUST play this `Edition`.
+
+Edition   | FlagHidden | FlagDefault | Edition to play
+----------|------------|-------------|----------------
+Edition 1 | true       | false       |
+Edition 2 | true       | true        |
+Edition 3 | false      | false       | X
 
 
-#### EditionFlagDefault
+If no `Default Edition` is specified a `Matroska Player` MUST play the first `Edition` with the `EditionFlagHidden Flag` is set to `false`. 
 
-It is RECOMMENDED that no more than one `Edition` have an `EditionFlagDefault Flag` set to `true`. The first `Edition` with both the `EditionFlagDefault Flag` set to `true` and the `EditionFlagHidden Flag` set to `false` is the `Default Edition`. When all `EditionFlagDefault Flags` are set to `false`, then the first `Edition` with the `EditionFlagHidden Flag` is set to `false`, MUST be played by the Matroska Player.
+Edition   | FlagHidden | FlagDefault | Edition to play
+----------|------------|-------------|----------------
+Edition 1 | true       | false       |
+Edition 2 | false      | false       | X
+Edition 3 | false      | false       |
 
-#### EditionFlagOrdered
+
+When all `EditionFlagHidden Flags` are set to `true`, then the first `Edition` MUST be played by the `Matroska Player`.
+
+Edition   | FlagHidden | FlagDefault | Edition to play
+----------|------------|-------------|----------------
+Edition 1 | true       | false       | X
+Edition 2 | true       | false       |
+Edition 3 | true       | false       |
+
+### EditionFlagOrdered
 
 The `EditionFlagOrdered Flag` is a significant feature as it enables an `Edition` of `Ordered Chapters` which defines and arranges a virtual timeline rather than simply labeling points within the timeline. For example, with `Editions` of `Ordered Chapters` a single `Matroska file` can present multiple edits of a film without duplicating content. Alternatively if a videotape is digitized in full, one `Ordered Edition` could present the full content (including colorbars, countdown, slate, a feature presentation, and black frames), while another `Edition` of `Ordered Chapters` can use `Chapters` that only mark the intended presentation with the colorbars and other ancillary visual information excluded. If an `Edition` of `Ordered Chapters` is enabled then the `Matroska Player` MUST play those Chapters in their stored order from the timestamp marked in the `ChapterTimeStart Element` to the timestamp marked in to `ChapterTimeEnd Element`.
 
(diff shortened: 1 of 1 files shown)


robUx4:
It seems all 3 commits are unrelated and should be different Pull Requests. It will be easier to discuss each commit individually (and merge one while the others are being discussed).

Also you should use a better description in your Pull Request. Saying it changes chapters doesn't give any idea of what it changes (in fact it doesn't change a lot, just add a lot of new text).

chapters.md
```diff
@@ -189,6 +189,28 @@ Chapter 2                | false              | no
 ++Nested Chapter 2.2.1   | true               | no
 ++Nested Chapter 2.2.2   | false              | no
 
+### ChapterSegmentUID
+
+
+The `ChapterSegmentUID` is a binary value with a size of 128 bits and the base element to set up a `Linked Chapter` in 2 variations. For both variations the following 3 conditions MUST be met.
```
robUx4: use `:` to tell the conditions follow

robUx4: Before listing the common conditions you may list the 2 variations first. The suspens to find out what they are is unbearable ;)

chapters.md
```diff
@@ -189,6 +189,28 @@ Chapter 2                | false              | no
...
+
+ 1. The `EditionFlagOrdered Flag` is set to true
```
robUx4: You could use anchors to the definition of the element: `#editionflagordered-element`

chapters.md
robUx4: You could use anchors to the definition of the element: `#chaptersegmentuid-element`

chapters.md
```diff
@@ -189,6 +189,28 @@ Chapter 2                | false              | no
...
+ 2. The `ChapterSegmentUID` is not the own `SegmentUID`
```
robUx4: IMO it sounds better as
> The `ChapterSegmentUID` MUST NOT be the `SegmentUID` of its own Segment

chapters.md
```diff
@@ -189,6 +189,28 @@ Chapter 2                | false              | no
...
+ 3. The linked Matroska file/Segment is in the same folder
```
robUx4: I think "folder" it too specific. You could have a media library that keeps track of segment UIDs and tell you where to find them. That should be explained in the segment linking section. And this list item should mention that the UID must be a Segment UID that matches the conditions for segment linking.

chapters.md
```diff
@@ -189,6 +189,28 @@ Chapter 2                | false              | no
...
+  
+#### Variation 1: Linked-Duration
+
```
robUx4: It's missing a paragraph explaining what is a `Linked-Duration`.

chapters.md
```diff
@@ -189,6 +189,28 @@ Chapter 2                | false              | no
...
+Two more conditions MUST be met.
+
+ 1. `ChapterTimeStart` and `ChapterTimeEnd` timestamps MUST be in the range of the linked Matroska file/Segment duration
```
robUx4: Do not use "file" here this is adding more confusion than it helps. We usually try to avoid it. Especially as linked segments often come concatenated in a single file.

chapters.md
```diff
@@ -189,6 +189,28 @@ Chapter 2                | false              | no
...
+ 2. `ChapterSegmentEditionUID` MUST be not set
+
+A `Matroska Player` MUST play the content of the linked Matroska file/Segment from the `ChapterTimeStart` until `ChapterTimeEnd` timestamp.
+
+#### Variation 2: Linked-Edition
+
+Only one more condition MUST be met when the `ChapterSegmentEditionUID` is set with a valid `EditionUID` from the linked Matroska file/Segment. A `Matroska Player` MUST play these linked `Edition`.
```
robUx4: `these` should be  `this`, there's only one edition UID in the ChapterSegmentEditionUID.

I think we can remove `Only one more condition MUST be met` and rephrase to use a single sentence like

> When the `ChapterSegmentEditionUID` is set with a valid `EditionUID` from the linked Segment, the `Matroska Player` MUST play this linked `Edition`.

robUx4: It should also say somewhere that the 2 variants are mutually exclusive.

chapters.md
```diff
@@ -3,31 +3,84 @@
...
-## Edition and Chapter Flags
+The Matroska Chapters system can have multiple `Editions` and each `Edition` can consist of `Simple Chapters` where a chapter start time is used as marker in the timeline only. An `Edition` can be more complex with `Ordered Chapters` where a chapter end time stamp is additionally used or much more complex with `Linked Chapters`. The Matroska Chapters system can also have a menu structure, borrowed from the DVD menu system, or have it's own Native Matroska menu structure.
 
-### Chapter Flags
+## EditionEntry
 
-Two `Chapter Flags` are defined to describe the behavior of the `ChapterAtom Element`: `ChapterFlagHidden` and `ChapterFlagEnabled`.
+The `EditionEntry` is also called an `Edition`.  
+An `Edition` contains a set of `Edition Flags` and MUST contain at least one `ChapterAtom Element`.  
+Multiple `Editions` are allowed.
```
robUx4: This is already said in the previous paragraph.

chapters.md
```diff
@@ -3,31 +3,84 @@
 
 # Chapters
 
-## Edition and Chapter Flags
+The Matroska Chapters system can have multiple `Editions` and each `Edition` can consist of `Simple Chapters` where a chapter start time is used as marker in the timeline only. An `Edition` can be more complex with `Ordered Chapters` where a chapter end time stamp is additionally used or much more complex with `Linked Chapters`. The Matroska Chapters system can also have a menu structure, borrowed from the DVD menu system, or have it's own Native Matroska menu structure.
```
robUx4: `Simple Chapters` is not name that has a proper definition in the document. No upper casing or backticking.

robUx4: The DVD menu system is one example of multiple possibilities. We should not mention in a generic paragraph. It's a chapter codec, and just like the other codecs, should/will be in another document.

chapters.md
```diff
@@ -3,31 +3,84 @@
...
+### EditionUID
 
-As an example, consider a `Parent ChapterAtom Element` that has its `ChapterFlagHidden` set to `true` and also contains two child `ChapterAtoms`, the first with `ChapterFlagHidden` set to `true` and the second with `ChapterFlagHidden` either set to `false` or not present at all (in which case the default value of the Element applies, which is `false`). Since the parent `ChapterAtom` has its `ChapterFlagHidden` set to `true` then all of its children `ChapterAtoms` MUST also be interpreted as if their `ChapterFlagHidden` is also set to `true`. However, if a `Control Track` toggles the parent's `ChapterFlagHidden` flag to `false`, then only the parent `ChapterAtom` and its second child `ChapterAtom` MUST be interpreted as if `ChapterFlagHidden` is set to `false`. The first child `ChapterAtom` which has the `ChapterFlagHidden` flag set to `true` retains its value until its value is toggled to `false` by a `Control Track`.
+A unique ID with a range from 1 to 18446744073709551615 can be used to identify the `Edition`. It's useful for tagging an `Edition`.
```
robUx4: We don't need to know the exact integer value or the maximum edition UID. In fact there's no restriction on the values apart that it must fit in a uinteger. If you want to restrict the number of bits, say it in bits. This value may be a max unsigned 32 or 64 bits but I have no idea.

chapters.md
```diff
@@ -3,31 +3,84 @@
...
 
-### Edition Flags
 
-Three `Edition Flags` are defined to describe the behavior of the `EditionEntry Element`: `EditionFlagHidden`, `EditionFlagDefault` and `EditionFlagOrdered`.
+### EditionFlagHidden
 
+When the `EditionFlagHidden Flag` is set to `false` means the `Edition` is visible and selectablly in a `Matroska Player`.
```
robUx4: *it* means

robUx4: selectablly > usable

chapters.md
```diff
@@ -3,31 +3,84 @@
...
+All `ChapterAtoms Elements` MUST be interpreted of their own `ChapterFlagHidden Flags`.
```
robUx4: I don't think this sentence is necessary. ChapterFlagHidden is a child of the ChapterAtom so it applies to only this ChapterAtom.

chapters.md
```diff
@@ -3,31 +3,84 @@
...
 
-#### EditionFlagHidden
+ChapterAtom / ChapterFlagHidden | False | True | visible
+--------------------------------|-------|------|--------
+Chapter 1                       |   X   |      | yes
+Chapter 2                       |       | X    | no
 
-The `EditionFlagHidden Flag` behaves similarly to the `ChapterFlagHidden Flag`: if `EditionFlagHidden` is set to `true`, its `Child ChapterAtoms Elements` MUST also be interpreted as if their `ChapterFlagHidden` is also set to `true`, regardless of their own `ChapterFlagHidden Flags`. If `EditionFlagHidden` is toggled by a `Control Track` to `false`, the `ChapterFlagHidden Flags` of the `Child ChapterAtoms Elements` SHALL determine whether the `ChapterAtom` is hidden or not.
+When the `EditionFlagHidden Flag` is set to `true` the `Edition` is hidden and SHOULD not be selectablly in a `Matroska Player`.
```
robUx4: SHOULD be hidden and not usable

The SHOULD applies to both criteria.

chapters.md
```diff
@@ -3,31 +3,84 @@
...
+It exists one case where a hidden `Edition` MUST be played:
```
robUx4: The English syntax is not correct. I think you mean

> There is on case where a hidden Edition MUST be played:

chapters.md
```diff
@@ -3,31 +3,84 @@
...
-#### EditionFlagHidden
+ChapterAtom / ChapterFlagHidden | False | True | visible
+--------------------------------|-------|------|--------
+Chapter 1                       |   X   |      | yes
+Chapter 2                       |       | X    | no
 
-The `EditionFlagHidden Flag` behaves similarly to the `ChapterFlagHidden Flag`: if `EditionFlagHidden` is set to `true`, its `Child ChapterAtoms Elements` MUST also be interpreted as if their `ChapterFlagHidden` is also set to `true`, regardless of their own `ChapterFlagHidden Flags`. If `EditionFlagHidden` is toggled by a `Control Track` to `false`, the `ChapterFlagHidden Flags` of the `Child ChapterAtoms Elements` SHALL determine whether the `ChapterAtom` is hidden or not.
+When the `EditionFlagHidden Flag` is set to `true` the `Edition` is hidden and SHOULD not be selectablly in a `Matroska Player`.
+It exists one case where a hidden `Edition` MUST be played:  
+All `Editions` `EditionFlagHidden Flags` are set to true, so there is no visible `Edition`.
```
robUx4: Should be:

> All EditionFlagHidden flags are set to true in all Editions, ...

chapters.md
```diff
@@ -3,31 +3,84 @@
...
+ChapterAtom / ChapterFlagHidden | False | True | visible
+--------------------------------|-------|------|--------
+Chapter 1                       |   X   |      | yes
+Chapter 2                       |       | X    | no
 
-The `EditionFlagHidden Flag` behaves similarly to the `ChapterFlagHidden Flag`: if `EditionFlagHidden` is set to `true`, its `Child ChapterAtoms Elements` MUST also be interpreted as if their `ChapterFlagHidden` is also set to `true`, regardless of their own `ChapterFlagHidden Flags`. If `EditionFlagHidden` is toggled by a `Control Track` to `false`, the `ChapterFlagHidden Flags` of the `Child ChapterAtoms Elements` SHALL determine whether the `ChapterAtom` is hidden or not.
+When the `EditionFlagHidden Flag` is set to `true` the `Edition` is hidden and SHOULD not be selectablly in a `Matroska Player`.
+It exists one case where a hidden `Edition` MUST be played:  
+All `Editions` `EditionFlagHidden Flags` are set to true, so there is no visible `Edition`.  
+In this case all `ChapterAtoms Elements` MUST also be interpreted as if their `ChapterFlagHidden Flag` is also set to `true`, regardless of their own `ChapterFlagHidden Flags`.
```
robUx4: It's missing the important part of telling which Edition to use. But IMO this is wrong. If they are all hidden then too bad.

As described here https://matroska.org/technical/specs/chapters/index.html a Control Track could toggle the hidden flag during playback. But the player should not use hidden chapters/editions on its own.

chapters.md
```diff
@@ -3,31 +3,84 @@
...
+
+If the `Default Edition's` `EditionFlagHidden Flag` is set to `true`, then a `Matroska Player` SHOULD play this `Edition` only if all other `Edition` `EditionFlagHidden Flags` are set to `true`. 
+
+Edition   | FlagHidden | FlagDefault | Edition to play
+----------|------------|-------------|----------------
+Edition 1 | true       | false       |
+Edition 2 | true       | true        | X
+Edition 3 | true       | false       |
+
+Exists an `Edition`  with `EditionFlagHidden Flag` set to `false`, the `Matroska Player` MUST play this `Edition`.
```
robUx4: Exists > If there is

chapters.md
```diff
@@ -60,13 +113,81 @@ Track/TrackTranslate               |   -   |  X
...
-##### Ordered-Edition and Matroska Segment-Linking
+#### Ordered-Edition and Matroska Segment-Linking
 
-- Hard Linking: `Ordered-Chapters` supersedes the `Hard Linking`.
+- Hard Linking: `Ordered Chapters` supersedes the `Hard Linking`.
 - Soft Linking: In this complex system `Ordered Chapters` are REQUIRED and a `Chapter CODEC` MUST interpret the `ChapProcess` of all chapters.
 - Medium Linking: `Ordered Chapters` are used in a normal way and can be combined with the `ChapterSegmentUID` element which establishes a link to another Matroska file/Segment.
 
-See [the section on the `Linked Segments`](#linked-segments)) for more information about `Hard Linking`, `Soft Linking` and `Medium Linking`.
+See [the section on the `Linked Segments`](notes.md/#linked-segments) for more information about `Hard Linking`, `Soft Linking` and `Medium Linking`.
```
robUx4: Are you sure this produces the right link in the produced HTML ?

chapters.md
```diff
@@ -60,13 +113,81 @@ Track/TrackTranslate               |   -   |  X
...
+
+## ChapterAtom
+The `ChapterAtom` is also called a `Chapter`.
+A `Chapter` element can be used recursively. Such a child `Chapter` is called `Nested Chapter`. 
+
+### ChapterUID
+
+The `ChapterUID` is a mandatory Matroska element with a unique ID with a range from 1 to 18446744073709551615.
```
robUx4: No need for a range here either.

chapters.md
```diff
@@ -60,13 +113,81 @@ Track/TrackTranslate               |   -   |  X
...
+
+### ChapterStringUID
+A unique string ID to identify the `Chapter`.  
+Use for [WebVTT cue identifier storage](http://dev.w3.org/html5/webvtt/#webvtt-cue-identifier).
```
robUx4: You may split the reference in two. The proper RFC citation of the WebVTT spec is `[@!W3C.CR-webvtt1-20190404]`

(we should also do it in the `ChapterStringUID` definition)

chapters.md
```diff
@@ -60,13 +113,81 @@ Track/TrackTranslate               |   -   |  X
...
+
+### ChapterTimeStart
+A not scaled timestamp of the start of `Chapter` with nanosecond accuracy.  
+For `Simple Chapters` are the start time stamps equal to the chapter markers in the timeline.
```
robUx4: I don't understand this sentence.

Also the the timeline has no "chapter marker".

chapters.md
```diff
@@ -60,13 +113,81 @@ Track/TrackTranslate               |   -   |  X
...
+
+### ChapterTimeEnd
+A timestamp of the end of `Chapter` with nanosecond accuracy.  The timestamp is excluded and also not scaled.
+The end timestamp is used when the `Edition EditionFlagOrdered Flag` is set to `true`.
+A `Matroska Player` have to calculate a duration for this `Chapter` with the difference of end timestamp and start timestamp.
```
robUx4: have > has

chapters.md
```diff
@@ -60,13 +113,81 @@ Track/TrackTranslate               |   -   |  X
...
+
+### ChapterTimeStart
+A not scaled timestamp of the start of `Chapter` with nanosecond accuracy.  
+For `Simple Chapters` are the start time stamps equal to the chapter markers in the timeline.
+
+### ChapterTimeEnd
+A timestamp of the end of `Chapter` with nanosecond accuracy.  The timestamp is excluded and also not scaled.
+The end timestamp is used when the `Edition EditionFlagOrdered Flag` is set to `true`.
+A `Matroska Player` have to calculate a duration for this `Chapter` with the difference of end timestamp and start timestamp.  
+The end timestamp MUST be greater than the start timestamp otherwise the duration would be negative which is illegal. If the duration of a `Chapter` is 0, this `Chapter` will be ignored by a `Matroska Player`.
```
robUx4: No a 0 duration is not ignored. It may still have chapter codec commands to process (see `ChapProcess`).

chapters.md
```diff
@@ -60,13 +113,81 @@ Track/TrackTranslate               |   -   |  X
...
+Chapter 1                | false             | yes
++Nested Chapter 1.1      | false             | yes
++Nested Chapter 1.2      | true              | no
+Chapter 2                | true              | no
++Nested Chapter 2.1      | false             | yes
++Nested Chapter 2.2      | true              | no
+
+### ChapterFlagEnabled
+
+If the `ChapterFlagEnabled Flag` is set to `true` a `Matroska Player` MUST use this `Chapter`.
```
robUx4: Using chapters is not required for players. Web browsers don't use them and the files still play.

chapters.md
```diff
@@ -25,7 +25,7 @@ The `EditionFlagHidden Flag` behaves similarly to the `ChapterFlagHidden Flag`:
 
 #### EditionFlagDefault
 
-It is RECOMMENDED that no more than one `Edition` have an `EditionFlagDefault Flag` set to `true`. The first `Edition` with both the `EditionFlagDefault Flag` set to `true` and the `EditionFlagHidden Flag` set to `false` is the `Default Edition`. When all `EditionFlagDefault Flags` are set to `false`, then the first `Edition` is the `Default Edition`.
+It is RECOMMENDED that no more than one `Edition` have an `EditionFlagDefault Flag` set to `true`. The first `Edition` with both the `EditionFlagDefault Flag` set to `true` and the `EditionFlagHidden Flag` set to `false` is the `Default Edition`. When all `EditionFlagDefault Flags` are set to `false`, then the first `Edition` with the `EditionFlagHidden Flag` is set to `false`, MUST be played by the Matroska Player.
```
robUx4: `is` should be removed

robUx4: Why remove the concept of Default Edition ? You can just keep it and say that the player SHOULD play the Default Edition by default. But again, that's only if the player decides to handle chapters. It cannot be a MUST.

robUx4:
Closing as @hubblec4 cannot edit on top. Replaced by #340, #341 and #342.

Issue #315, opened 2019-05-07
ChapterTrackNumber


hubblec4:
The specs say you have to use a track UID. Should we rename the `ChapterTrackNumber` element to `ChapterTrackUID`?

dericed:
I agree this is confused, since there is a TrackNumber and TrackUID which are both uints and here the elements named ChapterTrackNumber is supposed to reference the "UID of the Track".

So if there's a Track with TrackNumber=1 and TrackUID=2 and the Chapter only applies to that Track, I'd guess that the ChapterTrackNumber should be set to 2, but is that right?

hubblec4:
> So if there's a Track with TrackNumber=1 and TrackUID=2 and the Chapter only applies to that Track, I'd guess that the ChapterTrackNumber should be set to 2, but is that right?

Yes value 2 is correct. The Matroska specs say it is a UID.

robUx4:
Fixed by #339

PR #339, opened 2019-09-02
Change ChapterTrackNumber to ChapterTrackUID

hubblec4:
issue #315


robUx4:
You need to add the `cppname` so that the generated code from the specs is still the same:

> <extension cppname="ChapterTrackNumber"/>

robUx4:
Merged with the required change on top (Github still allowed to merge even though the commit was already merged...).

PR #342, opened 2019-09-08
add more chapter flag explanations

robUx4:
Replaces https://github.com/cellar-wg/matroska-specification/pull/336/commits/c3618acabe49c515e782164a154b9df1bebe8456

I did not do any editing for now according to my comments.

### DIFF
--- a/chapters.md
+++ b/chapters.md
@@ -144,15 +144,6 @@ with the `ChapterSegmentUID` element which establishes a link to another Segment
 See (#linked-segments) on the Linked Segments for more information
 about `Hard Linking`, `Soft Linking`, and `Medium Linking`.
 
-- Hard Linking: `Ordered Chapters` supersedes the `Hard Linking`.
-- Soft Linking: In this complex system `Ordered Chapters` are REQUIRED and a `Chapter CODEC`
-  **MUST** interpret the `ChapProcess` of all chapters.
-- Medium Linking: `Ordered Chapters` are used in a normal way and can be combined with the
- `ChapterSegmentUID` element which establishes a link to another Matroska file/Segment.
-
-See [the section on the Linked Segments](#linked-segments) for more information about
-`Hard Linking`, `Soft Linking` and `Medium Linking`.
-
 ## ChapterAtom
 The `ChapterAtom` is also called a `Chapter`.
 A `Chapter` element can be used recursively. Such a child `Chapter` is called `Nested Chapter`.
--- a/chapters.md
+++ b/chapters.md
@@ -144,6 +144,32 @@ with the `ChapterSegmentUID` element which establishes a link to another Segment
 See (#linked-segments) on the Linked Segments for more information
 about `Hard Linking`, `Soft Linking`, and `Medium Linking`.
 
+### ChapterSegmentUID
+
+The `ChapterSegmentUID` is a binary value and the base element to set up a
+`Linked Chapter` in 2 variations: the Linked-Duration linking and the Linked-Edition
+linking. For both variations, the following 3 conditions **MUST** be met:
+
+ 1. The `EditionFlagOrdered Flag` **MUST** be true.
+ 2. The `ChapterSegmentUID` **MUST NOT** be the `SegmentUID` of its own `Segment`.
+ 3. The linked Segments **MUST** BE in the same folder.
+
+#### Variation 1: Linked-Duration
+
+Two more conditions **MUST** be met:
+
+ 1. `ChapterTimeStart` and `ChapterTimeEnd` timestamps **MUST** be in the range of the
+    linked Segment duration.
+ 2. `ChapterSegmentEditionUID` **MUST** be not set.
+
+A `Matroska Player` **MUST** play the content of the linked Segment from the
+`ChapterTimeStart` until `ChapterTimeEnd` timestamp.
+
+#### Variation 2: Linked-Edition
+
+When the `ChapterSegmentEditionUID` is set to a valid `EditionUID` from the linked
+Segment. A `Matroska Player` **MUST** play these linked `Edition`.
+
 ## ChapterAtom
 The `ChapterAtom` is also called a `Chapter`.
 A `Chapter` element can be used recursively. Such a child `Chapter` is called `Nested Chapter`.
--- a/chapters.md
+++ b/chapters.md
@@ -4,75 +4,111 @@ title: Chapters
 
 # Chapters
 
-## Edition and Chapter Flags
+The Matroska Chapters system can have multiple `Editions` and each `Edition` can consist of
+`Simple Chapters` where a chapter start time is used as marker in the timeline only. An
+`Edition` can be more complex with `Ordered Chapters` where a chapter end time stamp is additionally
+used or much more complex with `Linked Chapters`. The Matroska Chapters system can also have a menu
+structure, borrowed from the DVD menu system, or have it's own Native Matroska menu structure.
 
-### Chapter Flags
+## EditionEntry
 
-Two `Chapter Flags` are defined to describe the behavior of the `ChapterAtom Element`:
-`ChapterFlagHidden` and `ChapterFlagEnabled`.
+The `EditionEntry` is also called an `Edition`.
+An `Edition` contains a set of `Edition` flags and **MUST** contain at least one `ChapterAtom Element`.
+Chapters are always inside an `Edition` (or a Chapter itself part of an `Edition`).
+Multiple Editions are allowed. Some of these Editions **MAY** be ordered and others not.
 
-If a `ChapterAtom Element` is the `Child Element` of another `ChapterAtom Element`
-with a `Chapter Flag` set to `true`, then the `Child ChapterAtom Element` **MUST** be
-interpreted as having its same `Chapter Flag` set to `true`. If a `ChapterAtom Element`
-is the `Child Element` of another `ChapterAtom Element` with a `Chapter Flag` set to `false`,
-or if the `ChapterAtom Element` does not have a `ChapterAtom Element` as its `Parent Element`,
-then it **MUST** be interpreted according to its own `Chapter Flag`.
+### EditionFlagHidden
 
-As an example, consider a `Parent ChapterAtom Element` that has its `ChapterFlagHidden`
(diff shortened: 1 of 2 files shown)


hubblec4:
OK.

hubblec4:
What is the plan now with this changes?

robUx4:
Rebased on master

chapters.md
```diff
@@ -3,31 +3,84 @@
 
 # Chapters
 
-## Edition and Chapter Flags
+The Matroska Chapters system can have multiple `Editions` and each `Edition` can consist of `Simple Chapters` where a chapter start time is used as marker in the timeline only. An `Edition` can be more complex with `Ordered Chapters` where a chapter end time stamp is additionally used or much more complex with `Linked Chapters`. The Matroska Chapters system can also have a menu structure, borrowed from the DVD menu system, or have it's own Native Matroska menu structure.
```
hubblec4: The term `Simple Chapters` is simple not well, it should be `Basic Chapters`.
https://github.com/hubblec4/Matroska-Playback/blob/master/src/BasicChapters.md

Is it a good idea to describe the chapter types in front of this document?
https://github.com/hubblec4/Matroska-Playback/blob/master/src/NestedChapters.md
https://github.com/hubblec4/Matroska-Playback/blob/master/src/OrderedChapters.md
https://github.com/hubblec4/Matroska-Playback/blob/master/src/NestedOrderedChapters.md

chapters.md
```diff
@@ -3,31 +3,84 @@
...
 
-### Chapter Flags
+## EditionEntry
```
hubblec4: Here is a better description of the edition elements with samples.
https://github.com/hubblec4/Matroska-Playback/blob/master/src/EditionEntry.md

chapters.md
robUx4: I would prefer changing the term `Simple Chapters` in another PR. It's currently found in `EditionFlagOrdered`. You can open an issue for that.

robUx4: IMO Nested chapters are not a special kind of chapters. So they shouldn't be mentioned in the same paragraph which described different ways the chapters can be interpreted.

chapters.md
robUx4: I picked some stuff from the generale description. The rest don't seem to apply easily to this PR.

robUx4:
Updated some more text. Some of it was contradicting the text in `ebml_matroska.xml`.
@hubblec4 you should have a look to see if it's OK with you.

chapters.md
hubblec4: This is OK.

hubblec4:
> [rework the edition flags section](https://github.com/cellar-wg/matroska-specification/pull/342/commits/e105e32bd9a0ffbed6264c149d6e5152c8c97c1e)

LGTM.

But with this configuration is it now impossible to to play a hidden `Edition` by default.
I use for my menu.mkv's a hidden default `Edition`, which can be the first `Edition` or the last one.

This `Edition` is used to play an intro of a disc or trailer of a series.
Now I can select an edition in the edition menu to play them.
And of course the default edition is not visible and can't be selected. This means you can't play the intro/trailer edition, which is also intended that way.

I just don't understand why the `EditionFlagHidden` has to be considered.

OK, you could say, then set the default edition HiddenFlag to false
but in this case all chapters are visible to the user.
I use a lot of ordered chapters for such menu.mkv's and now I have to hide them all manually.

The simplest and best solution is to say:
The `Default Edition` is the first `Edition` with `EditionFlagDefault` flag set to true.
When all `EditionFlagDefault` flags are set to false then the first `Edition` with the `EditionFlagHidden` flag set to false is the `Default Edition`.
When all `EditionFlagDefault` flags are set to false  and all `EditionFlagHidden` set to true then the first `Edition` is the `Default Edition`.

hubblec4:
> [rewrite the ChapterTimeEnd section](https://github.com/cellar-wg/matroska-specification/pull/342/commits/8898a08317032bcffaa98b651c43188303049b4a)

> If the duration of a `Chapter` is 0, this `Chapter` MUST be ignored.

Yeah this at the moment the typical practice, but such a chapter could use a chapter-process with a command.
For example a last ordered chapter with a zero duration is used to jump back to a main menu edition.

hubblec4:
> [remove sentence I don't understand](https://github.com/cellar-wg/matroska-specification/pull/342/commits/98890c4325c200ecf07aaff10d80f3eba8f155b7)

When I played with DVDMenuXtractor(DMX)'s output, I saw that some end timestamps was empty or set to 00:00:00.000000000 in an ordered chapter, which is illegal to the current specs.

And the chapters uses a process and commands, and so I thought the end time is sometimes ignored.

hubblec4:
> [change the ChapterFlagHidden paragraph](https://github.com/cellar-wg/matroska-specification/pull/342/commits/cba050ce6476542f734b61d4f7caadf49511d3c1)

I know what the current specs say: A nested chapters is automatically always hidden when the parent chapter's `ChapterFlagHidden` flag is set to true.
And I can also understand the logic in this behaviour, it is like the `ChapterFlagEnabled` flag works.

But all tested players and Splitters shows the nested chapters.
So VLC or MPC-HC and other players are all buggy??
I say no, in this case the specs are wrong.

Personally I need this behaviour for my menu.mkv's because all episodes can now presented in level 2 which uses a "+" in front of the chapter name. It is now easy to see which chapters belong to an episode.
The top level chapters are used to link to other files in the folder and here I don't want to see these chapters in the timeline. Only the first level 1 chapter of an episode is visible to signal here starts a new episode.

robUx4:
You're right. DvdMenuXtractor uses the hidden chapter flag to hide the various DVD data levels from the user and only showing the ones that are meaningful. So nested chapters inside a hidden chapter MAY be visible. This is a good feature we should keep.

hubblec4:
> You're right. DvdMenuXtractor uses the hidden chapter flag to hide the various DVD data levels from the user and only showing the ones that are meaningful. So nested chapters inside a hidden chapter MAY be visible. This is a good feature we should keep.

Cool, that's nice to read.

We could add a new chapter flag to hide all nested chapters if this needed.
Maybe `ChapterFlagFullHidden`....

robUx4:
> > [remove sentence I don't understand](https://github.com/cellar-wg/matroska-specification/pull/342/commits/98890c4325c200ecf07aaff10d80f3eba8f155b7)
> 
> When I played with DVDMenuXtractor(DMX)'s output, I saw that some end timestamps was empty or set to 00:00:00.000000000 in an ordered chapter(start time greater than 0), which is illegal to the current specs.
> 
> And the chapters uses a process and commands, and so I thought the end time is sometimes ignored.

Yeah this is debatable. Since we are in ordered chapters the process command could be merged in the previous and next chapters and avoid having a zero duration chapter to play. It's tricky for players (I know for VLC) to consider playback of 0 duration. If no data is sent to the player it will not see the chapter passing. 

I think it's better to fix DvdMenuXtractor to avoid this case rather than making life harder for players. It may exist in DVDs but that because the playback of cells is driven by the IFO file and not the VOB files. That's not how Matroska works.

Merging of DVD commands may not be possible directly, but it's always possible to use nested chapter to add the before/after commands. It may even be possible to process the XML output of DMX with XLST without touching the code of DMX.

hubblec4:
> Yeah this is debatable. Since we are in ordered chapters the process command could be merged in the previous and next chapters and avoid having a zero duration chapter to play. It's tricky for players (I know for VLC) to consider playback of 0 duration. If no data is sent to the player it will not see the chapter passing.

Yes, I think a 0 duration chapter makes not really sense. it is easier to say such chapters are illegal!

robUx4:
> > [rework the edition flags section](https://github.com/cellar-wg/matroska-specification/pull/342/commits/e105e32bd9a0ffbed6264c149d6e5152c8c97c1e)
> 
> LGTM.
> 
> But with this configuration is it now impossible to play a hidden `Edition` by default.
> I use for my menu.mkv's a hidden default `Edition`, which can be the first `Edition` or the last one.
> 
> This `Edition` is used to play an intro of a disc or trailer of a series.
> Now I can select an edition in the edition menu to play them.
> And of course the default edition is not visible and can't be selected. This means you can't play the intro/trailer edition, which is also intended that way.
> 
> I just don't understand why the `EditionFlagHidden` has to be considered.
> 
> OK, you could say, then set the default edition HiddenFlag to false
> but in this case all chapters are visible to the user.
> I use a lot of ordered chapters for such menu.mkv's and now I have to hide them all manually.
> 
> The simplest and best solution is to say:
> The `Default Edition` is the first `Edition` with `EditionFlagDefault` flag set to true.
> When all `EditionFlagDefault` flags are set to false then the first `Edition` with the `EditionFlagHidden` flag set to false is the `Default Edition`.
> When all `EditionFlagDefault` flags are set to false and all `EditionFlagHidden` set to true then the first `Edition` is the `Default Edition`.

The purpose of hidden edition is to hide content from the user. It can only be enabled via Control Tracks. But so far there is no way to enable them (no chapter command can do it).

The feature you want to to have a visible edition as the default, but with all chapters hidden. You can then play content when you start but with no way to select chapters.

hubblec4:
> The purpose of hidden edition is to hide content from the user. It can only be enabled via Control Tracks. But so far there is no way to enable them (no chapter command can do it).

Yeah, a hidden edition is not selectable for a user, but can be played/used by a player.
In practice, if we have only one edition which is hidden, all the ordered chapters are used in a "normal way".
All the chapter markers are shown in the timeline -> current state of the players.

A hidden edition could also be played by a Control Track or chapter command.
I know there is no player which support this and there is also no chapter command defined, but we could create such commands.

hubblec4 (commented):
You should merge this PR and then we can discuss about smaller sections.
Mostly of this changes are fine, but the `ChapterFlagHidden` description should be changed.

hubblec4:
can this PR also merged?
The ChapterFlagHidden section looks fine now.

robUx4:
Rebased.
It will need a good amount of review since a lot of normative text is edited.

hubblec4:
For my first reading, it seems to be fine.

retokromer (approved):

JeromeMartinez (approved):

PR #349, opened 2019-10-23
Update to xml2rfc version 3 processing

dericed:
This is to apply the same approach as in the ebml repository and use xml2rfc version 3. For me, this builds though there may be more to do.


dericed:
I added more commits to address issues from the rfc tools at https://tools.ietf.org/tools/idnits/. Please review.

dericed:
bump on this

dericed:
works for me with your additions, so merging. Will push a new version soon.

PR #350, opened 2019-10-27
fix typo in make process for codecs doc

dericed:
Not sure where this regression occurred, but somewhere the name of the codecs output from changed from `codec` to `codecs`. The IETF data tracker thus sees updates as discontinuous, so I set it back.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,10 +1,10 @@
 _site
 .sass-cache
 *lhomme-cellar-matroska-*
-*lhomme-cellar-codecs-*
+*lhomme-cellar-codec*
 *lhomme-cellar-tags-*
 *ietf-cellar-matroska-*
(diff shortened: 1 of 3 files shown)


dericed:
I agree, that `codecs`  would be better, but I think we'd have to somehow correct that in the datatracker first. I updated .gitignore and made it more inclusive.

robUx4:
> 
> 
> I agree, that `codecs` would be better, but I think we'd have to somehow correct that in the datatracker first. I updated .gitignore and made it more inclusive.

Did you forget to push ?

dericed:
My update is there. In .gitignore I used ...codec* to match either ...codec... or ...codecs...

robUx4:
Oh, OK. LGTM then.

PR #352, opened 2019-10-27
Ci/1

robUx4:
The 3 documents are generated in one task

The `ebml_matroska.xml` Schema is validated in another task, using `EBMLSchema.xsd` from the EBML repository.

Examples of the output can be found at https://github.com/robUx4/matroska-specification/actions

We need to enable the Github Actions in order to use this.

### DIFF
--- a/.github/workflows/generate-rfcs.yaml
+++ b/.github/workflows/generate-rfcs.yaml
@@ -42,7 +42,7 @@ jobs:
         run: |
           mkdir -p artifacts
           mv draft-ietf-cellar-matroska-??.xml artifacts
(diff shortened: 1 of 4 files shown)


robUx4:
The Makefiles are safe. Also as long as we don't sign up for the Github Actions beta for `wg-cellar` the actions not be used. When it's unable it should tell if a PR is valid or not, or at least if the master is valid or not. (at least that's my goal) Possibly blocking invalid stuff until they are fixed (which is safer and faster to review if something is invalid in the first place).

robUx4:
Merging as it has no effect until we enable Github Actions

PR #353, opened 2019-10-27
Add block additional mapping for generic type/value

dericed:
This is based upon the timecode branch, but this offers a generic key/value structure for side data. For discussion.


block_additional_mappings_intro.md
```diff
@@ -48,3 +48,12 @@ Block Additional Mappings define the `BlockAddIDType` value reserved to identify
...
 
 The following Block Additional Mappings are defined.
+
+## Summary of Assigned BlockAddIDType Values
+
+For convenience, the following table shows the assigned BlockAddIDType values along with the BlockAddIDName and Citation.
+
+| BlockAddIDType | BlockAddIDName                                               | Citation                             |
+|:---------------|:-------------------------------------------------------------|:-------------------------------------|
+| 12             | SMPTE ST 12-1 timecode                                       | (#smpte-st12-1-timecode)             |
```
robUx4: Should be 121 according to #348.

If the goal is to add just the Key Value Pair after SMPTE 12-1, rebase this branch on top of  `add-timecode-block-additional-mapping`.

dericed: Yes, this pr is built upon https://github.com/cellar-wg/matroska-specification/pull/348 (which I suggest to resolve first). I rebased so now it matches 348.

block_additional_mappings/key_value.md
```diff
@@ -0,0 +1,17 @@
+## Key Value Pair
+
+### Key Value Pair Description
+
+The Block Additional Mapping for a `Key Value Pair` provides a generic method store data with a label in association with a Block. The label (referred to here as the `Key`) is stored within the `BlockAddIDExtraData` Element and the associate `Value` is stored within the corresponding BlockMore Element.
```
robUx4, line 5: method *to* store ?

block_additional_mappings/key_value.md
```diff
@@ -0,0 +1,17 @@
...
+
+### BlockAddIDType
+
+The BlockAddIDType value reserved for timecode is 107.
```
robUx4, line 9: Is it related to timecodes only of it could be used for anything ? If so which timecode format can use this ?

We can only have one Block Additional mapping with the `BlockAddIDType` (107) per track. Otherwise it would be impossible to tell the to which the `BlockMore` data belong to. So this format makes a loosely typed Block Addition but with a single possible use per track. And yet it doesn't tell how to interpret the data. So I'm not sure it's really useful.

robUx4: Poke on this.

mbunkus:
If I understand this PR correctly, there would not be a way to associate more than a single key/value pair with each frame. That I definitely don't like.

Additionally we don't have a way of associating a `BlockMore` with a certain frame inside the `BlockGroup`. Either we add one (via a new frame-number-in-block element inside `BlockMore`), or we mention explicitly that a `BlockMore` applies to all frames in the `BlockGroup`'s `Block`, and that writers that want to associate the `BlockMore` with a specific frame only would need to limit the number of frames in the `Block` to 1. Or we could spec that the `BlockMore` is always associated with the first frame only.

mkver:
In FFmpeg packets can have AV_PKT_DATA_STRINGS_METADATA side data, a list of zero terminated key/value strings. The whole side data is externally framed. Putting such a thing in a `BlockAdditional` would overcome the restriction to a single key-value pair; but it would have the downside that key would have to repeated every time.

Another way to overcome this restriction would be to allow several `BlockAdditionalMapping`s with the same `BlockAddIDType`, but different `BlockAddIDValue`. Each of these would have its own key.

robUx4:
IMO the FFmpeg usage sounds like the "key/value" type is too broad. It's a way to store the data, but it doesn't say what the data correspond to. There might be different systems using the same key.

It should be a `BlockAddIDType` per usage (one would be for FFmpeg) and then saying it's using key/value storage.

robUx4:
Rebased so it can be used. (SMPTE ST 12-1 has been merged)

robUx4:
Rebased on top of #936 to simplify.

There is nothing is the spec that says the same `BlockAddType` can't be used multiple times in the same `BlockGroup` so this seems OK to me.

block_additional_mappings/key_value.md
robUx4: Fixed in the latest version.

block_additional_mappings/key_value.md
robUx4: Reference to timecode is gone.

robUx4:
Moving this to Matroska v5 as this is not used yet (or in Rawcooked ?). (if it's merged before we finalize the codec document, it's fine).

PR #354, opened 2019-11-03
remove the element occurences from the path

robUx4:
Merge once https://github.com/cellar-wg/ebml-specification/pull/302 is merged.

### DIFF
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -1108,7 +1108,7 @@
     <documentation lang="en" purpose="definition">Specify if the chapters can be defined multiple times and the order to play them is enforced. (1 bit)</documentation>
     <extension webm="0"/>
   </element>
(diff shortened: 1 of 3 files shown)


robUx4:
BTW I noticed that an empty `name=""` in the Matroska Schema is considered valid. Is there a way to tell in the EBML XSD that an attribute is mandatory AND must not be empty ?

robUx4:
Found it ```    <xs:attribute name="name" use="required">
      <xs:simpleType>
        <xs:restriction base="xs:string">
          <xs:pattern value="([0-9A-Za-z.-])+"/>
        </xs:restriction>
      </xs:simpleType>
    </xs:attribute>
```

robUx4:
I updated the branch as the `minOccurs` and `maxOccurs` were missing from the cleaned `matroska_xsd.xml` file. Now the output is equivalent to the output before this branch.

The hierarchy of elements has gone up one level, so we get more elements in the index (TrackEntry could still go down one level at least).

robUx4:
Updated with the latest path syntax.

PR #359, opened 2019-12-18
use octet rather than byte

retokromer:


robUx4:
I'd rather keep it like that so the text matches the ISOBMFF spec https://aomediacodec.github.io/av1-isobmff/#subsample-encryption

retokromer:
Hmm… Why should our document match one that is less precise?

dericed:
I think that there's important value in having our document matching the referenced one, regardless of our judgement of its precision, in order to not provoke a conflict interpretation in a place where we aspire to simply adopt. I'm aware that the definitions of byte and octet don't completely align, but am very skeptical that AV1 have ever been implemented in an environment with non-octet bytes.
If important, I'd suggest addressing this upstream first, rather than creating a mismatch between the referenced source.

JeromeMartinez:
I am also in favor of keeping the match with "upstream" (here, AV1 specs).

retokromer:
no problem

PR #360, opened 2019-12-26
unify `16-byte`

retokromer:


robUx4:
TIL it's considered an adjective, and thus is not plural.

PR #362, opened 2020-01-04
use indent 2 everywhere

retokromer:


transforms/ebml_schema2clean.xsl
```diff
@@ -108,9 +108,9 @@
   <xsl:template name="parsePath">
     <xsl:param name="Path"/>
     <xsl:variable name="EBMLElementOccurrence" select="substring-before($Path,'(')"/>
-    <xsl:variable name="EBMLMinOccurrence"     select="substring-before($EBMLElementOccurrence,'*')"/>
-    <xsl:variable name="EBMLMaxOccurrence"     select="substring-after($EBMLElementOccurrence,'*')"/>
-    <xsl:variable name="EBMLMasterPath"   select="substring-before(substring-after($Path,'('),')')"/>
+    <xsl:variable name="EBMLMinOccurrence" select="substring-before($EBMLElementOccurrence,'*')"/>
```
robUx4: The indentation here is done so it's more readable and easier to compare what each does.
In the end this file is not used (and this part won't make sense once we remove the occurences from the EBML path).

retokromer: okay

PR #377, opened 2020-05-10
Add AVC and HEVC codec mappings

### LINKED ISSUE #373: Write down how extra data is saved for usage with AVC & HEVC
Hi,

The creator of MakeMKV have made a custom solution for saving Dolby Vision data within a mkv file.

https://makemkv.com/forum/viewtopic.php?f=12&t=21937

Is it possible to look into how we can make this a part of the standard, so we going forward can support Dolby Vision in matroska?

The creator says:

"Dolby Vision is not a separate codec, but rather a (HEVC) stream extension - DV data (EL+RPU) are encoded in separate NAL units, according to Dolby spec.
The presence of DV data can be checked by several different methods - most obvious are:
(1) DV descriptor in codecprivate data (there is an "standard" of storing additional elements in codec private for AVC and HEVC streams, that was used for ages since 3D MVC video) or
(2)presence of DV RPU unit in first frame."

Hope this can be looked into and approved so the solution can become a part of the standard instead of a custom solution

JeromeMartinez:
AVC and HEVC are widely used in Matroska, but without explicit codec mapping.

This PR adds the mapping of AVC and HEVC in Matroska, including the definition of extension blocks in `CodecPrivate`.

Fix #373.

### DIFF
--- a/codec_specs.md
+++ b/codec_specs.md
@@ -182,6 +182,26 @@ Description: The Matroska video stream will contain a demuxed Elementary Stream 
 
 Initialization: none
 
+### V_MPEG4/ISO/AVC
+
+Codec ID: V_MPEG4/ISO/AVC
+
+Codec Name: AVC/H.264
+
+Description: Individual pictures of AVC/H.264 stored as described in [@!ISO.14496-15.2014].
+
+Initialization: The `Private Data` contains a `AVCDecoderConfigurationRecord` structure as defined in [@!ISO.14496-15.2014], optionally followed by one or more [extension blocks](#private-data-extension-blocks).
+
+### V_MPEGH/ISO/HEVC
+
+Codec ID: V_MPEGH/ISO/HEVC
+
+Codec Name: HEVC/H.265
+
+Description: Individual pictures of HEVC/H.265 stored as described in [@!ISO.14496-15.2014].
+
+Initialization: The `Private Data` contains a `HEVCDecoderConfigurationRecord` structure as defined in [@!ISO.14496-15.2014], optionally followed by one or more [extension blocks](#private-data-extension-blocks).
+
 ### V_REAL/RV10
 
 Codec ID: V_REAL/RV10
@@ -763,3 +783,8 @@ Codec Name: VobBtn Buttons
 
 Description: Based on [MPEG/VOB PCI packets](http://dvd.sourceforge.net/dvdinfo/pci_pkt.html). The file contains a header consisting of the string "butonDVD" followed by the width and height in pixels (16 bits integer each) and 4 reserved bytes. The rest is full [PCI packets](http://dvd.sourceforge.net/dvdinfo/pci_pkt.html).
 
+## Private Data Extension Blocks
+
+When indicated in the codec description as possible, `Private Data` codec specific data MAY be followed by one or more extension blocks.
+
+Each extension block begins with a 4-byte extension block size field which is the size of the extension block minus 4 (excluding the size of the extension block size field), followed by an extension block identifier field.
--- a/codec_specs.md
+++ b/codec_specs.md
@@ -788,3 +788,29 @@ Description: Based on [MPEG/VOB PCI packets](http://dvd.sourceforge.net/dvdinfo/
 When indicated in the codec description as possible, `Private Data` codec specific data MAY be followed by one or more extension blocks.
 
 Each extension block begins with a 4-byte extension block size field which is the size of the extension block minus 4 (excluding the size of the extension block size field), followed by an extension block identifier field.
+
+Registered `Private Data` extension block identifier fields are:
+
+### dvcC
+
+Block identifier: 0x64766343
+
+Extension name: Dolby Vision configuration
+
+Description: `DOVIDecoderConfigurationRecord` structure as defined in [Dolby Vision Streams File Format](https://www.dolby.com/us/en/technologies/dolby-vision/dolby-vision-bitstreams-within-the-iso-base-media-file-format-v2.1.2.pdf).
+
+### hvcE
+
+Block identifier: 0x68766345
+
+Extension name: Dolby Vision enhancement-layer HEVC configuration
+
+Description: `HEVCDecoderConfigurationRecord` structure as defined in [@!ISO.14496-15.2014], as described in [Dolby Vision Streams File Format](https://www.dolby.com/us/en/technologies/dolby-vision/dolby-vision-bitstreams-within-the-iso-base-media-file-format-v2.1.2.pdf). This extension MUST NOT be used if `Codec ID` is not `V_MPEGH/ISO/HEVC`.
+
+### mvcC
+
+Block identifier: 0x6D766343
+
+Extension name: MVC configuration
+
+Description: `MVCDecoderConfigurationRecord` structure as defined in [@!ISO.14496-15.2014]. This extension MUST NOT be used if `Codec ID` is not `V_MPEG4/ISO/AVC`.
--- a/rfc_backmatter_codec.md
+++ b/rfc_backmatter_codec.md
@@ -22,3 +22,14 @@
   </front>
   <seriesInfo name="IEEE" value="Standard 754" />
 </reference>
+
+<reference anchor="ISO.14496-15.2014">
+  <front>
+    <title>Information technology — Coding of audio-visual objects — Part 15: Carriage of network abstraction layer (NAL) unit structured video in ISO base media file format</title>
+    <author>
+      <organization>International Organization for Standardization</organization>
+    </author>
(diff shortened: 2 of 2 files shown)


codec_specs.md
```diff
@@ -182,6 +182,26 @@ Description: The Matroska video stream will contain a demuxed Elementary Stream
...
 
+### V_MPEG4/ISO/AVC
+
+Codec ID: V_MPEG4/ISO/AVC
+
+Codec Name: AVC/H.264
+
+Description: Individual pictures of AVC/H.264 stored as described in ISO/IEC 14496-15.
+
+Initialization: The `Private Data` contains a `AVCDecoderConfigurationRecord` structure as defined in ISO/IEC 14496-15, optionally followed by 1 or more extension blocks; each extension block begins with a 4-byte extension block size field which is the size of the extension block minus 4 (excluding the size of the extension block size field), followed by a 4-byte extension block identifier field; extension block identifier field MAY be identical to the ones found in ISO/IEC 14496-12 and its extensions.
```
retokromer, line 193: I guess, the reviewers prefer `followed by one or more` rather than `followed by 1 or more`.

codec_specs.md
```diff
@@ -182,6 +182,26 @@ Description: The Matroska video stream will contain a demuxed Elementary Stream
...
+
+### V_MPEGH/ISO/HEVC
+
+Codec ID: V_MPEGH/ISO/HEVC
+
+Codec Name: HEVC/H.265
+
+Description: Individual pictures of HEVC/H.265 stored as described in ISO/IEC 14496-15.
+
+Initialization: The `Private Data` contains a `HEVCDecoderConfigurationRecord` structure as defined in ISO/IEC 14496-15, optionally followed by 1 or more extension blocks; each extension block begins with a 4-byte extension block size field which is the size of the extension block minus 4 (excluding the size of the extension block size field), followed by a 4-byte extension block identifier field; extension block identifier field MAY be identical to the ones found in ISO/IEC 14496-12 and its extensions.
```
retokromer, line 203: ditto

steffenmanden:
Still lacks a couple of reviews @mbunkus  and @robUx4 :)

JeromeMartinez:
I modified the patch with a dedicated `Private Data Extension Blocks` chapter, and codec descriptions link to it.

In a 2nd patch (in this PR) I add the 3 currently known and used block identifiers.

Please review again.

rcombs:
I'm a bit concerned about this Dolby Vision data being placed in CodecPrivate, and particularly having it placed directly after the MP4-style configuration data with no other boundary. This data will be useful to the high-level application (for detecting Dolby Vision streams, distinguishing BL from EL) in addition to being used by the decoder itself. This means the demuxer will want to expose the contents of the DOVIDecoderConfigurationRecord to the application as stream metadata. By placing it after a variable-length structure with no central length code, this adds a requirement that the demuxer internally parses the AVCDecoderConfigurationRecord/HEVCDecoderConfigurationRecord to determine its length, skip it, _then_ read the extension blocks. This also means the Matroska CodecPrivate's contents would differ from that used in ISOBMFF, which means existing tools converting Matroska to MP4 will confusingly produce output with Matroska-style data.
I'd be inclined to either move these into a dedicated "private data extensions" element (consistent with how these are represented in other containers), or add an element indicating the offset of the extension blocks (which still has the potential backwards-compatibility issues but at least doesn't complicate parsing).

mbunkus:
I'm not too happy that we're discussing this here instead of the corresponding issue #373. I'll continue there.

codec_specs.md
```diff
@@ -182,6 +182,26 @@ Description: The Matroska video stream will contain a demuxed Elementary Stream
 
 Initialization: none
 
+### V_MPEG4/ISO/AVC
+
+Codec ID: V_MPEG4/ISO/AVC
+
+Codec Name: AVC/H.264
+
+Description: Individual pictures of AVC/H.264 stored as described in ISO/IEC 14496-15.
```
robUx4, line 191: Does the reference like `[@!ISO.14496.15]` work when you generate the file ?
Also since that's more ISOBMFF related, shouldn't there be a like to the H264 specs ? Like `ITU-T Rec. H.264` or even `[@!ITU.REC.H264]` if possible. The list of "tags" that can be used as reference is [in here](https://xml2rfc.tools.ietf.org/public/rfc) but kinda hard to find something.

codec_specs.md
```diff
@@ -763,3 +783,34 @@ Codec Name: VobBtn Buttons
 
 Description: Based on [MPEG/VOB PCI packets](http://dvd.sourceforge.net/dvdinfo/pci_pkt.html). The file contains a header consisting of the string "butonDVD" followed by the width and height in pixels (16 bits integer each) and 4 reserved bytes. The rest is full [PCI packets](http://dvd.sourceforge.net/dvdinfo/pci_pkt.html).
 
+## Private Data Extension Blocks
+
+When indicated in the codec description as possible, `Private Data` codec specific data MAY be followed by one or more extension blocks.
+
+Each extension block begins with a 4-byte extension block size field which is the size of the extension block minus 4 (excluding the size of the extension block size field), followed by an extension block identifier field; extension block identifier field MAY be identical to the ones found in ISO/IEC 14496-12 and its extensions.
```
robUx4, line 790: The ISO reference seem to be specific to some ISO defined codecs. It should probably go in each individual codec.

codec_specs.md
```diff
@@ -763,3 +783,34 @@ Codec Name: VobBtn Buttons
...
+
+Registered `Private Data` extension block identifier fields are:
```
robUx4, line 792: Each codec should probably tell which indentifier are OK. For example `hvcE` shouldn't be legal in H264.

codec_specs.md
JeromeMartinez: > Also since that's more ISOBMFF related, shouldn't there be a like to the H264 specs ?

H.264 specs don't define the "file format" used in Matroska, only the "byte stream format", and ISO/IEC 14496-15 defines the "file format" used by Matroska. If you use H.264 specs only the individual frames are not decodable (the "file format" used is not described). So IMO reference to H.264 directly is not relevant here (Matroska references ISO/IEC 14496-15 which references the needed H.264 parts i.e. without the "byte stream format" stuff).

I'll try to used `[@!ISO.14496.15]` method.

codec_specs.md
JeromeMartinez: I removed the ISO reference.

codec_specs.md
JeromeMartinez: I limited `hvcE` and `mvcC` extensions to a specific `Codec ID`.

codec_specs.md
```diff
@@ -182,6 +182,26 @@ Description: The Matroska video stream will contain a demuxed Elementary Stream
 
 Initialization: none
 
+### V_MPEG4/ISO/AVC
+
+Codec ID: V_MPEG4/ISO/AVC
+
+Codec Name: AVC/H.264
+
+Description: Individual pictures of AVC/H.264 stored as described in [@!ISO.14496.15].
```
robUx4, line 191:
``` draft-ietf-cellar-codec-04.xml(0): Error: IDREF attribute target references an unknown ID "ISO.14496.15", at None
draft-ietf-cellar-codec-04.xml(0): Error: IDREF attribute target references an unknown ID "ISO.14496.15", at None
draft-ietf-cellar-codec-04.xml(0): Error: IDREF attribute target references an unknown ID "ISO.14496.15", at None
draft-ietf-cellar-codec-04.xml(0): Error: IDREF attribute target references an unknown ID "ISO.14496.15", at None
draft-ietf-cellar-codec-04.xml(0): Error: IDREF attribute target references an unknown ID "ISO.14496.15", at None
draft-ietf-cellar-codec-04.xml(0): Error: IDREF attribute target references an unknown ID "ISO.14496.15", at None
draft-ietf-cellar-codec-04.xml(3): Error: Invalid document before running preptool.
```

This results in [errors in the CI](https://github.com/cellar-wg/matroska-specification/pull/377/checks?check_run_id=699125042). I guess this reference is not possible. We'd have to create it ourselves, like this [file in EBML](https://raw.githubusercontent.com/cellar-wg/ebml-specification/master/rfc_backmatter.markdown).

JeromeMartinez: I need help here, I added "ISO.14496-15.2014" (BTW, I changed the ID in order to reference the date) to the backmatter but I still get the error.

robUx4: You put it in ` rfc_backmatter_matroska.md` instead of `rfc_backmatter_codec.md`.
And you used the `—` character which is UTF-8, forbidden in RFCs.

robUx4:
Closing in favor of #390 ?

PR #380, opened 2020-05-22
XML-escape HTML elements in <documentation> nodes

mbunkus:
The documentation uses HTML tags such as `<a>` for links, `<br/>` for line
breaks or `<strong>` for emphasis. Unfortunately those were not
XML-escaped. That meant that XML parsers determined `<a href=…>` to be a
whole new child node of the `<description>` instead of being part of the
text of the `<description>`.

In order to reproduce the HTML-equivalent of such child ones,
transformations or programs would have to iterate recursively over all
child nodes of `<documentation>`, iterate over all existing attributes
etc. Very tedious and superfluous.

### DIFF
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -89,8 +89,8 @@ The format depends on the ChapProcessCodecID used; see (#chapprocesscodecid-elem
     <extension cppname="TimecodeScale"/>
   </element>
   <!-- <element name="TimestampScaleDenominator" parent="/Segment/Info" level="2" id="0x2AD7B2" type="uinteger" minOccurs="1" maxOccurs="1" minver="4" default="1000000000">
-    <documentation lang="en">Timestamp scale numerator; see <a href="https://www.matroska.org/technical/elements.html#TimestampScale">TimestampScale</a>.</documentation>
-TimestampScale When combined with <a href="https://www.matroska.org/technical/elements.html#TimestampScaleDenominator">TimestampScaleDenominator</a> the Timestamp scale is given by the fraction TimestampScale/TimestampScaleDenominator in seconds.-->
+    <documentation lang="en">Timestamp scale numerator, see &lt;a href="https://www.matroska.org/technical/elements.html#TimestampScale"&gt;TimestampScale&lt;/a&gt;.</documentation>
+    TimestampScale When combined with &lt;a href="https://www.matroska.org/technical/elements.html#TimestampScaleDenominator"&gt;TimestampScaleDenominator&lt;/a&gt; the Timestamp scale is given by the fraction TimestampScale/TimestampScaleDenominator in seconds.-->
   <element name="Duration" path="\Segment\Info\Duration" id="0x4489" type="float" range="&gt; 0x0p+0" maxOccurs="1">
     <documentation lang="en" purpose="definition">Duration of the Segment in nanoseconds based on TimestampScale.</documentation>
   </element>


mbunkus:
Elaborating a bit more why this change is needed. Here's a small sample in Ruby that parses the XML file & outputs the text part of the `ChapterTranslateID` element _without this PR applied_:

```
[0 mosu@sweet-chili ~/prog/video/matroska-specification] cat ~/test/ruby1.rb
#!/usr/bin/env ruby

require 'rexml/document'

doc = REXML::Document.new File.new("ebml_matroska.xml")
puts doc.get_elements("//element[@name='ChapterTranslateID']/documentation").first.text
[0 mosu@sweet-chili ~/prog/video/matroska-specification] ~/test/ruby1.rb
The binary value used to represent this Segment in the chapter codec data. The format depends on the
[0 mosu@sweet-chili ~/prog/video/matroska-specification]
```

XML parsers in other languages will behave the same way: the `<a>` tag is a child element of `<documentation>`, even though semantically it should be part of the whole text of the `<documentation>` node.

You have similar issues with XSL stylesheets. Simply outputting the content of the `<documentation>` node will strip all of the XML-ized HTML tags. Here's a stylesheet example without this PR applied:

```
[0 mosu@sweet-chili ~/prog/video/matroska-specification] cat ~/test/chapter_translate_id.xsl
<?xml version="1.0"?>
<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" version="1.0"
   xmlns:ebml="https://ietf.org/cellar/ebml" exclude-result-prefixes="ebml">
  <xsl:output encoding="UTF-8" method="html" version="1.0"/>

  <xsl:template match="ebml:EBMLSchema">
    <xsl:value-of select="ebml:element[@name='ChapterTranslateID']/ebml:documentation" disable-output-escaping="yes"/>
  </xsl:template>
</xsl:stylesheet>
[0 mosu@sweet-chili ~/prog/video/matroska-specification] xsltproc ~/test/chapter_translate_id.xsl ebml_matroska.xml
<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.0 Transitional//EN" "http://www.w3.org/TR/REC-html40/loose.dtd">
The binary value used to represent this Segment in the chapter codec data. The format depends on the ChapProcessCodecID used.
[0 mosu@sweet-chili ~/prog/video/matroska-specification]
```

With the PR applied the results are what we want to see:

```
[0 mosu@sweet-chili ~/prog/video/matroska-specification] ~/test/ruby1.rb
The binary value used to represent this Segment in the chapter codec data. The format depends on the <a href="https://www.matroska.org/technical/specs/chapters/index.html#ChapProcessCodecID">ChapProcessCodecID</a> used.
[0 mosu@sweet-chili ~/prog/video/matroska-specification] xsltproc ~/test/chapter_translate_id.xsl ebml_matroska.xml
<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.0 Transitional//EN" "http://www.w3.org/TR/REC-html40/loose.dtd">
The binary value used to represent this Segment in the chapter codec data. The format depends on the <a href="https://www.matroska.org/technical/specs/chapters/index.html#ChapProcessCodecID">ChapProcessCodecID</a> used.
[0 mosu@sweet-chili ~/prog/video/matroska-specification]
```

robUx4:
It is valid according to the EBML Schema. There are specifically 3 [XHTML elements allowed in documentation](https://github.com/cellar-wg/ebml-specification/blob/master/EBMLSchema.xsd#L135): `a`, `br` and `strong`. So the XML is correct and valid. The XSLT code already has some tricks to handle them properly in the output.

So IMO there is no issue.

There might be one with the [extra documentation I want to add to enums](https://github.com/cellar-wg/ebml-specification/pull/339).

mbunkus:
I'm not necessarily talking about validity here, but more about ease of use. Extensibility is also an issue; we're very limited if those are all the HTML elements we might use in the description.

robUx4:
I think idea is precisely to limit so we don't have to support a whole bunch (hundreds? thousands?) or elements. IMO only `<a>` still make sense. In the RFC there is no "strong" in the end. And new lines are done in the Markdown anyway. That's at least for the purpose of EBML Schema to RFC. In this case even `<a>` should not be supported or transformed into a reference at the end of the document. That's probably an issue we have to deal with. That also means any other elements would have to be stripped, and how is a `<br/>` element supposed to be handled or a `<table>` ?

The EBML Schema can be used for other things (like make our website) so it's still OK to have at least the proper XHTML links in it. The rest probably not.
(BTW, if we automate the website generation further with Jekyll I suggest generating a Markdown version of table instead of the HTML page, then the Markdown can be included in another page via Jekyll).

dericed (approved):
Seems ok to me, though if these element was later uncommented, I'd suggest rewording to avoid the <a href...>.

robUx4:
Rebasing the last remaining change with make this commit empty.
Reopen if the issue remains elsewhere.

Issue #379, opened 2020-05-22
ebml_schema2html.xsl "max" column wrong if no maxOccurs attribute


mbunkus:
For example, the `BlockAdditionMapping` element is defined as having neither a `minOccurs` nor a `maxOccurs` attribute, However, the "max" column created contains the value 1, which is obviously wrong for such elements.

PR #381, opened 2020-05-22
Ebml schema2html fixes

### LINKED ISSUE #378: ebml_schema2html.xsl transform creates empty "level" column
Running the following yields nice HTML, but the "Level" column is completely empty:

```
xsltproc transforms/ebml_schema2html.xsl ebml_matroska.xml
```

Additionally the row's `class` attribute is incomplete, for example:

```
  <tr class="level">
```

robUx4:
Fixes #378 and #379


mbunkus:
Works fine, thanks.

PR #382, opened 2020-05-22
add an XSLT script to replace data2spec

robUx4:
So we can generate XHTML code that matches the style of the old website.

The XML formatting isn't 100% exact as XSLT produces clean line returns, indentation and empty elements.

Comparing with the current matroska.org there are many new elements, many additions in WebM. So element description text is also a bit different.

### DIFF
--- a/transforms/ebml_schema2spec.xsl
+++ b/transforms/ebml_schema2spec.xsl
@@ -1,0 +1,469 @@
+<?xml version="1.0"?>
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" version="1.0"
+   xmlns:ebml="https://ietf.org/cellar/ebml" exclude-result-prefixes="ebml">
+  <xsl:output encoding="UTF-8" method="xml" version="1.0" indent="yes"/>
+  <xsl:template match="ebml:EBMLSchema">
+    <html>
+      <style>
+        .techdef table{
+          font-family:sans-serif;
+          background:#fff;
+          margin:0;
+          padding:0;
+        }
+
+        .techdef th{
+          font-size:larger;
+          border:5px solid #ddd;
+          background:#eee;
+          padding:.5em 0 .5em 0;
+          margin:0;
+        }
+
+        .techdef tr{
+          margin:0;
+          padding:0;
+        }
+
+        .techdef td{
+          margin:0;
+          border:1px solid #eee;
+          padding:2px;
+        }
+
+        .level0{
+          background:#fff;
+        }
+
+        .level1{
+          background:#eff;
+        }
+
+        .level2{
+          background:#eef;
+        }
+
+        .level3{
+          background:#dde;
+        }
+
+        .level4{
+          background:#ccd;
+        }
+
+        .level5{
+          background:#bbc;
+        }
+
+        .level6{
+          background:#aab;
+        }
+
+        .level7{
+          background:#99a;
+        }
+
+        .level8{
+          background:#889;
+        }
+
+        .version2{
+          color:#aaa;
+          background:#eee;
+        }
+      </style>
+      <div class="techdef">
+      <h1><xsl:value-of select="@docType"/></h1>
+      <p>Version: <xsl:value-of select="@version"/></p>
+      <table class="specstable">
+        <xsl:call-template name="OutputHeader"/>
+
+        <xsl:apply-templates select="//ebml:element"/>
+
+        <xsl:call-template name="OutputHeader"/>
+      </table>
+    </div>
+    </html>
+  </xsl:template>
+
+  <xsl:template name="GetLevel">
+    <xsl:param name="String"/>
+    <xsl:param name="Level"/>
+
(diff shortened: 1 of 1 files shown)


mbunkus:
Yeah, thanks. I've replaced the web site with output generated by my Ruby script (which is, as said, closer to what `ebml_schema2html.xsl` generates). Unfortunately the specs seem quite bad wrt. to the WebM property; therefore I disabled that column temporarily.

I'm planning on fixing the WebM flags in the spec in the next couple of days.

As I prefer the new table style/information (especially as they're more precise) I will likely not use the `data2spec`-style stylesheet.

mbunkus:
Now that I think about it — is the absence of `<extension webm="0"/>` an indication that the element is actually part of WebM? Why is there `<extension webm="1"/>` sometimes, then? I find that highly confusing.

robUx4:
I think the webm="1" was just replacing some elements that had webm="0" but are actually valid.

robUx4:
Mh, the look is quite different. Compared to what I generate you kept the links (I'll have to work on that). But it's missing all the enum values :( And the IDs are in hexadecimal but in a little endian machine it may not end up as intended. I preferred the notation with braces.

mbunkus:
For the links, just merge #380.

mbunkus:
Yeah, I know about the enums; I'll tackle those tomorrow.

mbunkus:
> And the IDs are in hexadecimal but in a little endian machine it may not end up as intended. 

1. You cannot copy-paste the variants in braces. I've always disliked that fact. Makes for easy manually-typed-but-wrong mistakes.
2. Big vs Little Endian only comes into play when you write data directly from memory to file or the other way around. Simply using the ID in source code is perfectly fine. We use the same notation in libMatroska(2), after all, and no, we don't have two sections in e.g. `src/KaxSemantic.cpp`, one with the IDs in Little Endian, one in Big Endian.

robUx4:
You could at least ask before changing the content of the website with a very different content and half of the content missing. I don't think there was a rush to set the website in a half baked version as it is now.
I did the tools to generate the website before and I already had worked on doing it after all the changes in the XML sources.

mbunkus:
Yeah, I'm sorry the intermediate result was so incomplete. I would probably have reverted, but Drupal doesn't store all changes, only those that one explicitly marks as a revision (which I didn't know about until after the fact). That's also why I pimped the `ebml_schema2html.xsl` yesterday evening & updated the website with its latest output — meaning the website isn't incomplete anymore.

I'm not trying to change the content, nor do I want to change the table radically. That being said, I do have reasons for the change:

1. The old table was incomplete wrt. the description. While it did include the enumerated valid values, it didn't include their description. As www.matroska.org is still the first point of contact for everyone looking for the specs, I consider it to be important to have it complete. Yes, even more reason for me not to have replaced the existing content with incomplete one; I know.
2. The form of presentation of several spec parameter on the website differed quite a bit from how it's presented in `ebml_matroska.xml`. My goal was to have both presentations be closer so the people interested in the specs could easily make the connection between what they see in both places.
3. That list of columns with `*` in it (or not) was pretty hard to read; you'd really have to scroll up all the time in order to determine which column indicates which feature/Matroska version. I consider `2-4` to be much easier to read.
4. Searching for given element IDs is easier if all you have to do is type `123abc` instead of having to type `[12][3a][bc]`, and it makes copy-pasting between source code & the website much easier. Our spec was pretty much the only place where IDs were written using brackets.

I'm fine with changing certain aspects back to what they were before, e.g. using single-letter abbreviations for the type, and if you're feeling strongly about the element IDs using brackets, I wouldn't object. I do feel strongly about the presentation of min/max/range/version, though, and would like to see some arguments why the keeping the old form would be better.

robUx4:
This new version produces XHTML code that matches the old design and data2spec (with cleaner XML formatting). Having a consistent format with the old version allowed me to [find an issue in ebml_matroska.xml](https://github.com/cellar-wg/matroska-specification/commit/6ed1232b8d60a76ebb39038453e5a3d038c1a724) because I could compare what we had before and what we generate now. (which is not possible when you change the format).

The design of the old page that I kept all this time wasn't random. It was designed to be as readable in one page as I could come up. 

- Using one letter per column with mouse-over a description was one of the nice features that is now gone. It can easily be extended to the columns with a `*` in it and have even more text for each.
- the mult./mand. text was clearer than 0 or 1 in a column (min/max somes from XML but IMO is less clear than knowing if an element is mandatory and can be found more than once)
- the deprecated elements was greyed out so it's clearer those don't matter
- there was more space for the description so there's more than 4 words per line

I understand the concern with the IDs, I prefer the old way but understand one might want to copy/paste one.

Your version is still missing crucial information found in the `usage notes` or `implementation_note`. For example the "default" value for DisplayWidth is gone. That's also a reason why, before changing the design you should have matched the content with the existing source.

mbunkus:
Yeah OK, let's go back.

mbunkus:
I've replaced the website's content with the output of `ebml_schema2spec.xsl` again. I'll add a couple of potential changes to the stylesheet and create a new PR for discussion.

PR #383, opened 2020-05-22
Spec to html table improvements

mbunkus:
This implements a lot of enhancements & changes to the `ebml_spec2html.xsl` stylesheet in order to generate the element spec table visible at www.matroska.org

As the stylesheet now includes the git revision of `ebml_matroska.xml` it was running with, a script has been added that determines the git revision & author date of the top-most commit. The script then calls `xsltproc` with the stylesheet & `ebml_matroska.xml`, passing the git information as a parameter. The script can be run with:

```
./transforms/ebml_spec2html.sh
```

This PR requires PR #380 to be committed. It's based on it, in fact.

This PR pretty much obsoletes the need for the old, `data2spec`-style table. It therefore also supersedes PR #382.

### DIFF
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -65,7 +65,7 @@
     <extension webm="0"/>
   </element>
   <element name="ChapterTranslateCodec" path="\Segment\Info\ChapterTranslate\ChapterTranslateCodec" id="0x69BF" type="uinteger" minOccurs="1" maxOccurs="1">
-    <documentation lang="en" purpose="definition">The <a href="https://www.matroska.org/technical/specs/index.html#ChapProcessCodecID">chapter codec</a></documentation>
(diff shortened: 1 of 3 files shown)


mbunkus:
Note: I'm perfectly aware that the table as it is currently shown on www.matroska.org is very wide. I'm planning on migrating the website from the current CMS, Drupal, to Wordpress rather soon as this Drupal installation and the corresponding theme are very, very hard to maintain (at the moment I cannot even add more custom CSS due to… reasons involving Node.JS, being too old or too new or the code being too old for the newer Node.JS, some modules not being installable… don't ask). With that migration I also plan on using more horizontal space, which should benefit the spec table.

mbunkus:
Using Steve's implementation from PR #382 instead.

PR #385, opened 2020-05-23
Jekyll xslt

robUx4:
We can now generate the specs (the legacy way) in the static website on GitHub.

The XSLT transformation of the EBML Schema is done in the browser.


robUx4:
Updated the names of the stylesheets to keep using the same name assumed in #384

_includes/head.html
```diff
@@ -10,3 +10,71 @@
   <link rel="canonical" href="{{ page.url | replace:'index.xml','' | prepend: site.baseurl | prepend: site.url }}" />
   <link rel="alternate" type="application/rss+xml" title="{{ site.title }}" href="{{ "/feed.xml" | prepend: site.baseurl | prepend: site.url }}" />
 </head>
+<style>
```
mbunkus, line 13: Technically, `<style>` is only allowed as a child of `<head>` (see [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/style) and elsewhere).

retokromer:
I will test it later during the weekend. Just one nit: you don’t need to put a space before you close an empty tag (`<foo="bar" \>`). This was needed in the late 1990s for one then popular browser, but today all tools should process regular XML.

_includes/head.html
robUx4: Thanks, I always wonder about that. I'll move it.

robUx4:
Pushing as I can't wait to see the new page on GitHub.

robUx4:
Mh, it doesn't show the new page. Maybe it takes time to propagate...
(I tested locally with `jekyll serve`)

robUx4:
[It's alive!](http://cellar-wg.github.io/matroska-specification/technical-index.xml)

PR #388, opened 2020-05-25
allow extensions to CodecPrivate via BlockAdditionMapping

mbunkus:
This prepares the element specs for the storage of elements such as
the mvcC extension for AVC video or the dvcC and hvcE extensions for
HEVC video. It does so by

* changing the verbiage to make BlockAdditionMapping not only
  referring to BlockAdditions in tracks,
* making BlockAddIDValue non-mandatory which would indicate that the
  BlockAdditionMapping is actually an extension to CodecPrivate and
* includes a first registered value for BlockAddIDType which signals
  that BlockAddIDExtraData contains a type & content like similar
  header atoms in ISO Base Media File Format files.

See #373 for the discussion how to store things. Will affect PR #377
which will have to be adjusted to make use of these elements.

### DIFF
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -317,11 +317,11 @@
     <extension webm="0"/>
   </element>
   <element name="BlockAdditionMapping" path="\Segment\Tracks\TrackEntry\BlockAdditionMapping" id="0x41E4" type="master" minver="4">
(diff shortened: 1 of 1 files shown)


mbunkus:
I like both of your approaches better, and they're obviously more complete. No need for my one anymore :)

PR #390, opened 2020-05-26
Add AVC and HEVC codec mappings with BlockAdditionMapping 2

JeromeMartinez:
Another alternative of #377, using `BlockAddIDType` for storing 4-byte ISOBMFF values.
Discussion on https://github.com/cellar-wg/matroska-specification/issues/373#issuecomment-633879582.

### DIFF
--- a/codec_specs.md
+++ b/codec_specs.md
@@ -803,3 +803,51 @@ Block type identifier: 1
 Block type name: Opaque data
 
 Description: the `BlockAdditional` data is interpreted as opaque additional data passed to the codec with the Block data. `BlockAddIDValue` MUST be 1.
+
+### ITU T.35 metadata
+
+Block type identifier: 4
+
+Block type name: ITU T.35 metadata
+
+Description: the `BlockAdditional` data is interpreted as ITU T.35 metadata as defined by ITU-T T.35 terminal codes. `BlockAddIDValue` MUST be 4.
+
+### avcE
+
+Block type identifier: 0x61766345
+
+Block type name: Dolby Vision enhancement-layer AVC configuration
+
+Type Description: the `BlockAddIDExtraData` data is interpreted as the Dolby Vision enhancement-layer AVC configuration box as described in [@!DolbyVisionWithinIso.2020-02]. This extension MUST NOT be used if `Codec ID` is not `V_MPEG4/ISO/AVC`.
+
+### dvcC
+
+Block type identifier: 0x64766343
+
+Block type name: Dolby Vision configuration
+
+Description: the `BlockAddIDExtraData` data is interpreted as `DOVIDecoderConfigurationRecord` structure as defined in [@!DolbyVisionWithinIso.2020-02], for Dolby Vision profiles less than and equal to 7.
+
+### dvvC
+
+Block type identifier: 0x664767643
+
+Block type name: Dolby Vision configuration
+
+Description: the `BlockAddIDExtraData` data is interpreted as `DOVIDecoderConfigurationRecord` structure as defined in [@!DolbyVisionWithinIso.2020-02], for Dolby Vision profiles greater than 7.
+
+### hvcE
+
+Block type identifier: 0x68766345
+
+Block type name: Dolby Vision enhancement-layer HEVC configuration
+
+Type Description: the `BlockAddIDExtraData` data is interpreted as the Dolby Vision enhancement-layer HEVC configuration as described in [@!DolbyVisionWithinIso.2020-02]. This extension MUST NOT be used if `Codec ID` is not `V_MPEGH/ISO/HEVC`.
+
+### mvcC
+
+Block type identifier: 0x6D766343
+
+Block type name: MVC configuration
+
+Description: the `BlockAddIDExtraData` data is interpreted as `MVCDecoderConfigurationRecord` structure as defined in [@!ISO.14496-15.2014]. This extension MUST NOT be used if `Codec ID` is not `V_MPEG4/ISO/AVC`.
--- a/codec_specs.md
+++ b/codec_specs.md
@@ -784,3 +784,22 @@ Codec Name: VobBtn Buttons
 
 Description: Based on [MPEG/VOB PCI packets](http://dvd.sourceforge.net/dvdinfo/pci_pkt.html). The file contains a header consisting of the string "butonDVD" followed by the width and height in pixels (16 bits integer each) and 4 reserved bytes. The rest is full [PCI packets](http://dvd.sourceforge.net/dvdinfo/pci_pkt.html).
 
+## Block Addition Mappings
+
+Registered `BlockAddIDType` are:
+
+### Use BlockAddIDValue
+
+Block type identifier: 0
+
+Block type name: Use BlockAddIDValue
+
+Description: This value indicates that the actual type is stored in `BlockAddIDValue` instead. This value is expected to be used when it is important to have a strong compatibility with players or derived formats not supporting `BlockAdditionMapping` but using `BlockAdditions` with an unknown `BlockAddIDValue`, and SHOULD NOT be used if it is possible to use another value.
+
+### Opaque data
+
+Block type identifier: 1
+
+Block type name: Opaque data
+
+Description: the `BlockAdditional` data is interpreted as opaque additional data passed to the codec with the Block data. `BlockAddIDValue` MUST be 1.
--- a/codec_specs.md
+++ b/codec_specs.md
@@ -183,6 +183,26 @@ Description: The Matroska video stream will contain a demuxed Elementary Stream 
 
 Initialization: none
 
+### V_MPEG4/ISO/AVC
+
+Codec ID: V_MPEG4/ISO/AVC
+
+Codec Name: AVC/H.264
+
+Description: Individual pictures of AVC/H.264 stored as described in [@!ISO.14496-15.2014].
+
+Initialization: The `Private Data` contains a `AVCDecoderConfigurationRecord` structure as defined in [@!ISO.14496-15.2014].
+
+### V_MPEGH/ISO/HEVC
+
+Codec ID: V_MPEGH/ISO/HEVC
+
+Codec Name: HEVC/H.265
+
+Description: Individual pictures of HEVC/H.265 stored as described in [@!ISO.14496-15.2014].
+
+Initialization: The `Private Data` contains a `HEVCDecoderConfigurationRecord` structure as defined in [@!ISO.14496-15.2014].
+
 ### V_REAL/RV10
 
 Codec ID: V_REAL/RV10
--- a/codec_specs.md
+++ b/codec_specs.md
@@ -191,7 +191,7 @@ Codec Name: AVC/H.264
 
 Description: Individual pictures of AVC/H.264 stored as described in [@!ISO.14496-15.2014].
 
-Initialization: The `Private Data` contains a `AVCDecoderConfigurationRecord` structure as defined in [@!ISO.14496-15.2014].
+Initialization: The `Private Data` contains a `AVCDecoderConfigurationRecord` structure as defined in [@!ISO.14496-15.2014]. For legacy reasons ([Block Addition Mappings](#block-addition-mappings) are preferred), the `AVCDecoderConfigurationRecord` structure MAY be followed by an extension block beginning with a 4-byte extension block size field in Big Endian byte order which is the size of the extension block minus 4 (excluding the size of the extension block size field) and a 4-byte field corresponding to a `BlockAddIDType` of `[mvcC](#mvcC)` followed by a content corresponding to the content of `BlockAddIDExtraData` for `[mvcC](#mvcC)`.
 
 ### V_MPEGH/ISO/HEVC
 
--- a/codec_specs.md
+++ b/codec_specs.md
@@ -189,7 +189,7 @@ Codec ID: V_MPEG4/ISO/AVC
 
 Codec Name: AVC/H.264
 
-Description: Individual pictures of AVC/H.264 stored as described in [@!ISO.14496-15.2014].
+Description: Individual pictures (which could be a frame, a field, or 2 fields having the same timestamp) of AVC/H.264 stored as described in [@!ISO.14496-15.2014].
 
 Initialization: The `Private Data` contains a `AVCDecoderConfigurationRecord` structure as defined in [@!ISO.14496-15.2014]. For legacy reasons ([Block Addition Mappings](#block-addition-mappings) are preferred), the `AVCDecoderConfigurationRecord` structure MAY be followed by an extension block beginning with a 4-byte extension block size field in Big Endian byte order which is the size of the extension block minus 4 (excluding the size of the extension block size field) and a 4-byte field corresponding to a `BlockAddIDType` of `[mvcC](#mvcC)` followed by a content corresponding to the content of `BlockAddIDExtraData` for `[mvcC](#mvcC)`.
 
@@ -199,7 +199,7 @@ Codec ID: V_MPEGH/ISO/HEVC
 
 Codec Name: HEVC/H.265
 
-Description: Individual pictures of HEVC/H.265 stored as described in [@!ISO.14496-15.2014].
+Description: Individual pictures (which could be a frame, a field, or 2 fields having the same timestamp) of HEVC/H.265 stored as described in [@!ISO.14496-15.2014].
 
 Initialization: The `Private Data` contains a `HEVCDecoderConfigurationRecord` structure as defined in [@!ISO.14496-15.2014].
 
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -150,7 +150,7 @@
     <extension webm="1"/>
   </element>
   <element name="BlockAddID" path="\Segment\Cluster\BlockGroup\BlockAdditions\BlockMore\BlockAddID" id="0xEE" type="uinteger" range="not 0" default="1" minOccurs="1" maxOccurs="1">
-    <documentation lang="en" purpose="definition">An ID to identify the BlockAdditional level. A value of 1 means the BlockAdditional data is interpreted as additional data passed to the codec with the Block data.</documentation>
+    <documentation lang="en" purpose="definition">An ID to identify the BlockAdditional level. If BlockAddIDType of the corresponding block is 0, this value is also the value of BlockAddIDType for the meaning of the content of BlockAdditional.</documentation>
     <extension webm="1"/>
   </element>
   <element name="BlockAdditional" path="\Segment\Cluster\BlockGroup\BlockAdditions\BlockMore\BlockAdditional" id="0xA5" type="binary" minOccurs="1" maxOccurs="1">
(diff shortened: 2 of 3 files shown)


codec_specs.md
```diff
@@ -190,7 +190,7 @@ Codec Name: AVC/H.264
 
 Description: Individual pictures of AVC/H.264 stored as described in [@!ISO.14496-15.2014].
 
-Initialization: The `Private Data` contains a `AVCDecoderConfigurationRecord` structure as defined in [@!ISO.14496-15.2014].
+Initialization: The `Private Data` contains a `AVCDecoderConfigurationRecord` structure as defined in [@!ISO.14496-15.2014]; for legacy reasons ([Block Addition Mappings](#block-addition-mappings) are prefered), the `AVCDecoderConfigurationRecord` structure MAY be followed by an extension block beginning with a 4-byte extension block size field which is the size of the extension block minus 4 (excluding the size of the extension block size field) and a 4-byte field corresponding to `BlockAddIDType` followed by a content corresponding to the content of `BlockAddIDExtraData`.
```
mbunkus, line 193: I forgot to mention it in the other PR, but as this exception for legacy files is about one specific use case (`mvcC`) maybe we should limit the exception to that one possible type.

codec_specs.md
```diff
@@ -783,3 +783,22 @@ Codec Name: VobBtn Buttons
...
+
+Registered `BlockAddIDType` are:
+
+### Use BlockAddIDValue
+
+Block type identifier: 0
+
+Block type name: Use BlockAddIDValue
+
+Description: This value indicates that `BlockAddIDValue` should be used as `BlockAddIDType`.
```
mbunkus, line 796: I find this wording to be somewhat confusing (same in the xml). I'd prefer something more like

> This value indicates that the actual type is stored in `BlockAddIDValue` instead.

and maybe some more words mentioning that this is due to compatibility with legacy files or some such.

ebml_matroska.xml
```diff
@@ -150,7 +150,7 @@
     <extension webm="1"/>
   </element>
   <element name="BlockAddID" path="\Segment\Cluster\BlockGroup\BlockAdditions\BlockMore\BlockAddID" id="0xEE" type="uinteger" range="not 0" default="1" minOccurs="1" maxOccurs="1">
-    <documentation lang="en" purpose="definition">An ID to identify the BlockAdditional level. A value of 1 means the BlockAdditional data is interpreted as additional data passed to the codec with the Block data.</documentation>
+    <documentation lang="en" purpose="definition">An ID to identify the BlockAdditional level. If BlockAddIDType of the corresponding block is 0, BlockAddID is considered as the corresponding BlockAddIDType.</documentation>
```
mbunkus, line 153: Same objection about the wording being confusing as in `codec_specs.md`.

ebml_matroska.xml
```diff
@@ -317,11 +317,11 @@
     <extension webm="0"/>
   </element>
   <element name="BlockAdditionMapping" path="\Segment\Tracks\TrackEntry\BlockAdditionMapping" id="0x41E4" type="master" minver="4">
-    <documentation lang="en" purpose="definition">Contains elements that describe each value of <a href="https://www.matroska.org/technical/specs/index.html#BlockAddID">BlockAddID</a> found in the Track.</documentation>
+    <documentation lang="en" purpose="definition">Contains elements that extend the track format, by adding content either to frames (with <a href="https://www.matroska.org/technical/specs/index.html#BlockAddID">BlockAddID</a>) or to the format definition</documentation>
```
mbunkus, line 320: Same objection as in 389: "the format definition" is a very ambiguous term, one that I wouldn't understand as an implementer. The "format definition" is e.g. ISO 14496-12. We don't extend that definition. What we do extend is codec-specific data present in the track headers.

codec_specs.md
```diff
@@ -763,3 +783,54 @@ Codec Name: VobBtn Buttons
...
+
+### dvcC
+
+Block type identifier: 0x64766343
+
+Block type name: Dolby Vision configuration
+
+Description: the `BlockAddIDExtraData` data is interpreted as `DOVIDecoderConfigurationRecord` structure as defined in [Dolby Vision Streams File Format](ttps://www.dolby.com/us/en/technologies/dolby-vision/dolby-vision-bitstreams-within-the-iso-base-media-file-format-v2.1.2.pdf).
+
+### hvcE
```
MikeChenMM, line 822: Please note, Dolby Vision is also possible (and is defined) for AVC streams, not only for HEVC.
In case of AVC related IDs are dvcC and avcE . We probably should change dvcC to be allowed on both HEVC and AVC, and define avcE for DV on AVC.

mbunkus: That's a very good point, Mike. @JeromeMartinez can you please adjust your PR accordingly? The PDF even talks about using Dolby Vision with other codecs such as VP9.

Additionally the PDF also mentions `dvvC` atoms in addition to `dvcC` when talking about Dolby Vision configuration boxes. Are they relevant to us somehow?

codec_specs.md
retokromer: >and maybe some more words mentioning that this is due to compatibility with legacy files

Indeed, I strongly second this.

codec_specs.md
JeromeMartinez: I don't see where I limit `dvcC` to a specific `Codec ID`, I limit only for `mvcC` and `hvcE`.
I'll add `dvvC` and `avcE` to the list.

JeromeMartinez:
@mbunkus @MikeChenMM I updated the PR, please review again.

codec_specs.md
```diff
@@ -802,3 +802,51 @@ Block type identifier: 1
...
+
+Description: the `BlockAdditional` data is interpreted as ITU T.35 metadata as defined by ITU-T T.35 terminal codes. `BlockAddIDValue` MUST be 4.
+
+### avcE
+
+Block type identifier: 0x61766345
+
+Block type name: Dolby Vision enhancement-layer AVC configuration
+
+Type Description: the `BlockAddIDExtraData` data is interpreted as `AVCDecoderConfigurationRecord` structure for the enhancement layer as defined in [@!ISO.14496-15.2014], as described in [Dolby Vision Streams File Format](ttps://www.dolby.com/us/en/technologies/dolby-vision/dolby-vision-bitstreams-within-the-iso-base-media-file-format-v2.1.2.pdf). This extension MUST NOT be used if `Codec ID` is not `V_MPEGH/ISO/AVC`.
```
mbunkus, line 820: Wrong CodecID, must be `V_MPEG4/ISO/AVC` not `MPEGH`.

codec_specs.md
```diff
@@ -190,7 +190,7 @@ Codec Name: AVC/H.264
 
 Description: Individual pictures of AVC/H.264 stored as described in [@!ISO.14496-15.2014].
 
-Initialization: The `Private Data` contains a `AVCDecoderConfigurationRecord` structure as defined in [@!ISO.14496-15.2014].
+Initialization: The `Private Data` contains a `AVCDecoderConfigurationRecord` structure as defined in [@!ISO.14496-15.2014]. For legacy reasons ([Block Addition Mappings](#block-addition-mappings) are prefered), the `AVCDecoderConfigurationRecord` structure MAY be followed by an extension block beginning with a 4-byte extension block size field which is the size of the extension block minus 4 (excluding the size of the extension block size field) and a 4-byte field corresponding to a `BlockAddIDType` of `[mvcC](#mvcC)` followed by a content corresponding to the content of `BlockAddIDExtraData` for `[mvcC](#mvcC)`.
```
mbunkus, line 193: Please mention that the size field is stored in Big Endian.

JeromeMartinez:
> Apart from those two small things [...]

Fixed.

> I'd really like to see a review from @robUx4 , though, before we merge any of our proposals.

@robUx4 🙏

JeromeMartinez:
> Please rebase.

Done.

JeromeMartinez:
is it OK to merge this PR now?
a review from @robUx4 would be great ;-).

codec_specs.md
```diff
@@ -183,6 +183,26 @@ Description: The Matroska video stream will contain a demuxed Elementary Stream
...
 
+### V_MPEG4/ISO/AVC
+
+Codec ID: V_MPEG4/ISO/AVC
+
+Codec Name: AVC/H.264
+
+Description: Individual pictures of AVC/H.264 stored as described in [@!ISO.14496-15.2014].
+
+Initialization: The `Private Data` contains a `AVCDecoderConfigurationRecord` structure as defined in [@!ISO.14496-15.2014]. For legacy reasons ([Block Addition Mappings](#block-addition-mappings) are prefered), the `AVCDecoderConfigurationRecord` structure MAY be followed by an extension block beginning with a 4-byte extension block size field in Big Endian byte order which is the size of the extension block minus 4 (excluding the size of the extension block size field) and a 4-byte field corresponding to a `BlockAddIDType` of `[mvcC](#mvcC)` followed by a content corresponding to the content of `BlockAddIDExtraData` for `[mvcC](#mvcC)`.
```
robUx4, line 194: typo: prefered -> preferred

codec_specs.md
```diff
@@ -183,6 +183,26 @@ Description: The Matroska video stream will contain a demuxed Elementary Stream
 
 Initialization: none
 
+### V_MPEG4/ISO/AVC
+
+Codec ID: V_MPEG4/ISO/AVC
+
+Codec Name: AVC/H.264
+
+Description: Individual pictures of AVC/H.264 stored as described in [@!ISO.14496-15.2014].
```
robUx4, line 192: Is it pictures or fields ? In other words interlaced content should group two fields together ? Does the ISO spec refer to fields or pictures ?

codec_specs.md
```diff
@@ -183,6 +183,26 @@ Description: The Matroska video stream will contain a demuxed Elementary Stream
...
+
+Initialization: The `Private Data` contains a `AVCDecoderConfigurationRecord` structure as defined in [@!ISO.14496-15.2014]. For legacy reasons ([Block Addition Mappings](#block-addition-mappings) are prefered), the `AVCDecoderConfigurationRecord` structure MAY be followed by an extension block beginning with a 4-byte extension block size field in Big Endian byte order which is the size of the extension block minus 4 (excluding the size of the extension block size field) and a 4-byte field corresponding to a `BlockAddIDType` of `[mvcC](#mvcC)` followed by a content corresponding to the content of `BlockAddIDExtraData` for `[mvcC](#mvcC)`.
+
+### V_MPEGH/ISO/HEVC
+
+Codec ID: V_MPEGH/ISO/HEVC
+
+Codec Name: HEVC/H.265
+
+Description: Individual pictures of HEVC/H.265 stored as described in [@!ISO.14496-15.2014].
```
robUx4, line 202: Same question on picture vs field.

codec_specs.md
```diff
@@ -764,3 +784,70 @@ Codec Name: VobBtn Buttons
...
+
+Type Description: the `BlockAddIDExtraData` data is interpreted as `AVCDecoderConfigurationRecord` structure for the enhancement layer as defined in [@!ISO.14496-15.2014], as described in [Dolby Vision Streams File Format](ttps://www.dolby.com/us/en/technologies/dolby-vision/dolby-vision-bitstreams-within-the-iso-base-media-file-format-v2.1.2.pdf). This extension MUST NOT be used if `Codec ID` is not `V_MPEG4/ISO/AVC`.
+
+### dvcC
+
+Block type identifier: 0x64766343
+
+Block type name: Dolby Vision configuration
+
+Description: the `BlockAddIDExtraData` data is interpreted as `DOVIDecoderConfigurationRecord` structure as defined in [Dolby Vision Streams File Format](https://www.dolby.com/us/en/technologies/dolby-vision/dolby-vision-bitstreams-within-the-iso-base-media-file-format-v2.1.2.pdf), for Dolby Vision profiles less than and equal to 7.
```
robUx4, line 829: The link should be turned into a reference, otherwise the TXT version of the RFC will likely miss the link.

codec_specs.md
```diff
@@ -764,3 +784,70 @@ Codec Name: VobBtn Buttons
...
+
+Description: the `BlockAddIDExtraData` data is interpreted as `DOVIDecoderConfigurationRecord` structure as defined in [Dolby Vision Streams File Format](https://www.dolby.com/us/en/technologies/dolby-vision/dolby-vision-bitstreams-within-the-iso-base-media-file-format-v2.1.2.pdf), for Dolby Vision profiles greater than 7.
+
+### hvcE
+
+Block type identifier: 0x68766345
+
+Block type name: Dolby Vision enhancement-layer HEVC configuration
+
+Type Description: the `BlockAddIDExtraData` data is interpreted as `HEVCDecoderConfigurationRecord` structure for the enhancement layer as defined in [@!ISO.14496-15.2014], as described in [Dolby Vision Streams File Format](ttps://www.dolby.com/us/en/technologies/dolby-vision/dolby-vision-bitstreams-within-the-iso-base-media-file-format-v2.1.2.pdf). This extension MUST NOT be used if `Codec ID` is not `V_MPEGH/ISO/HEVC`.
```
robUx4, line 845: Same remark on the reference vs link.

robUx4: Does that mean we have the same `HEVCDecoderConfigurationRecord` twice in the `CodecPrivate` and in the `hvcE` ?

codec_specs.md
JeromeMartinez: > Is it pictures or fields ?

I used "picture" on purpose: AFAIU, in AVC and HEVC specs "pictures" refer to either a frame (progressive content) or to a field (interlaced content) or to 2 fields (progressive coded interlaced aka PPF / Picture-per-field?). Neutral wording.

Note that PPF is more or less classic in MPEG-2 Video and current Matroska spec is not more precise about that. currently "Elementary Stream (ES)" wording is used for MPEG-2 Video, would you prefer this wording?

> In other words interlaced content should group two fields together ?

I kept the wording neutral because AFAIK there is no limitation here: as in other formats (MP4, MXF...) all is possible.

In my opinion we should not limit here the possibilities, and the general specification would apply i.e. the element having the content has a timestamp which is the timestamp of the first stream element, as for other formats e.g. MPEG-2 Video.
If you prefer to limit, please indicate which limitation you want.

> Does the ISO spec refer to fields or pictures?

it uses "elementary stream" and "sample" and "picture" a bit randomly without defining them precisely. "elementary stream" seems to be more used, so we could use this wording.

JeromeMartinez:
@robUx4 I modified the PR accordingly, please review again.
Let me know if you prefer "elementary stream (ES)" instead of "picture".

robUx4:
`Elementary Stream` seems to be the whole stream rather than a picture. So if anything I would prefer `picture`.

I think it should be explained how to handled the interlacing a bit further if it can be one field, one picture (double field). At in which case should the `FlagInterlaced` of the Track should be set or not.

Just like for AV1 we should also explain how the fields of `AVCDecoderConfigurationRecord` should be mapped to the Matroska field. It's not a blocker for this PR but we should open an issue so we don't forget to do it.

codec_specs.md
JeromeMartinez: I forgot this one, `hvcE` contains `HEVCDecoderConfigurationRecord` of the **enhancement layer**, it is not a duplicate of the one in `CodecPrivate` (at least in its definition). It is described in the Dolby doc.
IIRC stuff like color info may is usually different i.e. nothing incompatible with base layer, but in practice whatever is the content of `hvcE`, it is mandatory by Dolby specs so should be in the track header.

JeromeMartinez:
> I think it should be explained how to handled the interlacing a bit further if it can be one field, one picture (double field). At in which case should the FlagInterlaced of the Track should be set or not.

`FlagInterlaced` (and other elements) setting is not indicated for others formats which can have similar issues e.g. MPEG-2 Video (ok, indicated "where block boundaries are still to be defined", so not the best example) or MPEG-4 Visual or ProRes (not possible to have 1 field per block, so easier, but still a lack of documentation), so this request seems generic to all formats supported in Matroska, and it is more an help for implementers than a codec mapping spec (see below) shouldn't it be a global request not blocking this PR rather than a blocker for this PR?

I added a commit for making explicit the idea behind the word `picture`, without more description e.g. about the mapping to other Matroska fields in order to avoid redundancy (more documentation about it would be in dedicated sections), would it be acceptable?

> Just like for AV1 we should also explain how the fields of AVCDecoderConfigurationRecord should be mapped to the Matroska field. It's not a blocker for this PR but we should open an issue so we don't forget to do it.

AV1 documentation is the only one with such verbose description, more an help for developers than what should be in a specification, actually the only one with its own .md file, not sure it is good to do that in the standard itself.

>  It's not a blocker for this PR but we should open an issue so we don't forget to do it.

I tried to write an issue for that, as well as for `FlagInterlaced`, but actually I don't see the exact goal: should I open an issue per Matroska field, an issue per supported format, a global issue for all Matroska fields and all supported formats? It changes a lot the amount of content in codec_specs.md (or other md files linked to codec specs).

***

In my opinion there is no change really to do to this PR as is (so it could be merged), but rather a global quality update on all codec mappings, which should have its own issue tickets.

JeromeMartinez:
nudge.
As AVC & HEVC codec mappings seems the only remaining issue, it is also possible to split the PR in 2 and merge the BlockAddiitonMapping part only, let me know if this split helps in merging the BlockAddiitonMapping part.
If Dolby Vision part is also an issue, the PR could also be limited to update BlockAddiitonMapping description + Opaque data + ITU T.35 metadata.

codec_specs.md
robUx4: OK for picture if that can represent a field in interlaced content.

codec_specs.md
robUx4: > data is interpreted as `HEVCDecoderConfigurationRecord` structure for the enhancement layer as defined in [@!ISO.14496-15.2014], as described in [@!DolbyVisionWithinIso.2020-02]

IMO the wording is not very clear. It doesn't show the difference with

> contains a `HEVCDecoderConfigurationRecord` structure as defined in [@!ISO.14496-15.2014].

I propose 

> data contains the enhancement layer of `HEVCDecoderConfigurationRecord` from [@!ISO.14496-15.2014]  as described in [@!DolbyVisionWithinIso.2020-02]

JeromeMartinez: @robUx4 your proposal is for me less clear, as it is not an "enhancement layer of `HEVCDecoderConfigurationRecord`", it is a `HEVCDecoderConfigurationRecord` structure, but not for the "classic" stream (it is for the Dobly Vision additions).

In order to be less controversial and for more difference with the `CodecPrivate` part, I modified the PR with the exact wording from the [Dolby spec section 2.3](https://www.dolby.com/us/en/technologies/dolby-vision/dolby-vision-bitstreams-within-the-iso-base-media-file-format-v2.1.2.pdf#_OPENTOPIC_TOC_PROCESSING_d100e6125) for the description of the data : "data is interpreted as the Dolby Vision enhancement-layer HEVC configuration as described in [@!DolbyVisionWithinIso.2020-02]", this way there is no more direct reference to `HEVCDecoderConfigurationRecord` (just an indirect one through the Dolby Vision spec) so no misinterpretation possible, and it may be more future proof (we describe nothing in the content here, the Dolby spec is the only source for the description of the content).

I understand this is the only change request, so please review again.

robUx4: LGTM. I was going to propose dropping the HEVCDecoderConfigurationRecord from the text to make it clearer but I wasn't sure.

dyspr0sium:
It may be of relevance to note that the original link for "Dolby Vision Streams Within the ISO Base Media File Format" now leads to a 404, and has been moved to [here](https://professional.dolby.com/siteassets/content-creation/dolby-vision-for-content-creators/dolby_vision_bitstreams_within_the_iso_base_media_file_format_dec2017.pdf).

codec_specs.md
efraim-il: > Please note, Dolby Vision is also possible (and is defined) for AVC streams, not only for HEVC.
> In case of AVC related IDs are dvcC and avcE . We probably should change dvcC to be allowed on both HEVC and AVC, and define avcE for DV on AVC.

HEVC Dolby Vision is not supported? I have seen a (MKV HEVC) movie online and it says Dolby Vision + HDR

JeromeMartinez: > HEVC Dolby Vision is not supported?

The text says the opposite, it is explicitly supported for both AVC and HEVC (other formats could be supported by just adding the corresponding 4CC for the enhanced layer configuration).

JeromeMartinez: The text says the opposite, it is explicitly supported for both AVC and HEVC (other formats could be supported by just adding the corresponding 4CC for the enhanced layer configuration).

JeromeMartinez: > HEVC Dolby Vision is not supported?

The text says the opposite, it is explicitly supported for both AVC and HEVC (other formats could be supported by just adding the corresponding 4CC for the enhanced layer configuration).

PR #392, opened 2020-05-26
Fixes for website migration

mbunkus:
Various modifications that target splitting off pure website related stuff into its own repository.


robUx4:
didn't this code move already ?

mbunkus:
Well, these commits actually remove the moved files from the specification repository. "Fixes…" was probably not the best name :)

mbunkus:
I'll push more changes onto this branch from time to time as I progress through the migration; probably rebase when necessary, too. No need to merge soon.

mbunkus:
The required changes are mostly complete now. I'll merge in a week or so if no one objects.

Makefile
```diff
@@ -34,7 +34,7 @@ ebml_matroska_elements4rfc.md: transforms/ebml_schema2markdown4rfc.xsl matroska_
 	xsltproc transforms/ebml_schema2markdown4rfc.xsl matroska_xsd.xml > $@
 
 $(OUTPUT_MATROSKA).md: index_matroska.md diagram.md matroska_schema_section_header.md ebml_matroska_elements4rfc.md ordering.md chapters.md attachments.md cues.md streaming.md menu.md notes.md rfc_backmatter_matroska.md
-	cat $^ | grep -v '^---' > $@
+	sed -e '/^---/,/^---/d' $^ > $@
```
robUx4, line 37: This is probably better but it's probably still not goo enough. Any text in between the two `---` at the top the file will be kept. Luckily for now we don't use this section at all.

robUx4, line 37: Good change !

index_matroska.md
```diff
@@ -64,7 +64,7 @@ Matroska is an open standards project. This means for personal use it is absolut
 
 # Status of this document
 
-This document is a work-in-progress specification defining the Matroska file format as part of the [IETF Cellar working group](https://datatracker.ietf.org/wg/cellar/charter/). But since it's quite complete it is used as a reference for the development of libmatroska. Legacy versions of the specification can be found [here](https://matroska.org/files/matroska.pdf) (PDF doc by Alexander Noé -- outdated).
+This document is a work-in-progress specification defining the Matroska file format as part of the [IETF Cellar working group](https://datatracker.ietf.org/wg/cellar/charter/). But since it's quite complete it is used as a reference for the development of libmatroska.
```
robUx4, line 67: I'd like to keep the file and the link somewhere.

notes.md
```diff
@@ -191,7 +192,7 @@ For separate tracks, Matroska needs to define exactly which track does what. `Tr
 
 The 3D support is still in infancy and may evolve to support more features.
 
-The StereoMode used to be part of Matroska v2 but it didn't meet the requirement for multiple tracks. There was also a bug in libmatroska prior to 0.9.0 that would save/read it as 0x53B9 instead of 0x53B8. `Matroska Readers` may support these legacy files by checking Matroska v2 or 0x53B9. The [older values](http://www.matroska.org/node/1/revisions/74/view#StereoMode) were 0: mono, 1: right eye, 2: left eye, 3: both eyes.
+The StereoMode used to be part of Matroska v2 but it didn't meet the requirement for multiple tracks. There was also a bug in libmatroska prior to 0.9.0 that would save/read it as 0x53B9 instead of 0x53B8. `Matroska Readers` may support these legacy files by checking Matroska v2 or 0x53B9. The older values were 0: mono, 1: right eye, 2: left eye, 3: both eyes.
```
robUx4, line 195: What was in the old values description ? Given it's not the RFC website but matroska.org we should keep documenting these old changes.

notes.md
```diff
@@ -192,7 +192,7 @@ For separate tracks, Matroska needs to define exactly which track does what. `Tr
...
```
robUx4, line 195: I was worried about this removal but it seems [even in 2010](https://web.archive.org/web/20120326131802/http://www.matroska.org/node/1/revisions/74/view) these values were the ones in use.

mbunkus: I had checked the link before removing it, and that node didn't exist anymore in Drupal. Hence the link was already dead, and I couldn't find an active page anywhere else in Drupal describing older values.

index_matroska.md
mbunkus: I can definitely add it back.

index_matroska.md
```diff
@@ -64,7 +64,7 @@ Matroska is an open standards project. This means for personal use it is absolut
 
 # Status of this document
 
-This document is a work-in-progress specification defining the Matroska file format as part of the [IETF Cellar working group](https://datatracker.ietf.org/wg/cellar/charter/). But since it's quite complete it is used as a reference for the development of libmatroska.
+This document is a work-in-progress specification defining the Matroska file format as part of the [IETF Cellar working group](https://datatracker.ietf.org/wg/cellar/charter/). But since it's quite complete it is used as a reference for the development of libmatroska. Legacy versions of the specification can be found [here](https://www.matroska.org/files/matroska_file_format_alexander_noé.pdf) (PDF doc by Alexander Noé -- outdated).
```
retokromer, line 67: Nit: why is “legacy versions” plural? To me the PDF is one working version.

index_matroska.md
mbunkus: The old specs PDF by Alexander Noé is up again, as is the link in this file. I've renamed the file to be bit a less vague about its contents.

index_matroska.md
mbunkus: Yeah. I just reversed the commit that got rid of the line, but I agree that singular's better.

mbunkus: Ugh, our generate-specs-process doesn't allow non-ASCII characters in hrefs.

retokromer: I guess, you can encode it with %nm (I have to check the numbers), like in the old times.

mbunkus: Nah, I just decided to use `…noe` instead.

mbunkus:
Soooo… all of the objections have been addressed as far as I can tell. Merging!

PR #395, opened 2020-07-03
Fix link and others

mbunkus:


mbunkus:
Sigh, the xml2rfc tool doesn't like having backticks inside links… At least it agrees with kramdown in that it doesn't actually convert the following into URLs:

```
`[text](URL)`
```

The generated HTML code:

```
<code>[text](URL)</code>
```

mbunkus:
My new variant simply removes the backticks from the `mvcC` links as `xml2rfc` doesn't like them either way (inside/outside).

PR #399, opened 2020-07-19
fix the EBML Schema namespace

robUx4:
It's now "urn:ietf:rfc:8794"

Otherwise `make check` doesn't work anymore.


dericed:
tested and works for me so merging

PR #398, opened 2020-07-19
Misc automated processing and cleaning

robUx4:
The next step will be to cut lines so it's easier to spot changes in the text. And when the RFC editors make changes we don't have to edit the whole file (see https://github.com/cellar-wg/ebml-specification/commit/6276603f754c51ce0ce77b4d2fd3db95093673f0).

PR #400, opened 2020-07-20
Make lines in markdown shorter

robUx4:
Before we start doing a lot of edits in the code it's better to have the format that will not change later. (from EBML experience).

It will also make comparing changes in Githu easier.

No text has been modified, only newlines added.

This PR sits on top of #399 and #398, so they should be merged before.


robUx4:
Rebased on master since the other PRs are merged.

robUx4:
Merging so we can move on with more real changes.

retokromer:
Oops, I missed this one, sorry!

PR #405, opened 2020-08-02
Clean/Fix Punctuation

robUx4:
Similar to RFC Editor reviews

### DIFF
--- a/codec_specs.md
+++ b/codec_specs.md
@@ -308,7 +308,7 @@ Codec ID: V_QUICKTIME
 
 Codec Name: Video taken from QuickTime(TM) files
 
-Description: Several codecs as stored in QuickTime, e.g. Sorenson or Cinepak.
+Description: Several codecs as stored in QuickTime, e.g., Sorenson or Cinepak.
 
 Initialization: The `Private Data` contains all additional data that is stored in the 'stsd' (sample description) atom
 in the QuickTime file **after** the mandatory video descriptor structure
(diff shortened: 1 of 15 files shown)


codec_specs.md
```diff
@@ -812,7 +812,7 @@ Codec ID: S_TEXT/SSA
 Codec Name: Subtitles Format
 
 Description: The [Script Info] and [V4 Styles] sections are stored in the codecprivate. Each event is stored in its own Block.
-For more information, see [the SSA/ASS section in Subtitles](#subtitles).
+For more information; see [the SSA/ASS section in Subtitles](#subtitles).
```
mbunkus, line 815: This change looks wrong. When in doubt if you should use a comma or a semicolon try using a full stop instead. If the result still makes sense, you use a semicolon. And that's not the case here: "For more information. See URL." is not OK.

I'm not going to add a comment for each of these "For more information, see…" changes, but they're all wrong IMHO.

index_matroska.md
```diff
@@ -78,7 +78,7 @@ the [IETF Cellar working group](https://datatracker.ietf.org/wg/cellar/charter/)
 But since it's quite complete it is used as a reference for the development of libmatroska.
 A legacy version of the specification can be found [here](https://www.matroska.org/files/matroska_file_format_alexander_noe.pdf) (PDF doc by Alexander Noé -- outdated).
 
-For a simplified diagram of the layout of a Matroska file, see the [Diagram page](diagram.md).
+For a simplified diagram of the layout of a Matroska file; see the [Diagram page](diagram.md).
```
mbunkus, line 81: Same argument as above: the "replace with full stop" test fails. A comma is appropriate, a semicolon isn't.

index_matroska.md
```diff
@@ -128,7 +128,7 @@ Specific terms are defined below:
 
 Matroska is a Document Type of EBML (Extensible Binary Meta Language).
 This specification is dependent on the [EBML Specification](https://github.com/Matroska-Org/ebml-specification/blob/master/specification.markdown).
-For an understanding of Matroska's EBML Schema, see in particular the sections of the EBML Specification covering
+For an understanding of Matroska's EBML Schema; see in particular the sections of the EBML Specification covering
```
mbunkus, line 131: Same as above; please revert.

robUx4:
Rebased and removed the bogus changes noted by @mbunkus

PR #417, opened 2020-08-25
aesthetic nit

retokromer:


robUx4:
I also don't like files missing a new line at the end.

PR #411, opened 2020-08-16
Clean text references in the EBML schema

robUx4:
It may make the rendered website look funny. Some additional scripting may be needed there.

PR #421, opened 2020-08-30
use legacy and RFC element references HTML anchors

robUx4:
So the internal links generated by the EBML Schema match the RFC and the website.
And we keep the old links alive for external references.

This is needed following #411

### DIFF
--- a/transforms/ebml_schema2spec_common.xsl
+++ b/transforms/ebml_schema2spec_common.xsl
@@ -110,7 +110,11 @@
       </tr>
     </xsl:if>
 
-    <tr id="{@name}">
+    <tr>
+      <xsl:attribute name="id">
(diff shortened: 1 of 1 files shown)


robUx4:
This is missing the (#anchored-element) transformation to HTML links

robUx4:
This XSLT file should probably be maintained outside of this repo.

retokromer:
>This XSLT file should probably be maintained outside of this repo.

Why? And where would you suggest to move it? (I am possibly missing something.)

robUx4:
The website is now generated here https://github.com/Matroska-Org/infrastructure
I'm not sure these files are of any use in the IETF/CELLAR repo.

retokromer:
Thank you for the answer. I was actually missing this and I second the idea to move it.

Issue #422, opened 2020-09-07
Consider providing a facility for integer-fraction timescales


rcombs:
It's pretty well-established that Matroska's poor timebase support is one of the format's worst properties. While it support very precise timestamps (down to the nanosecond), it's very inefficient to do so (and the resulting values still aren't exact for most input rates), so muxers tend to default to 1ms timestamps, which can lead to a variety of subtle issues, especially with high-packet-rate streams (e.g. audio) and VFR video content. Muxers _can_ choose rates that are closer to the time base of their inputs (or the packet rate of the content), but exactly how best to do so has always been unclear, and some of the possible options would lead to either worse player behavior, or timestamp drift. I'm proposing a format addition to remedy this.

The only actual normative change I propose is this: in addition to the classic nanosecond-denominator time scale, muxers could provide 2 additional integers, serving as a numerator and denominator time base value, which is required to round to the existing nanosecond-scaled value.

This should be paired with some advice for muxer implementations on how to make use of this feature. This depends on the properties of the input. For reference, here are some examples of the error produced by rounding a variety of common time bases to the nearest nanosecond, scaled by 3 hours (a reasonable target for the duration of a film):

```
nearest_ns(x) = round(x * 1,000,000,000) / 1,000,000,000
ceil_ns(x) = ceil(x * 1,000,000,000) / 1,000,000,000
floor_ns(x) = floor(x * 1,000,000,000) / 1,000,000,000
nearest_error(x) = 1 - (x / nearest_ns(x))
ceil_error(x) = 1 - (x / ceil_ns(x))
floor_error(x) = 1 - (x / floor_ns(x))
nearest_error_3h(x) = nearest_error(x) * 60 * 60 * 3
ceil_error_3h(x) = ceil_error(x) * 60 * 60 * 3
floor_error_3h(x) = floor_error(x) * 60 * 60 * 3
e(x) = nearest_error_3h(1 / x)
ce(x) = ceil_error_3h(1 / x)
fe(x) = floor_error_3h(1 / x)

# Integer video frame rates
e(24)        => 8.64e-5
e(25)        => 0
e(30)        => -0.0001
e(48)        => -0.0002
e(50)        => 0
e(60)        => 0.0002
e(120)       => -0.0004

# NTSC video frame rates
e(24/1.001)  => -8.6314e-5
e(30/1.001)  => 0.0001
e(48/1.001)  => 0.0002
e(60/1.001)  => -0.0002
[...]

PR #425, opened 2020-10-11
Proposal to add some rational number in timestamps

robUx4:
As discussed in #422 there are some cases where proper timing and backward compatibility is possible.

Wherever TimestampScale was used, if the new fraction is found, it should be used instead.

The fraction cannot just be in each track as the Cluster timestamp is also involved in the equation. So the precision has to be kept there as well.

If we had a [fraction type in EBML](https://github.com/cellar-wg/ebml-specification/issues/164) we could use that type to store the whole fraction.

### DIFF
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -88,11 +88,18 @@ The format depends on the ChapProcessCodecID used; see (#chapprocesscodecid-elem
     <documentation lang="en" purpose="definition">Timestamp scale in nanoseconds (1.000.000 means all timestamps in the Segment are expressed in milliseconds).</documentation>
     <extension cppname="TimecodeScale"/>
   </element>
-  <!-- <element name="TimestampScaleDenominator" parent="/Segment/Info" level="2" id="0x2AD7B2" type="uinteger" minOccurs="1" maxOccurs="1" minver="4" default="1000000000">
-    <documentation lang="en">Timestamp scale numerator; see <a href="https://www.matroska.org/technical/elements.html#TimestampScale">TimestampScale</a>.</documentation>
-TimestampScale When combined with <a href="https://www.matroska.org/technical/elements.html#TimestampScaleDenominator">TimestampScaleDenominator</a> the Timestamp scale is given by the fraction TimestampScale/TimestampScaleDenominator in seconds.-->
+
+  <element name="TimestampNumerator" path="\Segment\Tracks\TrackEntry\TimestampNumerator" id="0x2AF6A9" type="uinteger" range="not 0" maxOccurs="1" minver="4">
+    <documentation lang="en" purpose="definition">Timestamp numerator to apply instead of the TimestampScale, when the TimestampDenominator is present as well.</documentation>
+    <extension webm="0"/>
+  </element>
+  <element name="TimestampDenominator" path="\Segment\Tracks\TrackEntry\TimestampDenominator" id="0x2AF6AA" type="uinteger" range="not 0" maxOccurs="1" minver="4">
+    <documentation lang="en" purpose="definition">Timestamp denominator to apply instead of the TimestampScale, when the TimestampNumerator is present as well.</documentation>
+    <extension webm="0"/>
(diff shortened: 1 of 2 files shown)


robUx4:
There *could* be a fraction in each track as well that is not dependent on the other tracks clock. In the end we would use this fraction rather than the global one (which would mostly be for Cluster timestamps) given more range per Blocks for each track. But the value would not be close to what old parsers would understand. In that case that means using a Matroska v5 and marking it unreadable by v4 parsers.

I could live with that. After all MP4 CMAF didn't exist a few years back and now everyone switched to it.

ebml_matroska.xml
```diff
@@ -206,7 +213,7 @@ This information **SHOULD** always be referenced by a seek entry.</documentation
 The duration of DiscardPadding is not calculated in the duration of the TrackEntry and **SHOULD** be discarded during playback.</documentation>
     <extension webm="1"/>
   </element>
-  <element name="Slices" path="\Segment\Cluster\BlockGroup\Slices" id="0x8E" type="master" maxOccurs="1">
+  <element name="Slices" path="\Segment\Cluster\BlockGroup\Slices" id="0x8E" type="master" maxver="1" maxOccurs="1">
```
robUx4, line 216: Doesn't belong in this PR

ebml_matroska.xml
```diff
@@ -88,11 +88,18 @@ The format depends on the ChapProcessCodecID used; see (#chapprocesscodecid-elem
     <documentation lang="en" purpose="definition">Timestamp scale in nanoseconds (1.000.000 means all timestamps in the Segment are expressed in milliseconds).</documentation>
     <extension cppname="TimecodeScale"/>
   </element>
-  <!-- <element name="TimestampScaleDenominator" parent="/Segment/Info" level="2" id="0x2AD7B2" type="uinteger" minOccurs="1" maxOccurs="1" minver="4" default="1000000000">
-    <documentation lang="en">Timestamp scale numerator; see <a href="https://www.matroska.org/technical/elements.html#TimestampScale">TimestampScale</a>.</documentation>
-TimestampScale When combined with <a href="https://www.matroska.org/technical/elements.html#TimestampScaleDenominator">TimestampScaleDenominator</a> the Timestamp scale is given by the fraction TimestampScale/TimestampScaleDenominator in seconds.-->
+
+  <element name="TimestampNumerator" path="\Segment\Tracks\TrackEntry\TimestampNumerator" id="0x2AF6A9" type="uinteger" range="not 0" maxOccurs="1" minver="4">
+    <documentation lang="en" purpose="definition">Timestamp numerator to apply instead of the TimestampScale, when the TimestampDenominator is present as well.</documentation>
```
robUx4, line 93: If `TrackTimestampScale` is revived it should be mentioned its value MUST be the rounded value in nanoseconds of this fraction.

notes.md
```diff
@@ -453,6 +453,32 @@ Some general notes for a program:
...
+
+This formula remains but there is no rounding involved anymore:
+
+    (a + b) * c
+
+    a = `Block`'s Timestamp
+    b = `Cluster`'s Timestamp
+    c = `TimestampScale`
+
+For compatibility with older readers that don't understand these elements, the `TimestampScale` value *MUST*
```
robUx4, line 472: More info about the `TrackTimestampScale` should be added.

robUx4:
This might be an addition to #437 but less prioritary. A fraction for the global `TimestampScale` would give better result as the precision of the nanosecond is the most limiting factor. The floating `TrackTimestampScale` with a lot of precision even for odd fractions and guidelines on how to interpret the rounding (nearest, floor or top) should be enough.

robUx4:
Marked as v5.

It may also wait for [EBML rationale values](https://github.com/ietf-wg-cellar/ebml-specification/issues/164) to be supported.

hubblec4:
Maybe it is better to use the new Rational element type(which is not implemented yet in EBML) .

Nothing is more precise as a rational number.

PR #426, opened 2020-10-26
explain why the video framerate is informational only

robUx4:
Also removes the last usage of the `<strong>` HTML keyword, (see #380)

In another section we can probably explain that common framerates can be deduced from this value (especially the NSTC ones).

### DIFF
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -617,7 +617,7 @@ This value is similar in scope to the biCompression value of AVI's BITMAPINFOHEA
     <extension cppname="VideoGamma"/>
   </element>
   <element name="FrameRate" path="\Segment\Tracks\TrackEntry\Video\FrameRate" id="0x2383E3" type="float" minver="0" maxver="0" range="&gt; 0x0p+0" maxOccurs="1">
-    <documentation lang="en" purpose="definition">Number of frames per second. <strong>Informational</strong> only.</documentation>
+    <documentation lang="en" purpose="definition">Number of frames per second. This value is Informational only. It is intended for constant frame rate streams, and SHOULD NOT be used for a variable frame rate TrackEntry.</documentation>
     <extension webm="0"/>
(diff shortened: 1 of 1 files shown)


JeromeMartinez:
Maybe it is a change in the meaning of the element, but what about adding e.g. "intended for constant frame rate streams, SHOULD NOT be used in case of variable frame rate".

robUx4:
Fine with me. Although if we want a non informational value we should store a proper value, which would be a fraction.

JeromeMartinez:
Informational is enough, at least for that element (backward compatibility).

ebml_matroska.xml
```diff
@@ -617,7 +617,7 @@ This value is similar in scope to the biCompression value of AVI's BITMAPINFOHEA
     <extension cppname="VideoGamma"/>
   </element>
   <element name="FrameRate" path="\Segment\Tracks\TrackEntry\Video\FrameRate" id="0x2383E3" type="float" minver="0" maxver="0" range="&gt; 0x0p+0" maxOccurs="1">
-    <documentation lang="en" purpose="definition">Number of frames per second. This value is Informational only as the Segment may have a varying frame rate.</documentation>
+    <documentation lang="en" purpose="definition">Number of frames per second. This value is Informational only. It is intended for constant frame rate streams, and SHOULD NOT be used for variable frame rate Segments.</documentation>
```
JeromeMartinez, line 620: Segments --> Tracks (you may have a CFR track beside a VFR track in a segment, and the element could still be used on the CFR track).

JeromeMartinez (changes requested):

ebml_matroska.xml
robUx4: Indeed. Fixed.

retokromer (approved):

JeromeMartinez (approved):

dericed (approved):

PR #431, opened 2020-10-27
fix Makefile so that it makes all targets by default

robUx4:


retokromer (approved):

dericed (approved):
thx, I noticed that when marking the last release.

PR #434, opened 2020-10-27
Use references for links in the tags documentation

robUx4:
Also remove some text that is only meant for the website, not the specs.


dericed (approved):
Thanks for this. I see it was a lot of work. :)

retokromer (approved):

PR #437, opened 2020-11-01
allow TrackTimestampScale in Matroska v4

robUx4:
Following the investigation in #422 it seems like a good tool to achieve sample accurate timestamp in many (more) cases.

In posts starting from https://github.com/cellar-wg/matroska-specification/issues/422#issuecomment-711123080

### DIFF
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -312,9 +312,9 @@ If set to 0, the reference pseudo-cache system is not used.</documentation>
 see (#defaultdecodedfieldduration) for more information</documentation>
     <extension type="libmatroska" cppname="TrackDefaultDecodedFieldDuration"/>
   </element>
-  <element name="TrackTimestampScale" path="\Segment\Tracks\TrackEntry\TrackTimestampScale" id="0x23314F" type="float" maxver="3" range="&gt; 0x0p+0" default="0x1p+0" minOccurs="1" maxOccurs="1">
-    <documentation lang="en" purpose="definition">DEPRECATED, DO NOT USE. The scale to apply on this track to work at normal speed in relation with other tracks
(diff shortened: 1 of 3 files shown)


retokromer (approved):

robUx4:
@mbunkus any opinion on that ? It seems libmatroska is not ready for that move but it's possible to [work around it](https://github.com/Matroska-Org/libmatroska/pull/52) without breaking everything. That has a direct impact on support for this feature in VLC as well.

I have patches for libmatroska, VLC and libavformat (read and write). So it's at least possible to use it (with sample "accurate" timestamps, see #425). It should not be used by default because most existing readers, if not all, will not handle the value properly. But it can gradually be added. After all it's something that should have been supported since the beginning. And it is there in the specs up to version 3.

robUx4:
For the record, the feature is now [available in libavformat](https://github.com/FFmpeg/FFmpeg/commit/3a2b786db0fe52ccdcb4a3276758c8ddd4887ce6#diff-29f77e0fca7fc65e99d135bb5ece0f0e523898046f62ee5aee2a758ea3b5092a) for reading. For writing it depends on codec packing size(s) so we need a list of known packing sizes (in my test to generate files I used 8 by default which should give long enough Clusters in most cases).

robUx4:
This is necessary but not sufficient to do sample accurate timestamps. For that we also need proper #439 documentation.

robUx4:
Turned into a draft until #521 is merged.

This now contains the formula to use `TrackTimestampScale` to get the Track ticks matching the sample number of each track.

robUx4:
The consensus on #422 is that reusing the element with some extra trick will likely not bring real world support. So it should be done with new elements instead.

Therefore I'm closing this, but parts could be reused to describe properly these elements.

PR #438, opened 2020-11-01
mark recurring elements with a maxOccurs or 1

### LINKED ISSUE #436: Why is the Info element multiple?
The top-level element 'Info' is a multiple one, and so I can/could set twice or more a TimestampScale element with different values.
Also the SegementUID can then be set multiple times.

robUx4:
In the past there was some confusion between elements that could be repeated and elements that could be multiple.

The EBML specs makes it clear that recurring elements are found multiple times (of the same copy) and that maxOccurs is ignored in that case.

Adding the `maxOccurs` value makes it clearer that there cannot be multiple variants of this element, only exact copies.

That should fix #436


retokromer (approved):

hubblec4:
Sounds good.

Issue #322, opened 2019-05-28
ChapterTimeEnd should not be marked as informational


robUx4:
See the table in `EditionFlagOrdered` that says it might be.

robUx4:
Needed for #318

hubblec4:
When we allow a ChapterTimeEnd in Basic chapter(only ChapterTimeStart is used) than we must describe a special cases. It will be to complicated.

ChapterContentType is a new feature an should be available in ordered chapters only.

Basic chapters should stay basic/simple.

robUx4:
IMO the `ChapterContentType` from #318 doesn't need ordered chapters. Ordered chapters should only be used for reordering of content or multiple editions on the same file. Most players don't support the feature and it's not expect they do in the foreseeable future.

The `EditionFlagOrdered` table, which is a bit confusing to read, doesn't mean that `ChapterTimeEnd` is forbidden or "informational" (that term is not written). So IMO we don't need any change regarding the `ChapterTimeEnd`.

robUx4:
Fixed by #440 which does mention ChapterTimeEnd as ordered chapters specific anymore.

PR #440, opened 2020-11-29
Simplify EditionFlagOrdered table

robUx4:
We don't need to tell what is allowed in Simple Chapters nor mention elements that are allowed in both (nothing that is allowed in Simple Chapters is forbidden in Orderer Chapters).

What we need is to know which elements don't need to be handled if ordered chapters are not supported (ie most players). `ChapterTimeEnd` is not one of them (cf #322).

Plus less backticking (there is ton of cleaning to be done there).

### DIFF
--- a/chapters.md
+++ b/chapters.md
@@ -72,9 +72,9 @@ play those Chapters in their stored order from the timestamp marked in the
 If the `EditionFlagOrdered Flag` is set to `false`, `Simple Chapters` are used and
 only the `ChapterTimeStart` of a `Chapter` is used as chapter mark to jump to the
 predefined point in the timeline. With `Simple Chapters`, a `Matroska Player` **MUST**
-ignore certain `Chapter Elements`. All these elements are now informational only.
+ignore certain Chapter elements. All these elements are now informational only.
 
-The following list shows the different usage of `Chapter Elements` between an ordered
+The following list shows the different usage of Chapter elements between an ordered
 and non-ordered `Edition`.
 
 | Chapter elements / ordered Edition | False | True |
--- a/chapters.md
+++ b/chapters.md
@@ -88,9 +88,6 @@ The following list shows the different Chapter elements only found in `Ordered C
 | TrackEntry/TrackTranslate             |
 Table: elements only found in ordered chapters{#orderedOnly}
 
-These other `Elements` belong to the Matroska DVD menu system and are only used
-when the `ChapProcessCodecID Element` is set to 1.
-
 ##### Ordered-Edition and Matroska Segment-Linking
 
 - Hard Linking: `Ordered-Chapters` supersedes the `Hard Linking`.
--- a/chapters.md
+++ b/chapters.md
@@ -78,7 +78,6 @@ The following list shows the different Chapter elements only found in `Ordered C
 
 | Ordered Chapter elements              |
 |:--------------------------------------|
-| ChapterAtom/ChapterTimeEnd            |
 | ChapterAtom/ChapterSegmentUID         |
 | ChapterAtom/ChapterSegmentEditionUID  |
 | ChapterAtom/ChapterTrack              |
--- a/chapters.md
+++ b/chapters.md
@@ -79,17 +79,10 @@ and non-ordered `Edition`.
 
 | Chapter elements / ordered Edition | False | True |
 |:-----------------------------------|:-----:|:----:|
-| ChapterUID                         |   X   |  X   |
-| ChapterStringUID                   |   X   |  X   |
-| ChapterTimeStart                   |   X   |  X   |
 | ChapterTimeEnd                     |   -   |  X   |
-| ChapterFlagHidden                  |   X   |  X   |
-| ChapterFlagEnabled                 |   X   |  X   |
 | ChapterSegmentUID                  |   -   |  X   |
 | ChapterSegmentEditionUID           |   -   |  X   |
-| ChapterPhysicalEquiv               |   X   |  X   |
 | ChapterTrack                       |   -   |  X   |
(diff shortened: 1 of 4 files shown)


hubblec4 (commented):
Why is the `ChapterTimeEnd` element not in the list anymore?

In which case is this element used in `Simple Chapters`?
The `Simple Chapters` definition currently says the `ChapterTimeStart` element is only used.

robUx4:
Chapters can be used to tag particular parts of a file, regardless of how it's played. To do that we need a start and an end, even with Simple Chapters.

hubblec4:
Mmmh... when an end time is allowed in Simple Chapters then we can also allow the ChapterTrack element and also the ChapterSegmentUID element.

When you want to tag a particular part of a file then you have to use ordered chapters.

robUx4:
> When you want to tag a particular part of a file then you have to use ordered chapters.

Not at all. Having a whole audio album in one file was always one of the goal of Matroska. You RIP your CD in one file and then have each track with a Chapter. CDs often have gaps between tracks (often 2s). You can also add those when mastering your own CDs in software. Those gaps should not be tagged (unless you tag the whole album). They should be played but only when you play your album continuously. If you want to play a track you should not be forced to handle that gap either. And ordered chapters are never involved in any of these. Ordered chapters are only useful if you want to reorder content or add some interactive things (which ends up also changing the order).

hubblec4:
I am not yet entirely clear how this should work.

For example we have a CD with 3 songs and 2 gaps. 
Song 1 -> 2min
Gap 1 -> 2sec
Song 2 -> 3min
Gap 2 -> 2sec
Song 3 -> 4min

The CD.mkv has now a duration of 9 min and 4 sec and the smart CD2mkv tool generates the following chapters
Edition 1
Chapter 1 - Song 1 - start time 0sec
Chapter 2 - Song 2 - start time 2min 2sec
Chapter 3 - Song 3 - start time 5min 4sec

> Those gaps should not be tagged (unless you tag the whole album). They should be played but only when you play your album continuously

When we start the CD.mkv it plays the entire file, all songs and all gaps.

> If you want to play a track you should not be forced to handle that gap either.

With Chapter 2 or 3 you can always play the song 2 or 3 directly without the gap.
If you choose song/chapter 2 then it plays the gap 2 before it plays the song 3.


What should happens if the chapters have an end time?
Edition 1
Chapter 1 - Song 1 - start time 0sec - end time 2min
Chapter 2 - Song 2 - start time 2min 2sec - end time 5min 2sec
Chapter 3 - Song 3 - start time 5min 4sec - end time 9min 4sec

Currently there is no player/splitter which supports the end time in `Simple Chapters`.
Lets say VLC can do it :-)
We start our CD.mkv and again the entire content of the file is played.

What happens if song 1 ends? Skip the gap? But we want to play the gap for continuously playback of the album.
How should a player/splitter decide if the gap is played or skipped?


> Ordered chapters are only useful if you want to reorder content or add some interactive things (which ends up also changing the order).

Of course ordered chapters are used mostly to reorder the content. But there are features which don't reorder the content like the `ChapterTrack` element. As well the `ChapterProcess` with a command can be used which don't reorder the content.

The word "ordered" is a bit to specific it suggest you can only reorder the content.
For me is it an alternate playback feature to manipulate the playback and the content but without changing the original content.

robUx4:
> When we start the CD.mkv it plays the entire file, all songs and all gaps.

Correct.

>  Lets say VLC can do it :-)

It can. Every file that has chapters creates a virtual chapter which helps keep track of which chapter is being played, regardless of ordered chapter or not. It helps show in the UI which is the current chapter.

In the case of the CD in the gaps (I think they are called lead-in and lead-out) should show that it's not playing a chapter (CD track). This is what happens on old CD players. Some would display 0:00 before the next track would play. Some would even display -0:02 -0:01 before starting the next track.

Knowing where the expected content ends also helps extracting each track individually without the extra gaps (but some purists may want to keep it). An audio player aware that the CD.mkv contains multilple track can also tell where they end better than if the end was not marked. Otherwise such a player will play 2s after each track (or 10s if the CD was mastered that way). This is usually not what the user expects.

hubblec4:
> which helps keep track of which chapter is being played
> It helps show in the UI which is the current chapter.
> can also tell where they end better than if the end was not marked
> also helps extracting each track

Sounds for me that the end time is informational only.
The end time is not really used to create a duration of the chapter.

> Otherwise such a player will play 2s after each track (or 10s if the CD was mastered that way).
This will always happens.
Like I asked, how will the player knows/decide what is to do. Play the gap or skip it.

> Lets say VLC can do it :-)   .....    It can.
It would be nice to have a working sample, could you create some one?

robUx4:
> > which helps keep track of which chapter is being played
> > It helps show in the UI which is the current chapter.
> > can also tell where they end better than if the end was not marked
> > also helps extracting each track
> 
> Sounds for me that the end time is informational only.
> The end time is not really used to create a duration of the chapter.

Yes. As I explained if you "rip" one track you don't want to keep the lead-in silence. The ChapterTimeEnd helps with that. Also this notion that ChapterTimeEnd is informational never existed. There is no reason to add it. That doesn't mean it's mandatory. But it has many use.

> > Otherwise such a player will play 2s after each track (or 10s if the CD was mastered that way).
> 
> This will always happens.
> Like I asked, how will the player knows/decide what is to do. Play the gap or skip it.

Right now the information that a file is "CD-like" doesn't exist. That's something we can think of.

> > Lets say VLC can do it :-)   .....    It can.
> 
> It would be nice to have a working sample, could you create some one?

I haven't touched such (non free) CD software in 10 or 15 years so not really. But if you find good software, sure.

hubblec4:
OK.
I would say you can merge this PR.

PR #446, opened 2020-12-27
Clarification about when CodecPrivate is mandatory

JeromeMartinez:
More wording about when CodecPrivate is mandatory, following https://mailarchive.ietf.org/arch/msg/cellar/8vUqFFDT-Dqu1kc2ZwE_qOBvH8M/.

Also comply to the sentence:

> Each encoding supported for storage in Matroska **MUST** have a defined Initialization. 

And use the same wording `Private Data` everywhere.

I am not 100% sure about the `Initialization: none` (when nothing was previously written, I consider that there is no init), please double-check.

### DIFF
--- a/codec_specs.md
+++ b/codec_specs.md
@@ -71,8 +71,9 @@ which **MUST** be stored within the `CodecPrivate Element`. When the Initializat
 within a track, then that updated Initialization data **MUST** be written into the `CodecState Element`
 of the first `Cluster` to require it. If the encoding does not require any form of Initialization,
 then `none` **MUST** be used to define the Initialization and the `CodecPrivate Element`
-**SHOULD NOT** be written and **MUST** be ignored. Data that is defined Initialization to be
-stored in the `CodecPrivate Element` is known as `Private Data`.
+**SHOULD NOT** be written and **MUST** be ignored. If the encoding does require any form of Initialization,
+then `CodecPrivate Element` **MUST** be written and **MUST** be provided to the decoder.
+The Initialization data to be stored in the `CodecPrivate Element` is referred to as `Private Data`.
 
 ### Codec BlockAdditions
 
@@ -332,8 +333,8 @@ Codec Name: Theora
 
 Initialization: The `Private Data` contains the first three Theora packets in order. The lengths of the packets precedes them. The actual layout is:
 
-* Byte 1: number of distinct packets `#p` minus one inside the CodecPrivate block. This **MUST** be "2" for current (as of 2016-07-08) Theora headers.
-* Bytes 2..n: lengths of the first `#p` packets, coded in Xiph-style lacing. The length of the last packet is the length of the CodecPrivate block minus the lengths coded in these bytes minus one.
+* Byte 1: number of distinct packets `#p` minus one inside the `Private Data`. This **MUST** be "2" for current (as of 2016-07-08) Theora headers.
+* Bytes 2..n: lengths of the first `#p` packets, coded in Xiph-style lacing. The length of the last packet is the length of the `Private Data` minus the lengths coded in these bytes minus one.
 * Bytes n+1..: The Theora identification header, followed by the commend header followed by the codec setup header. Those are described in the [Theora specs](http://www.theora.org/doc/Theora.pdf).
 
 ### V_PRORES
@@ -389,7 +390,7 @@ Description: FFV1 is a lossless intra-frame video encoding format designed to ef
 Compared to uncompressed video, FFV1 offers storage compression, frame fixity, and self-description,
 which makes FFV1 useful as a preservation or intermediate video format. [Draft FFV1 Specification](https://datatracker.ietf.org/doc/draft-ietf-cellar-ffv1/)
 
-Initialization: For FFV1 versions 0 or 1, `Private Data` **SHOULD NOT** be written. For FFV1 version 3 or greater, the `Private Data` **MUST** contain the FFV1 Configuration Record structure, as defined in https://tools.ietf.org/html/draft-ietf-cellar-ffv1-04#section-4.2, and no other data.
+Initialization: For FFV1 versions 0 or 1, none. For FFV1 version 3 or greater, the `Private Data` contains the FFV1 Configuration Record structure, as defined in https://tools.ietf.org/html/draft-ietf-cellar-ffv1-04#section-4.2, and no other data.
 
 ## Audio Codec Mappings
 
@@ -471,8 +472,10 @@ Codec ID: A_AC3
 
 Codec Name: (Dolby™) AC3
 
-Description: BSID <= 8 !! The private data is void ??? Corresponding ACM wFormatTag : 0x2000 ; channel number have
+Description: For BSID <= 8. Corresponding ACM wFormatTag : 0x2000 ; channel number have
 to be read from the corresponding audio element
+
+Initialization: none
 
 ### A_AC3/BSID9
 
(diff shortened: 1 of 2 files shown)


ebml_matroska.xml
```diff
@@ -398,7 +398,7 @@ If this Element is used, then any Language Elements used in the same TrackEntry
 see [@!I-D.ietf-cellar-codec] for more info.</documentation>
   </element>
   <element name="CodecPrivate" path="\Segment\Tracks\TrackEntry\CodecPrivate" id="0x63A2" type="binary" maxOccurs="1">
-    <documentation lang="en" purpose="definition">Private data only known to the codec.</documentation>
+    <documentation lang="en" purpose="definition">Private data only known to the codec. This element **MUST NOT** be present if the codec mapping specification defines no initialization or an initialization `none`, else **MUST NOT** be present.</documentation>
```
mbunkus, line 401: The "else" case is wrong :grin:

codec_specs.md
```diff
@@ -872,13 +888,15 @@ VobSubs consist of two files, the .idx containing information, and the .sub, con
...
 and is discarded. All remaining lines but the ones containing timestamps and file positions
-are put into the `CodecPrivate` element.
+are put into the `Private Data`.
 
 For each line containing the timestamp and file position data is read from the appropriate
 position in the .sub file. This data consists of a MPEG program stream which in turn
 contains SPU packets. The MPEG program stream data is discarded, and each SPU packet
 is put into one Matroska frame.
 
+Initialization: none
```
mbunkus, line 898: This isn't correct; the original text states which lines must be placed in `CodecPrivate`.

mbunkus (changes requested):

JeromeMartinez:
@mbunkus please review again.

codec_specs.md
```diff
@@ -815,24 +819,30 @@ Codec Name: UTF-8 Plain Text
...
 
 Codec ID: S_TEXT/ASS
 
 Codec Name: Advanced Subtitles Format
 
-Description: The [Script Info] and [V4 Styles] sections are stored in the codecprivate. Each event is stored in its own Block.
+Description: Each event is stored in its own Block.
 For more information, see (#ssa-ass-subtitles) on SSA/ASS.
 
+Initialization: The `Private Data` contains the [Script Info] and [V4 Styles] sections.
```
mbunkus, line 844: Unfortunately, the original text was wrong. For `S_TEXT/ASS` the two sections are `[Script Info]` (this one was correct) and `[V4+ Styles]` (this one was incorrect; it's `[V4 Styles]` without the `+` for SSA but `[V4+ Styles]` with the `+` for ASS). So let's correct it while we touch this part anyway.

Thanks!

codec_specs.md
```diff
@@ -71,7 +71,9 @@ which **MUST** be stored within the `CodecPrivate Element`. When the Initializat
 within a track, then that updated Initialization data **MUST** be written into the `CodecState Element`
 of the first `Cluster` to require it. If the encoding does not require any form of Initialization,
 then `none` **MUST** be used to define the Initialization and the `CodecPrivate Element`
-**SHOULD NOT** be written and **MUST** be ignored. Data that is defined Initialization to be
+**SHOULD NOT** be written and **MUST** be ignored. If the encoding does require any form of Initialization,
+then `CodecPrivate Element` **MUST** be written and **MUST** be provided to the decoder.
+Data that is defined Initialization to be
```
mbunkus, line 76: I know that this sentence wasn't changed, but to be honest, I didn't like the original anyway. Maybe we can change it right now, too. On the other hand changing verbiage might be better handled as a separate MR.

If we wanted to change it now, the following might be better:

> The Initialization data to be stored in the `CodecPrivate Element` is referred to as `Private Data`.

Again, I'd be fine accepting this MR without this particular change.

mbunkus (changes requested):
Thanks. I just noticed two more things, one of which is factually wrong in the original already, the other being a question of style.

I'd be fine having both of those issues resolved in separate MRs (especially the one about verbiage as there are a lot more pieces of text that would benefit from similar scrutiny) and accepting this one as-is. Thoughts?

codec_specs.md
JeromeMartinez: > If we wanted to change it now, the following might be better

Obvious bad typos, I fixed them in the same commit. Please review again.

ebml_matroska.xml
```diff
@@ -398,7 +398,7 @@ If this Element is used, then any Language Elements used in the same TrackEntry
 see [@!I-D.ietf-cellar-codec] for more info.</documentation>
   </element>
   <element name="CodecPrivate" path="\Segment\Tracks\TrackEntry\CodecPrivate" id="0x63A2" type="binary" maxOccurs="1">
-    <documentation lang="en" purpose="definition">Private data only known to the codec.</documentation>
+    <documentation lang="en" purpose="definition">Private data only known to the codec. This element **MUST NOT** be present if the codec mapping specification defines no initialization or an initialization `none`, else **MUST** be present.</documentation>
```
robUx4, line 401: IMO present and with a size of 0 would also mean 'none'. For example in VP9 you can have one to give the 10/12 bits profiles. But there may be none. IMO writing an empty field is better than no field at all in this case. It's clear that it's the default profile. (there are tons of 10 bits VP9 files with no CodecPrivate and there's no way to tell if they are 8 or 10 bits).

This is also in line with
> If the encoding does require any form of Initialization then `CodecPrivate Element` **MUST** be written

`V_PRORES` would probably benefit from that as well.

codec_specs.md
```diff
@@ -71,8 +71,9 @@ which **MUST** be stored within the `CodecPrivate Element`. When the Initializat
 within a track, then that updated Initialization data **MUST** be written into the `CodecState Element`
 of the first `Cluster` to require it. If the encoding does not require any form of Initialization,
 then `none` **MUST** be used to define the Initialization and the `CodecPrivate Element`
-**SHOULD NOT** be written and **MUST** be ignored. Data that is defined Initialization to be
-stored in the `CodecPrivate Element` is known as `Private Data`.
+**SHOULD NOT** be written and **MUST** be ignored. If the encoding does require any form of Initialization,
+then `CodecPrivate Element` **MUST** be written and **MUST** be provided to the decoder.
```
robUx4, line 75: I don't think we need to mention the decoder here. Or rather separately in a different sentence.

codec_specs.md
```diff
@@ -869,16 +884,16 @@ Codec Name: VobSub subtitles
...
-and is discarded. All remaining lines but the ones containing timestamps and file positions
-are put into the `CodecPrivate` element.
 
 For each line containing the timestamp and file position data is read from the appropriate
 position in the .sub file. This data consists of a MPEG program stream which in turn
 contains SPU packets. The MPEG program stream data is discarded, and each SPU packet
 is put into one Matroska frame.
 
+Initialization: The .idx file is stripped of all empty lines, of all comments and of lines beginning with `alt:` or `langidx:`.
+The line beginning with `id:` **SHOULD** be transformed into the appropriate Matroska track language element
```
robUx4, line 894: The SHOULD should probably be a MUST and mention that the line is discarded. But the wording for this whole codec is questionable and could be fixed separately.

robUx4 (changes requested):
Mostly OK apart from 'none' which IMO should also accept the empty value (meaning default value in Matroska). Especially for codecs that can have 'none' and some actual value. Or maybe they should NOT have 'none' but define the default value instead. (which would also save a tiny bit of space). 
It might be tricky to define a default value though. It could be a large binary blob (Vorbis).

ebml_matroska.xml
robUx4: poke

IMO this addition is not good, it's up to the codec to tell what it means if it's present or not.

codec_specs.md
robUx4: poke

If we leave *SHOULD* we have to explain in which case do use it and in which case we don't.

codec_specs.md
```diff
@@ -869,16 +884,16 @@ Codec Name: VobSub subtitles
...
+and is discarded. The `Private Data` contains all remaining lines but the ones containing timestamps and file positions.
```
robUx4, line 895: Should be `CodecPrivate` not `Private Data`.

robUx4 (changes requested):

robUx4:
After a rebase the remaining parts is

> If the decoding does require any form of Initialization,
then `CodecPrivate Element` **MUST** be written and **MUST** be provided to the decoder.

However some codec, like FFV1 or VP9, have optional Initialization data. But maybe they are not "hard requirements" as the text mentions. So it should MUST be provided to the decoder (worth mentioning as it's not done anywhere yet), but it may not always be there depending on how the Initiliazing section is worded.

For the rest, the Initialization attribute was added to all codec definitions.

robUx4:
We already have this sentence:

> The Initialization **MUST** describe the storage of data necessary to initialize the decoder,
which **MUST** be stored within the `CodecPrivate` element.

So IMO it's redundant to require that it's written.

robUx4:
Closing in favor of #996

PR #447, opened 2021-01-01
Improve track autoselection capabilities and documentation

### LINKED ISSUE #374: clarify use of forced flag
rebase #158, and adjust text to reflect consensus

rcombs:
This is a (long-delayed) successor to #158 and is intended to fix #374; there's a substantial amount of discussion in that issue and the previous PR.

This aims to solve a few problems:
- The lack of clarity around the meanings of the "default" and "forced" flags, redefining both to have more useful meanings:
  - The "default" flag previously had no particular value; its function was indistinguishable from simply making the "default" track the first of its type in the file (and, in fact, most files that set a single track as "default" seem to set it on the first track of a given type)
  - The "forced" flag previously was documented to have behavior that no player implemented, and that no authoring software supported. Instead, many files have used it in the way described in this PR, so there doesn't seem to be any reason not to make it official. Some players seem to have used the forced flag as a sort of a "super-default" (i.e. a forced track is preferred over a default track, and a default track over a no-flags track), which isn't especially useful either.
- The lack of any structured way to indicate tracks providing accessibility services
- The lack of a way to specify explicit associations between audio and subtitle tracks
- The lack of clear documentation on how to implement track autoselection

The algorithm described in the last commit is very similar to the ones implemented in mpv and Plex. I'd be curious to see additional feedback from other player developers.

### DIFF
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -298,13 +298,13 @@ This **SHOULD** be kept the same when making a direct stream copy of the Track t
     <extension cppname="TrackFlagEnabled"/>
   </element>
   <element name="FlagDefault" path="\Segment\Tracks\TrackEntry\FlagDefault" id="0x88" type="uinteger" range="0-1" default="1" minOccurs="1" maxOccurs="1">
-    <documentation lang="en" purpose="definition">Set if that track (audio, video or subs) **SHOULD** be active if no language found matches the user preference. (1 bit)</documentation>
+    <documentation lang="en" purpose="definition">Set if that track (audio, video or subs) **SHOULD** be eligible for automatic selection by the player. (1 bit)</documentation>
     <extension cppname="TrackFlagDefault"/>
   </element>
   <element name="FlagForced" path="\Segment\Tracks\TrackEntry\FlagForced" id="0x55AA" type="uinteger" range="0-1" default="0" minOccurs="1" maxOccurs="1">
-    <documentation lang="en" purpose="definition">Set if that track **MUST** be active during playback. There can be many forced track for a kind (audio, video or subs),
-the player **SHOULD** select the one which language matches the user preference or the default + forced track.
-Overlay **MAY** happen between a forced and non-forced track of the same kind. (1 bit)</documentation>
+    <documentation lang="en" purpose="definition">Applies only to subtitles. Set if that track **SHOULD** be eligible for automatic selection by the player if it matches the user's language preference,
+even if the user's preferences would normally not enable subtitles with the selected audio track;
+this can be used for tracks containing only translations of foreign-language audio or onscreen text. (1 bit)</documentation>
     <extension cppname="TrackFlagForced"/>
   </element>
   <element name="FlagLacing" path="\Segment\Tracks\TrackEntry\FlagLacing" id="0x9C" type="uinteger" range="0-1" default="1" minOccurs="1" maxOccurs="1">
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -307,6 +307,18 @@ even if the user's preferences would normally not enable subtitles with the sele
 this can be used for tracks containing only translations of foreign-language audio or onscreen text. (1 bit)</documentation>
     <extension cppname="TrackFlagForced"/>
   </element>
+  <element name="FlagHearingImpaired" path="\Segment\Tracks\TrackEntry\FlagHearingImpaired" id="0x55AB" type="uinteger" range="0-1" default="0" maxOccurs="1">
+    <documentation lang="en" purpose="definition">Set if that track is suitable for users with hearing impairments. (1 bit)</documentation>
+    <extension webm="0"/>
+  </element>
+  <element name="FlagVisualImpaired" path="\Segment\Tracks\TrackEntry\FlagVisualImpaired" id="0x55AC" type="uinteger" range="0-1" default="0" maxOccurs="1">
+    <documentation lang="en" purpose="definition">Set if that track is suitable for users with visual impairments. (1 bit)</documentation>
+    <extension webm="0"/>
+  </element>
+  <element name="FlagTextDescriptions" path="\Segment\Tracks\TrackEntry\FlagTextDescriptions" id="0x55AD" type="uinteger" range="0-1" default="0" maxOccurs="1">
+    <documentation lang="en" purpose="definition">Set if that track contains textual descriptions of video content. (1 bit)</documentation>
+    <extension webm="0"/>
+  </element>
   <element name="FlagLacing" path="\Segment\Tracks\TrackEntry\FlagLacing" id="0x9C" type="uinteger" range="0-1" default="1" minOccurs="1" maxOccurs="1">
     <documentation lang="en" purpose="definition">Set if the track **MAY** contain blocks using lacing. (1 bit)</documentation>
     <extension cppname="TrackFlagLacing"/>
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -298,13 +298,14 @@ This **SHOULD** be kept the same when making a direct stream copy of the Track t
     <extension cppname="TrackFlagEnabled"/>
   </element>
   <element name="FlagDefault" path="\Segment\Tracks\TrackEntry\FlagDefault" id="0x88" type="uinteger" range="0-1" default="1" minOccurs="1" maxOccurs="1">
-    <documentation lang="en" purpose="definition">Set if that track (audio, video or subs) **SHOULD** be eligible for automatic selection by the player. (1 bit)</documentation>
+    <documentation lang="en" purpose="definition">Set if that track (audio, video or subs) **SHOULD** be eligible for automatic selection by the player; see (#default-track-selection) for more details. (1 bit)</documentation>
     <extension cppname="TrackFlagDefault"/>
   </element>
   <element name="FlagForced" path="\Segment\Tracks\TrackEntry\FlagForced" id="0x55AA" type="uinteger" range="0-1" default="0" minOccurs="1" maxOccurs="1">
     <documentation lang="en" purpose="definition">Applies only to subtitles. Set if that track **SHOULD** be eligible for automatic selection by the player if it matches the user's language preference,
 even if the user's preferences would normally not enable subtitles with the selected audio track;
-this can be used for tracks containing only translations of foreign-language audio or onscreen text. (1 bit)</documentation>
+this can be used for tracks containing only translations of foreign-language audio or onscreen text.
+See (#default-track-selection) for more details. (1 bit)</documentation>
     <extension cppname="TrackFlagForced"/>
   </element>
   <element name="FlagHearingImpaired" path="\Segment\Tracks\TrackEntry\FlagHearingImpaired" id="0x55AB" type="uinteger" range="0-1" default="0" maxOccurs="1">
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -323,6 +323,10 @@ this can be used for tracks containing only translations of foreign-language aud
     <documentation lang="en" purpose="definition">Set if that track is in the content's original language (not a translation). (1 bit)</documentation>
     <extension webm="0"/>
   </element>
+  <element name="FlagCommentary" path="\Segment\Tracks\TrackEntry\FlagCommentary" id="0x55AF" type="uinteger" range="0-1" default="0" maxOccurs="1">
+    <documentation lang="en" purpose="definition">Set if that track contains commentary. (1 bit)</documentation>
+    <extension webm="0"/>
+  </element>
   <element name="FlagLacing" path="\Segment\Tracks\TrackEntry\FlagLacing" id="0x9C" type="uinteger" range="0-1" default="1" minOccurs="1" maxOccurs="1">
     <documentation lang="en" purpose="definition">Set if the track **MAY** contain blocks using lacing. (1 bit)</documentation>
     <extension cppname="TrackFlagLacing"/>
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -319,6 +319,10 @@ this can be used for tracks containing only translations of foreign-language aud
     <documentation lang="en" purpose="definition">Set if that track contains textual descriptions of video content. (1 bit)</documentation>
     <extension webm="0"/>
   </element>
+  <element name="FlagOriginal" path="\Segment\Tracks\TrackEntry\FlagOriginal" id="0x55AE" type="uinteger" range="0-1" default="0" maxOccurs="1">
+    <documentation lang="en" purpose="definition">Set if that track is in the content's original language (not a translation). (1 bit)</documentation>
+    <extension webm="0"/>
+  </element>
   <element name="FlagLacing" path="\Segment\Tracks\TrackEntry\FlagLacing" id="0x9C" type="uinteger" range="0-1" default="1" minOccurs="1" maxOccurs="1">
     <documentation lang="en" purpose="definition">Set if the track **MAY** contain blocks using lacing. (1 bit)</documentation>
     <extension cppname="TrackFlagLacing"/>
--- a/notes.md
+++ b/notes.md
@@ -358,6 +358,103 @@ it as 0x53B9 instead of 0x53B8. `Matroska Readers` may support these legacy file
 Matroska v2 or 0x53B9. The older values were 0: mono, 1: right eye, 2: left eye, 3: both eyes.
 
 
+# Default track selection
+
+This section provides some example sets of Tracks and hypothetical user settings, along with
+indications of which ones a similarly-configured `Matroska Player` **SHOULD** automatically
+select for playback by default in such a situation. A player **MAY** provide additional settings
+with more detailed controls for more nuanced scenarios. These examples are provided as guidelines
+to illustrate the intended usages of the various supported Track flags, and their expected behaviors.
+
+Track names are shown in English for illustrative purposes; actual files may have titles
+in the language of each track, or provide titles in multiple languages.
+
+## Audio Selection
+
+Example track set:
+
+| No. | Type  | Lang | Layout | Original | Default | Other flags     | Name                  |
+| --- | ----- | ---- | ------ | -------- | ------- | --------------- | --------------------- |
+| 1   | Video | und  | N/A    | N/A      | N/A     | None            |                       |
+| 2   | Audio | eng  | 5.1    | 1        | 1       | None            |                       |
+| 3   | Audio | eng  | 2.0    | 1        | 1       | None            |                       |
+| 4   | Audio | eng  | 2.0    | 1        | 0       | Visual-impaired | Descriptive audio     |
+| 5   | Audio | esp  | 5.1    | 0        | 1       | None            |                       |
+| 6   | Audio | esp  | 2.0    | 0        | 0       | Visual-impaired | Descriptive audio     |
+| 7   | Audio | eng  | 2.0    | 1        | 0       | Commentary      | Director's Commentary |
+| 8   | Audio | eng  | 2.0    | 1        | 0       | None            | Karaoke               |
+
+Here we have a file with 7 audio tracks, of which 5 are in English and 2 are in Spanish.
+
+The English tracks all have the Original flag, indicating that English is the original content language.
+
+Generally the player will first consider the track languages: if the player has an option to prefer
+original-language audio and the user has enabled it, then it should prefer one of the Original-flagged tracks.
+If configured to specifically prefer audio tracks in English or Spanish, the player should select one of
+the tracks in the corresponding language. The player may also wish to prefer an Original-flagged track
+if no tracks matching any of the user's explicitly-preferred languages are available.
+
+Two of the tracks have the Visual-impaired flag. If the player has been configured to prefer such tracks,
+it should select one; otherwise, it should avoid them if possible.
+
+If selecting an English track, when other settings have left multiple possible options,
+it may be useful to exclude the tracks that lack the Default flag: here, one provides descriptive service for
+the visually impaired (which has its own flag and may be automatically selected by user configuration,
+but is unsuitable for users with default-configured players), one is a commentary track
+(which has its own flag, which the player may or may not have specialized handling for),
+and the last contains karaoke versions of the music that plays during the film, which is an unusual
+specialized audio service that Matroska has no built-in support for indicating, so it's indicated
+in the track name instead. By not setting the Default flag on these specialized tracks, the file's author
+hints that they should not be automatically selected by a default-configured player.
+
+Having narrowed its choices down, our example player now may have to select between tracks 2 and 3.
+The only difference between these tracks is their channel layouts: 2 is 5.1 surround, while 3 is stereo.
+If the player is aware that the output device is a pair of headphones or stereo speakers, it may wish
+to prefer the stereo mix automatically. On the other hand, if it knows that the device is a surround system,
+it may wish to prefer the surround mix.
+
+If the player finishes analyzing all of the available audio tracks and finds that multiple seem equally
+and maximally preferable, it **SHOULD** default to the first of the group.
+
+## Subtitle selection
+
+Example track set:
+
+| No. | Type      | Lang  | Original | Default | Forced | Other flags      | Name                               |
+| --- | --------- | ----  | -------- | ------- | ------ | ---------------- | ---------------------------------- |
+| 1   | Video     | und   | N/A      | N/A     | N/A    | None             |                                    |
+| 2   | Audio     | fra   | 1        | 1       | N/A    | None             |                                    |
+| 3   | Audio     | por   | 0        | 1       | N/A    | None             |                                    |
+| 4   | Subtitles | fra   | 1        | 1       | 0      | None             |                                    |
+| 5   | Subtitles | fra   | 1        | 0       | 0      | Hearing-impaired | Captions for the hearing-impaired  |
+| 6   | Subtitles | por   | 0        | 1       | 0      | None             |                                    |
+| 7   | Subtitles | por   | 0        | 0       | 1      | None             | Signs                              |
+| 8   | Subtitles | por   | 0        | 0       | 0      | Hearing-impaired | SDH                                |
+
+Here we have 2 audio tracks and 5 subtitle tracks. As we can see, French is the original language.
+
+We'll start by discussing the case where the user prefers French (or Original-language)
+audio (or has explicitly selected the French audio track), and also prefers French subtitles.
+
+In this case, if the player isn't configured to display captions when the audio matches their
+preferred subtitle languages, the player doesn't need to select a subtitle track at all.
+
+If the user _has_ indicated that they want captions to be displayed, the selection simply
+comes down to whether Hearing-impaired subtitles are preferred.
+
+The situation for a user who prefers Portuguese subtitles starts out somewhat analogous.
+If they select the original French audio (either by explicit audio language preference,
+preference for Original-language tracks, or by explicitly selecting that track), then the
+selection once again comes down to the hearing-impaired preference.
+
+However, the case where the Portuguese audio track is selected has an important catch:
+a Forced track in Portuguese is present. This may contain translations of onscreen text
+from the video track, or of portions of the audio that are not translated (music, for instance).
+This means that even if the user's preferences wouldn't normally call for captions here,
+the Forced track should be selected nonetheless, rather than selecting no track at all.
+On the other hand, if the user's preferences _do_ call for captions, the non-Forced tracks
+should be preferred, as the Forced track will not contain captioning for the dialogue.
+
 # Timestamps
 
 Historically timestamps in Matroska were mistakenly called timecodes. The `Timestamp Element`
--- a/notes.md
+++ b/notes.md
@@ -257,24 +257,22 @@ so that the `Matroska Player` can quickly switch from one to the other.
 
 ## Default flag
 
-The "default track" flag is a hint for a `Matroska Player` and **SHOULD** always be changeable
-by the user. If the user wants to see or hear a track of a certain kind (audio, video, subtitles)
-and hasn't chosen a specific track, the `Matroska Player` **SHOULD** use the first track
-of that kind whose "default track" flag is set to "1". If no such track is found,
-then the first track of this kind **SHOULD** be chosen.
+The "default track" flag is a hint for a `Matroska Player` indicating that a given track
+**SHOULD** be eligible to be automatically selected as the default track for a given
+language. If no tracks in a given language have the default track flag set, then all tracks
+in that language are eligible for automatic selection. This can be used to indicate that
+a track provides "regular service" suitable for users with default settings, as opposed to
+specialized services, such as commentary, hearing-impaired captions, or descriptive audio.
 
-Only one track of a kind **MAY** have its "default track" flag set in a segment.
-If a track entry does not contain the "default track" flag element, then its
-default value "1" is to be used.
+The `Matroska Player` **MAY** override the "default track" flag for any reason, including
+user preferences to prefer tracks providing accessibility services.
 
 ## Forced flag
 
-The "forced" flag tells the `Matroska Player` that it **MUST** display/play this track
-or another track of the same kind that also has its "forced" flag set. When there are multiple
-"forced" tracks, the `Matroska Player` **SHOULD** determine the track based upon the language
-of the forced flag or use the default flag if no track matches the use languages.
-Another track of the same kind without the "forced" flag may be use simultaneously
-with the "forced" track, like DVD subtitles.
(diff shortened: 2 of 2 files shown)


robUx4:
I cherry-picked patches a1a57b6 and a7ae4a5, hoping it would make the actual changes more visible in GitHub, but that's not the case...

ebml_matroska.xml
```diff
@@ -298,15 +298,35 @@ This **SHOULD** be kept the same when making a direct stream copy of the Track t
...
+  <element name="FlagVisualImpaired" path="\Segment\Tracks\TrackEntry\FlagVisualImpaired" id="0x55AC" type="uinteger" range="0-1" default="0" minOccurs="0" maxOccurs="1">
+    <documentation lang="en" purpose="definition">Set if that track is suitable for users with visual impairments. (1 bit)</documentation>
+    <extension cppname="TrackFlagVisualImpaired"/>
+  </element>
+  <element name="FlagDescriptions" path="\Segment\Tracks\TrackEntry\FlagDescriptions" id="0x55AD" type="uinteger" range="0-1" default="0" minOccurs="0" maxOccurs="1">
+    <documentation lang="en" purpose="definition">Set if that track contains textual descriptions of video content. (1 bit)</documentation>
+    <extension cppname="TrackFlagDescriptions"/>
+  </element>
+  <element name="FlagOriginal" path="\Segment\Tracks\TrackEntry\FlagOriginal" id="0x55AE" type="uinteger" range="0-1" default="0" minOccurs="0" maxOccurs="1">
+    <documentation lang="en" purpose="definition">Set if that track is in the content's original language (not a translation). (1 bit)</documentation>
```
robUx4, line 323: I think that's kind of the `FlagDefault` as well. It may actually be a saner description for `FlagDefault`, which explains why it should be used in absence of user preferred language (or if the user prefers the default language in general, like me).

ebml_matroska.xml
```diff
@@ -298,15 +298,35 @@ This **SHOULD** be kept the same when making a direct stream copy of the Track t
...
-the player **SHOULD** select the one which language matches the user preference or the default + forced track.
-Overlay **MAY** happen between a forced and non-forced track of the same kind. (1 bit)</documentation>
+    <documentation lang="en" purpose="definition">Applies only to subtitles. Set if that track SHOULD be active during playback by default if it matches the user's language preference,
+even if the user's preferences would normally not enable subtitles with the selected audio track;
+this can be used for tracks containing only translations of foreign-language audio or onscreen text. (1 bit)</documentation>
     <extension cppname="TrackFlagForced"/>
   </element>
+  <element name="FlagHearingImpaired" path="\Segment\Tracks\TrackEntry\FlagHearingImpaired" id="0x55AB" type="uinteger" range="0-1" default="0" minOccurs="0" maxOccurs="1">
+    <documentation lang="en" purpose="definition">Set if that track is suitable for users with hearing impairments. (1 bit)</documentation>
+    <extension cppname="TrackFlagHearingImpaired"/>
```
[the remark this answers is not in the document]
robUx4: On the other hand you should add the `webm="0"` extension.

ebml_matroska.xml
```diff
@@ -431,9 +451,15 @@ the overlay track **SHOULD** be used instead. The order of multiple TrackOverlay
 If not found it **SHOULD** be the second, etc.</documentation>
     <extension webm="0"/>
   </element>
+  <element name="TrackLink" path="\Segment\Tracks\TrackEntry\TrackLink" id="0x6FAC" type="uinteger">
+    <documentation lang="en" purpose="definition">Specify that subtitle track is suitable for use with the audio Track specified (in the u-integer).
```
robUx4, line 455: IMO that should be in a different PR as it may delay the rest of the mundane stuff. This is a completely new feature (or at some point we had it and removed it).

As it can be multiple you may group them in a `TrackLinks` element, ending up with `path="\Segment\Tracks\TrackEntry\TrackLinks\TrackLink"`.

It adds more complexity in muxers as it forces them to recompute this mapping for every muxing (unless the track layout is untouched in the remuxing).

It may apply to audio as well. For example you may have 2 angles of the same scene in different rooms with different audio. The audio should match the room it corresponds to.

notes.md
```diff
@@ -257,24 +257,51 @@ so that the `Matroska Player` can quickly switch from one to the other.
 
 ## Default flag
 
-The "default track" flag is a hint for a `Matroska Player` and **SHOULD** always be changeable
-by the user. If the user wants to see or hear a track of a certain kind (audio, video, subtitles)
-and hasn't chosen a specific track, the `Matroska Player` **SHOULD** use the first track
-of that kind whose "default track" flag is set to "1". If no such track is found,
-then the first track of this kind **SHOULD** be chosen.
+The "default track" flag is a hint for a `Matroska Player` indicating that a given track
+**SHOULD** be eligible to be automatically selected as the default track for a given
```
robUx4, line 261: The concept of "automatic selection" is tricky. The automatic selection may/should not be the same for impaired people who would prefer a higher preference for such tracks.
IMO it's up to the player to select what fits best the user. At the Matroska level we should only provide the information of which tracks does what.
That's why the `FlagDefault` makes also more sense as a `FlagOriginal`. That's a more useful information for the player/user.

The user may end up wanting the Original+Audio Impaired+Commentary track by default.

notes.md
```diff
@@ -257,24 +257,51 @@ so that the `Matroska Player` can quickly switch from one to the other.
...
+language. If no tracks in a given language have the default track flag set, then all tracks
+in that language are eligible for automatic selection. This can be used to indicate that
+a track provides "regular service" suitable for users with default settings, as opposed to
+specialized services, such as commentary, hearing-impaired captions, or descriptive audio.
 
-Only one track of a kind **MAY** have its "default track" flag set in a segment.
-If a track entry does not contain the "default track" flag element, then its
-default value "1" is to be used.
+The `Matroska Player` **MAY** override the "default track" flag for any reason, including
+user preferences to prefer tracks providing accessibility services.
```
robUx4, line 268: Then everything said above doesn't apply and the player is left with no clear rule on what to do. IMO we don't even need these rules.

notes.md
```diff
@@ -257,24 +257,51 @@ so that the `Matroska Player` can quickly switch from one to the other.
...
-The "forced" flag tells the `Matroska Player` that it **MUST** display/play this track
-or another track of the same kind that also has its "forced" flag set. When there are multiple
-"forced" tracks, the `Matroska Player` **SHOULD** determine the track based upon the language
-of the forced flag or use the default flag if no track matches the use languages.
-Another track of the same kind without the "forced" flag may be use simultaneously
-with the "forced" track, like DVD subtitles.
+The "forced" flag tells the `Matroska Player` that it **SHOULD** display this subtitle track,
+even if user preferences usually would not call for any subtitles to be displayed alongside
+the current selected audio track. This can be used to indicate that a track contains translations
+of onscreen text, or of dialogue spoken in a different language than the track's primary one.
```
robUx4, line 275: OK but does the player have to show the Forced subtitle track if the user is watching the commentary version of the content ? The Forced information should tell which track(s) it corresponds to.

notes.md
```diff
@@ -300,6 +327,23 @@ Overlay tracks **SHOULD** be rendered in the same channel as the track its linke
...
+- A forced subtitle track containing onscreen text consistent with the dialogue in an audio track
+  of the same language should be linked to that track, but should not be linked to another audio track
+  featuring a different translation into the same language.
+- A subtitle track featuring a transcription of an audio commentary track should be linked to
+  that audio commentary track, but not to other tracks featuring separate commentary recordings.
+
+Track links are only necessary when the set of tracks available creates ambiguity about which
+subtitles are appropriate for which audio, and function as a hint to the player to improve
+its automatic track selection behavior. In cases where there is no such ambiguity, there is
+no need to explicitly specify track links.
```
robUx4, line 345: A "default value" for `TrackLinks` (the list of `TrackLink` for a track) would help. An empty `TrackLink` means the track is not associated with any other track. So if you remux a file and remove an audio track, you also remove all the TrackLink for that track. The `TrackLinks` for those tracks might end up empty.

So setting a `TrackLinks` to empty is the opposite of having no `TrackLinks` at all which means it applies to all audio/video tracks.

notes.md
```diff
@@ -331,6 +375,69 @@ it as 0x53B9 instead of 0x53B8. `Matroska Readers` may support these legacy file
 Matroska v2 or 0x53B9. The older values were 0: mono, 1: right eye, 2: left eye, 3: both eyes.
 
 
+# Default track selection
+
+This section provides example algorithms that a `Matroska Player` **MAY** use to select
+default audio and subtitle tracks. These algorithms rely on the following settings being
+available to the user:
+
+- Preferred audio languages (zero or more language codes)
```
robUx4, line 384: "zero, one or more"
"zero or more" sounds strange.

notes.md
```diff
@@ -331,6 +375,69 @@ it as 0x53B9 instead of 0x53B8. `Matroska Readers` may support these legacy file
...
+- Preferred subtitle languages (zero or more language codes)
+- Prefer original-language audio (bool)
+- Display captions (bool)
+- Hearing-impaired (bool)
+- Visually-impaired (bool)
```
robUx4, line 389: - select commentaries (bool)

Most users would never do that but in some professional circles it may be useful.

notes.md
```diff
@@ -331,6 +375,69 @@ it as 0x53B9 instead of 0x53B8. `Matroska Readers` may support these legacy file
...
+2.  Iterate over the user's preferred subtitle languages, and for each one, append all matching tracks to the list.
+3.  If the list is empty, and either the "display captions" setting is enabled
+    or the selected audio track's language does not match one of the user's preferred subtitle languages,
+    append all subtitle tracks with unspecified language to the list.
+4.  If the list is empty, return no subtitle track. Stop.
+5.  If the "display captions" setting is enabled, and the list contains any non-forced tracks, remove all forced tracks from the list.
+6.  If the hearing-impaired setting is true, perform the following steps:
+  1. If the list contains tracks with the hearing-impaired flag, remove all tracks without the hearing-impaired flag from the list.
+  2. If the list contains tracks without the forced flag, remove all tracks with the forced flag from the list.
+7.  If the hearing-impaired setting is false, perform the following steps:
```
robUx4, line 431: If you move the `TrackLink` PR separately, this would move as well.

robUx4 (changes requested):
It's better to move the `TrackLink` in a different PR. The rest looks good with some minors changes needed here and there.
I would also merge `FlagOriginal` into `FlagDefault`.

ebml_matroska.xml
```diff
@@ -298,15 +298,35 @@ This **SHOULD** be kept the same when making a direct stream copy of the Track t
...
+  <element name="FlagHearingImpaired" path="\Segment\Tracks\TrackEntry\FlagHearingImpaired" id="0x55AB" type="uinteger" range="0-1" default="0" minOccurs="0" maxOccurs="1">
+    <documentation lang="en" purpose="definition">Set if that track is suitable for users with hearing impairments. (1 bit)</documentation>
+    <extension cppname="TrackFlagHearingImpaired"/>
+  </element>
+  <element name="FlagVisualImpaired" path="\Segment\Tracks\TrackEntry\FlagVisualImpaired" id="0x55AC" type="uinteger" range="0-1" default="0" minOccurs="0" maxOccurs="1">
+    <documentation lang="en" purpose="definition">Set if that track is suitable for users with visual impairments. (1 bit)</documentation>
+    <extension cppname="TrackFlagVisualImpaired"/>
+  </element>
+  <element name="FlagDescriptions" path="\Segment\Tracks\TrackEntry\FlagDescriptions" id="0x55AD" type="uinteger" range="0-1" default="0" minOccurs="0" maxOccurs="1">
+    <documentation lang="en" purpose="definition">Set if that track contains textual descriptions of video content. (1 bit)</documentation>
```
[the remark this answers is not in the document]
rcombs: I'm not entirely sure of the technical details here, but this exists in WebM (via `D_WEBVTT/DESCRIPTIONS`) and is described in the WebVTT spec, and apparently is used with both TTS and braille systems: https://www.w3.org/TR/media-accessibility-reqs/#text-video-description
It's also supported in MPEG-TS.

ebml_matroska.xml
rcombs: The issue is that a sizable portion of existing files don't use `FlagDefault` that way (i.e. they set it on a track that's _not_ in the original language); the new description of `FlagDefault` is designed to provide reasonable behavior even with existing files. The use I imagine for it is "provides regular service in a given language" (as opposed to any of the features that one would need to opt-in to get auto-selected, like descriptive audio or commentary or any other possible extended service), and there may be multiple such tracks per language (e.g. a 5.1 and a 2.0 track are reasonably common, or sometimes even 2 different codecs with the same channel layout).

ebml_matroska.xml
rcombs: Sure, I'll pull this out for now and we can discuss it in more detail in a separate PR after this one's all resolved.

notes.md
rcombs: Ultimately I intend this entire section to be advisory; "this is a reasonable algorithm that you can use as-is, or extend with additional functionality, or implement some subset of". Currently players are very inconsistent about how to implement any of this. It certainly doesn't help that the flags have been poorly-specified, but even with clear descriptions of what the individual flags do, I think it's helpful to provide an explanation of how they all end up used together, and having some sort of authoritative guidelines (even if they're fairly loose) would likely help with consistency across players, and thus with adoption of the flags by file authors.

If we decide not to describe an example algorithm, I'd suggest at least providing a few example sets of tracks, explain which flags a file author should set on which tracks, and show which ones a player would be expected to select by default in some example configuration cases.

rcombs: Erm, not the "default flag" section, the autoselection-algorithm section. Though the "default" flag is ultimately advisory (to the player) as well.

notes.md
rcombs: > OK but does the player have to show the Forced subtitle track if the user is watching the commentary version of the content?

Well, we can walk through the algorithm to see what happens here:

If the audio commentary matches the user's preferred subtitle language, and the user doesn't have captions enabled, then they get the forced track. If they _do_ have captions enabled, they get the regular captions track, unless there's a `TrackLink` (now-deferred to a later PR) to the audio commentary, in which case they get the commentary captions.

If the audio commentary _doesn't_ match the user's preferred subtitle language, then the user gets the default subtitles track in their preferred language, unless there's a `TrackLink` to the audio commentary, in which case they get the commentary subtitles.

We could in theory suggest that commentary-flagged subtitles be automatically selected when subtitles are appropriate (i.e. non-matching audio, or captions enabled) and a commentary audio track is selected, but I don't think that's reasonable functionality to try to provide without `TrackLink`, since it's reasonably common to have multiple commentary tracks, and autoselection without `TrackLink` in those cases would mean you'd get the wrong subtitles if you selected the second (or later) commentary audio.

> The Forced information should tell which track(s) it corresponds to.

That's what I'm going for with `TrackLink`, but you're right that it's a more complex and Matroska-specific feature, whereas everything else here is fairly generic in concept and exists in at least one other format.

notes.md
rcombs: I'm not sure this is a reasonable feature to attempt to provide, since there may be multiple entirely-different commentary tracks (and I don't think we really want to get into the nitty-gritty details of, say, director's commentary vs cast commentary vs translator's commentary vs [etc etc etc], that seems like more of a job for the track title). If a player does want to provide this, they're free to extend the functionality as they please.

ebml_matroska.xml
```diff
@@ -298,15 +298,35 @@ This **SHOULD** be kept the same when making a direct stream copy of the Track t
     <extension cppname="TrackFlagEnabled"/>
   </element>
   <element name="FlagDefault" path="\Segment\Tracks\TrackEntry\FlagDefault" id="0x88" type="uinteger" range="0-1" default="1" minOccurs="1" maxOccurs="1">
-    <documentation lang="en" purpose="definition">Set if that track (audio, video or subs) **SHOULD** be active if no language found matches the user preference. (1 bit)</documentation>
+    <documentation lang="en" purpose="definition">Set if that track (audio, video or subs) **SHOULD** be eligible for automatic selection by the player. (1 bit)</documentation>
```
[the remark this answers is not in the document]
rcombs: Not entirely sure how to add a reference like that; are there examples?

ebml_matroska.xml
```diff
@@ -298,15 +298,35 @@ This **SHOULD** be kept the same when making a direct stream copy of the Track t
...
     <extension cppname="TrackFlagDefault"/>
   </element>
   <element name="FlagForced" path="\Segment\Tracks\TrackEntry\FlagForced" id="0x55AA" type="uinteger" range="0-1" default="0" minOccurs="1" maxOccurs="1">
-    <documentation lang="en" purpose="definition">Set if that track **MUST** be active during playback. There can be many forced track for a kind (audio, video or subs),
-the player **SHOULD** select the one which language matches the user preference or the default + forced track.
-Overlay **MAY** happen between a forced and non-forced track of the same kind. (1 bit)</documentation>
+    <documentation lang="en" purpose="definition">Applies only to subtitles. Set if that track SHOULD be active during playback by default if it matches the user's language preference,
```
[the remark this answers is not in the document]
rcombs: Reworded; is this better?

notes.md
rcombs: (Removed for now)

rcombs:
Addressed technical and minor wording comments, removed TrackLink for now (for further discussion later), and rebased on upstream (which now has the whitespace changes). For future reference, you can add `?w=1` to a GitHub PR or commit URL to view the diff ignoring whitespace-only changes (like `diff -w` or `show -w` in git CLI).

As described above, I don't think we can reasonably redefine the Default flag to mean original-language without breaking backwards-compat, and also rendering Matroska's "Default" fundamentally different from any other format's. My goal there was to provide a rigorous definition for a "Default" flag that can be sanely used with existing files, while also applying to other formats, but at best it's still tenuous. I wouldn't object to deprecating it altogether, to be honest.

ebml_matroska.xml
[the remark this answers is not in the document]
robUx4: Yes, everything in the text with `(#xxx)`. In this case that would be `(#default-flag)` (as seen in the generated HTML source).

ebml_matroska.xml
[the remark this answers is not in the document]
robUx4: Yup

ebml_matroska.xml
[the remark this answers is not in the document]
robUx4: OK. I'd still prefer a less vague name. It's easier to read when referenced in other parts of the text. If not today, someday it will be.

ebml_matroska.xml
robUx4: Even a new flag may be misused so existing files may not be the right reference. Especially if some muxing tools just automatically set it on the first track or something.
On the other hand the fact there may be multiple versions of the same language track is definitely a good point.

notes.md
robUx4: If we provide guidelines, even if informative, they will become de-facto the standard way. Or someone will have to fork a better guideline if it has some flaws.

In the end if you look at web streaming platforms (supposedly the most flexible technology) they don't follow the same rules nor provide the same set of choices regarding the language, accessibility, or even audio channels.

So I'd rather let the player choose its strategy given the possibilities it has and explain which fields are useful to select the proper track(s). Some examples could be useful as long as they are not too complex that some players will not even implement those examples properly.

notes.md
robUx4: So this should go in the `TrackLink` PR as well ?

notes.md
robUx4: OK, but as said above we should at least define which elements are involved to pick the right track on startup (then the user can select whatever else). Two tracks with the same flags except for the commentary one should be picked in most cases. So it may be an option that noone ticks but at least the player should know that's it's preferred.

robUx4:
A few small unresolved left and then it's good to merge. Thanks for your work. This is definitely something needed in the specs.

ebml_matroska.xml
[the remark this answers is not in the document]
rcombs: Re: name, wouldn't it be `FlagTextDescriptions` or `FlagTextDescription` here? I've got no issue with that.

notes.md
rcombs: I don't think so; the `TrackLink` PR (or, well, currently it's just a branch but it can be a PR soon) describes a wholly new feature, whereas this description of Forced describes how an existing one is used.

rcombs:
Rewrote the autoselection guidelines in the form of a couple example track sets and descriptions of how a player might behave in various situations. Any thoughts?

My biggest concern here is that we make the intended usage and behavior of the Forced flag as clear as possible, to ensure that file authors understand when to use it, and player developers understand how it should behave, since historically both usage and behavior have been fairly inconsistent (and inconsistent player behavior results in file authors being unwilling to rely on it).

notes.md
```diff
@@ -358,6 +358,96 @@ it as 0x53B9 instead of 0x53B8. `Matroska Readers` may support these legacy file
...
+
+This section provides some example sets of Tracks and hypothetical user settings, along with
+indications of which ones a similarly-configured `Matroska Player` **SHOULD** automatically
+select for playback by default in such a situation. A player **MAY** provide additional settings
+with more detailed controls for more nuanced scenarios. These examples are provided as guidelines
+to illustrate the intended usages of the various supported Track flags, and their expected behaviors.
+
+## Audio Selection
+
+Example track set:
```
robUx4, line 371: you should describe (in parenthesis), what each field is:
(track type, language, audio channels, original , default flags).

Or maybe a table would look better to differentiate flags.

notes.md
```diff
@@ -358,6 +358,96 @@ it as 0x53B9 instead of 0x53B8. `Matroska Readers` may support these legacy file
...
+preference for Original-language tracks, or by explicitly selecting that track), then the
+selection once again comes down to the hearing-impaired preference.
+
+However, the case where the Portuguese audio track is selected has an important catch:
+a Forced track in Portuguese is present. This may contain translations of onscreen text
+from the video track, or of portions of the audio that are not translated (music, for instance).
+This means that even if the user's preferences wouldn't normally call for captions here,
+the Forced track should be selected nonetheless, rather than selecting no track at all.
+On the other hand, if the user's preferences _do_ call for captions, the non-Forced tracks
+should be preferred, as the Forced track will not contain captioning for the dialogue.
```
robUx4, line 449: Interresting. I think on DVDs the forced track doesn't have any language and is displayed no matter what. In this case the forced subtitles should have `lang=und` and be used no matter what.

Maybe we need another example with forced tracks where you have multiple language variants and an `und` fallback version which should be used even if there's no subtitle selected. IMO the forced flag is the most tricky one.

robUx4 (changes requested):

ebml_matroska.xml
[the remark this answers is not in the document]
robUx4: `FlagTextDescriptions`

notes.md
rcombs: > I think on DVDs the forced track doesn't have any language and is displayed no matter what.

DVDs don't generally have distinct forced tracks at all (though theoretically an author can set up the menus and media content any way they like); instead, they generally have a single track for each language, with individual events flagged as either forced or not at the codec level; same's true for blu-ray. Some software for converting DVD and BD media to Matroska has an option to create 2 tracks (one regular containing all events, and one containing only the forced ones, the latter having the matroska forced flag set). As far as I'm aware, this "per-event forced" concept doesn't exist in any subtitle codecs other than the DVD and BD ones.

> Maybe we need another example with forced tracks where you have multiple language variants and an `und` fallback version which should be used even if there's no subtitle selected.

What for? I've never heard of anyone trying to use the flag this way, and I can't think of a use-case (why would you want to display an additional text overlay that isn't part of the video, and that isn't in a language the user can read?). Even if we did say "if a forced `und` track exists and no other forced track matches the user's language preferences, default to it", saying "even if there's no subtitle selected" implies that the user can't disable subtitles altogether, which I can basically guarantee that no player is going to comply with.

notes.md
rcombs: Agreed that this is clearer as a table; reformatted and re-pushed.

notes.md
robUx4: I don't understand the remark of `saying "even if there's no subtitle selected" implies that the user can't disable subtitles altogether, which I can basically guarantee that no player is going to comply with`. I don't think it implies that. You can decide for a file/stream you don't want subtitles. That doesn't mean your preferred languages are not stored (they should be in your OS at the very least).

Multiple forced tracks should be cover. It may exist in the real world, and if it doesn't it should. For example imagine a scene where the whole dialog is not verbal but via SMS, displayed as forced subtitles over the video (I just had the example yesterday in Euphoria). This is a forced track, because even if you don't want subtitles (my case) you still want to know what they are saying (part of the plot). But I'd rather have it in English (also the original language) than French (what I actually got). So at least this stream should have 2 "forced" tracks. And the player should pick one for the user (English or original in my case).

So yes, maybe the `lang=und` doesn't make a lot of sense here. It might be used for imaginary languages (elf) which doesn't really have an ISO value.

In the end the language selection of the subtitle forced track is more or less the same as the regular subtitle track. It might be worth mentioning that there should only be one of each at any time.

robUx4 (changes requested):
The content of tables need some changes.

notes.md
rcombs: > I don't understand the remark of `saying "even if there's no subtitle selected" implies that the user can't disable subtitles altogether, which I can basically guarantee that no player is going to comply with`. I don't think it implies that. You can decide for a file/stream you don't want subtitles. That doesn't mean your preferred languages are not stored (they should be in your OS at the very least).

I don't quite get what you were trying to say to begin with then.

> Multiple forced tracks should be cover.

This is already handled, it's just not shown in the example currently.

> It may exist in the real world, and if it doesn't it should.

It definitely does. Take e.g. the Godfather trilogy: the blu-rays have regular captioning in several languages, but the tracks corresponding to the languages audio is available in (at a minimum, english, german, and french) also have forced events showing translations of dialogue spoken in Italian, and some onscreen text (like overlays naming the location or time period of a scene). So in Matroska, you end up with several regular tracks and 3 forced ones, and track selection (under the rules I'm proposing, at least) works exactly the way you're asking for it to.

> It might be used for imaginary languages (elf) which doesn't really have an ISO value.

I've never seen a subtitle track _in_ Elvish or some other fictional language, but the standard ISO code for that would be `mis` ("uncoded languages").

> In the end the language selection of the subtitle forced track is more or less the same as the regular subtitle track.

Basically, yeah; the language part is identical for either case, and then you just pick either a forced track or a non-forced one (or nothing if there isn't a track in the target language of the type you're looking for).

notes.md
```diff
@@ -331,6 +358,103 @@ it as 0x53B9 instead of 0x53B8. `Matroska Readers` may support these legacy file
...
+in the language of each track, or provide titles in multiple languages.
+
+## Audio Selection
+
+Example track set:
+
+| No. | Type  | Lang | Layout | Original | Default | Other flags     | Name                  |
+| --- | ----- | ---- | ------ | -------- | ------- | --------------- | --------------------- |
+| 1   | Video | und  | N/A    | N/A      | N/A     | None            |                       |
+| 2   | Audio | eng  | 5.1    | ✅       | ✅      | None            |                       |
```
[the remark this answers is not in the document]
rcombs: Switched to 0/1.

notes.md
robUx4: So my point was just the forced track selection should have an example of its own. In particular because it exists in parallel of the regular subtitles and may have different choices of languages. But it can be done later.

robUx4 (approved):
OK with a side note that the forced track might need another example and also note the choice is in parallel to the regular subtitle track.

rcombs:
Just to confirm, what do you mean by "in parallel"? Like, it's a similar decision, but every time a player's autoselection logic runs, it's only going to pick one track (which may either be forced or non-forced).

robUx4:
Well that's the thing, there are 2 tracks to select and run at the same time. I don't think regular subtitle tracks are required to contain the "forced" content in addition to the regular dialog. From my memory of how DVDs work there are 2 different layers of "transparency" over the video. 

There might be cases where all subtitle tracks contain the forced content and should not be used in parallel. So maybe we need something to let the player know which way it should be. Either via the `TrackLink` or maybe a global flag in the `Segment`. This flag could also have 3 values
- forced content is independant
- forced content is exclusive to non-forced content
- see TrackLink for track relations

rcombs:
> I don't think regular subtitle tracks are required to contain the "forced" content in addition to the regular dialog. From my memory of how DVDs work there are 2 different layers of "transparency" over the video.

No, again, in BD and DVD it's a single track that contains both (with per-event flags on the "forced" lines). I'm not sure if "dialogue only, no forced content" tracks have ever been used in any format; I've never seen any myself. People ripping BDs and DVDs to Matroska produce either 2 tracks (1 with all events, 1 forced-only; never dialogue-only) or a single one (with both types); people authoring subtitles for multi-audio MKV files do the former. I'm not aware of any player that supports overlaying multiple tracks this way (mpv supports playing 2 subtitle tracks at once, but the second one is moved from the bottom to the top of the screen, intended to be used to _compare_ tracks), and if players started doing so, it would break every file that already has a forced track and expects it not to be rendered alongside a normal track.

> There might be cases where all subtitle tracks contain the forced content and should not be used in parallel.

This is, to reiterate, _all current cases_, or at least all those that actually work in any player.

> TrackLink

This isn't what I intended `TrackLink` to be used for, and I think it'd be excessively confusing for it to perform both of these tasks.

If you'd like to introduce _new_ functionality for combining multiple tracks together, I'm sure it'd get some use; it's just not what the forced flag as it currently exists is used. This is actually already used in MPEGTS for _audio_ streams (a "dependent" stream can contain only secondary content, like commentary, descriptive audio service, or an overdubbed translation), and I don't think it's unreasonable to apply the same concept to subtitle streams. In MPEGTS as far as I'm aware it's a simple "this is a dependent stream" flag, but in Matroska it could be a more featureful system, perhaps extending the existing `TrackOverlay` element. But I don't think that necessarily has anything to do with this PR.

robUx4:
OK with what you said. And the subtitle track example is sufficient to cover regular use cases.

mkver:
I know I am a bit late here as this has already been merged, but when adding new fields with a default value, the default value has to be chosen in such a way as to not change the semantics of older files. This is not true for the recently added flags: The default value of `FlagOriginal` has just declared all Matroska tracks ever written to be dubbed; and `FlagCommentary` has just declared that no track ever written is a commentary and similarly for the accessibility flags. This is not ok.

mbunkus:
Please open an issue for it. Thanks.

Zwyx:
Hi guys and thanks for bringing to us the excellent Matroska format.

I have a few questions about this PR, as it brings a big change, if I understand correctly.

Let's take a file with two audio tracks like this, as an example:
```
| Type  | Lang | Default flag |
| ----- | ---- | ------------ |
...
| Audio | eng  | 0            |
| Audio | fre  | 1            |
...
```
At the moment, if open in VLC, the French track will be played, because it has the `default` flag.

But this PR changes the goal of the `default` flag. A `default` track is not _the_ default track of its kind (`audio`) anymore; it is the default of its kind _plus its language_ (`audio` and `fre`).

Which means that when VLC will implement these changes, then it will be the English track of our example file which will be played. Because it is first in the list (assuming that the user didn't parameter their preferred language).

It also means that we can have more than one `default` audio track; one per language.

This is a big change, it changes the meaning of `default` in a way that it might be hard to understand for users. It's so different, it feels like the `default` flag should have been totally abandoned, and a brand new word created for the new flag, to prevent confusion.

Anyway, do I understand this PR correctly? If yes, I might try to make a PR — or at least raise an issue — in MKVToolNix, as the GUI doesn't allow two tracks of the same kind to have the `default` flag. (Unless you're working on it at the moment @mbunkus?)

robUx4:
This is due to the fact that the `default` flag was incorrectly used so far. It could be set on multiple tracks on the same file. Plus it collides with the preferred language selection of the user, which is usually the thing to respect first (I don't think this is the case in VLC yet).

We could rename it to `eligible` but that doesn't sound too good either.

In the end this flag is not that much useful as other flags may override it. The non-presence of the other flags make a track a "default" track (eligible), unless marked so, ie explicitly writing the value 0 in the file.

Zwyx:
Thanks for your answer. This is really interesting. It indeed makes sense that the user preference is the thing to respect first, and that the description of the tracks is "for information only".

And it's true that `default` is probably the best word for this flag.

Hey, it might seems stupid, but I can't understand why there aren't much more people discussing in this repo. It seems so important to me what's happening here, especially since Matroska is the base of WebM. There should be folks from Google, Netflix, Amazon, etc. participating here!

Anyway, thanks again for your work :-)
Cheers

robUx4:
We have been working on the RFC for 5 (?) years. I can understand people move away after a while.

All these companies have found their savior in the form of fragmented MP4. I believe they only used WebM when they use VP9. I think no one is using AV1 in WebM in production.

Issue #453, opened 2021-02-21
Clarify distinctions of 0 values from missing elements for new non-mandatory flag elements


rcombs:
Following up on #394 and #449. It sounds like existing flag elements are going to be designated as simple booleans (where not-present always means the same as either 0 or 1, depending on which flag), but when discussing with @mkver, we came to the conclusion that it'd be useful for the newly-added flags (original, commentary, HI/VI, etc) to be able to be explicitly specified as 0, and for this to be distinct from not setting them at all. My understanding is that this is the behavior specified by the current spec and RFC, but it's unclear when looking at the current spec, particularly since the description of the "default" field on the Matroska website doesn't match what's in the RFC. I'm not sure how best to clarify this; maybe something in the notes file? Or possibly just by updating the legend entry for "Default" [in the Matroska-Org/Infrastructure repo](https://github.com/Matroska-Org/infrastructure/blob/5c235294cd9c8d7dc847ef1855bae3c2e48de569/website/transforms/ebml_schema2spectable.xsl#L17). Maybe this GHI would be more appropriate in that repo, it's just a little awkward since the documentation text is split across 2 places.

robUx4:
While the website is a good/quick/easy source to find something, it's no longer the most authoritative source. The Matroska RFC was pretty much the website content put in a RFC and then edited/formatted/corrected. The website is likely missing a lot of these changes.

As for the new track flags, you're correct in assuming if they are not present the meaning is different than being present (with a length of 0). They are not mandatory, so nothing written means we don't know.

rcombs:
Right, just, as-is someone reading the docs can easily come to the incorrect conclusion that 0 and unset are the same here, and I'm concerned that it'll result in improper implementation code. In particular, @mkver pointed out that the current mkvmerge code (and help text) doesn't clearly expose the distinction.

mbunkus:
> While the website is a good/quick/easy source to find something, it's no longer the most authoritative source. The Matroska RFC was pretty much the website content put in a RFC and then edited/formatted/corrected. The website is likely missing a lot of these changes.

The whole "Technical/Info" tree of pages is generated directly from the Matroska and EBML RFCs. It is definitely up to date. When in doubt, look at the [element specs page](https://www.matroska.org/technical/elements.html). Right below the "Matroska v4 element specification" header there's the information about the revision & date of the spec that was used to build the pages from.

[...]

PR #455, opened 2021-03-01
Clarify semantics of newly added flags elements

mkver:
Fixes issue #453.

I opted not to add a "In the absence of the element nothing can be inferred about ..." to every flag given that it is self-evident; but I can change this if desired.

### DIFF
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -308,27 +308,27 @@ this can be used for tracks containing only translations of foreign-language aud
 See (#default-track-selection) for more details. (1 bit)</documentation>
     <extension cppname="TrackFlagForced"/>
   </element>
-  <element name="FlagHearingImpaired" path="\Segment\Tracks\TrackEntry\FlagHearingImpaired" id="0x55AB" type="uinteger" range="0-1" default="0" maxOccurs="1">
+  <element name="FlagHearingImpaired" path="\Segment\Tracks\TrackEntry\FlagHearingImpaired" id="0x55AB" type="uinteger" range="0-1" maxOccurs="1">
     <documentation lang="en" purpose="definition">Set if that track is suitable for users with hearing impairments. (1 bit)</documentation>
     <extension webm="0"/>
     <extension divx="0"/>
   </element>
-  <element name="FlagVisualImpaired" path="\Segment\Tracks\TrackEntry\FlagVisualImpaired" id="0x55AC" type="uinteger" range="0-1" default="0" maxOccurs="1">
+  <element name="FlagVisualImpaired" path="\Segment\Tracks\TrackEntry\FlagVisualImpaired" id="0x55AC" type="uinteger" range="0-1" maxOccurs="1">
     <documentation lang="en" purpose="definition">Set if that track is suitable for users with visual impairments. (1 bit)</documentation>
     <extension webm="0"/>
     <extension divx="0"/>
   </element>
-  <element name="FlagTextDescriptions" path="\Segment\Tracks\TrackEntry\FlagTextDescriptions" id="0x55AD" type="uinteger" range="0-1" default="0" maxOccurs="1">
+  <element name="FlagTextDescriptions" path="\Segment\Tracks\TrackEntry\FlagTextDescriptions" id="0x55AD" type="uinteger" range="0-1" maxOccurs="1">
     <documentation lang="en" purpose="definition">Set if that track contains textual descriptions of video content. (1 bit)</documentation>
     <extension webm="0"/>
     <extension divx="0"/>
   </element>
-  <element name="FlagOriginal" path="\Segment\Tracks\TrackEntry\FlagOriginal" id="0x55AE" type="uinteger" range="0-1" default="0" maxOccurs="1">
+  <element name="FlagOriginal" path="\Segment\Tracks\TrackEntry\FlagOriginal" id="0x55AE" type="uinteger" range="0-1" maxOccurs="1">
     <documentation lang="en" purpose="definition">Set if that track is in the content's original language (not a translation). (1 bit)</documentation>
     <extension webm="0"/>
     <extension divx="0"/>
   </element>
-  <element name="FlagCommentary" path="\Segment\Tracks\TrackEntry\FlagCommentary" id="0x55AF" type="uinteger" range="0-1" default="0" maxOccurs="1">
+  <element name="FlagCommentary" path="\Segment\Tracks\TrackEntry\FlagCommentary" id="0x55AF" type="uinteger" range="0-1" maxOccurs="1">
     <documentation lang="en" purpose="definition">Set if that track contains commentary. (1 bit)</documentation>
     <extension webm="0"/>
     <extension divx="0"/>
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -309,27 +309,27 @@ See (#default-track-selection) for more details. (1 bit)</documentation>
     <extension cppname="TrackFlagForced"/>
   </element>
   <element name="FlagHearingImpaired" path="\Segment\Tracks\TrackEntry\FlagHearingImpaired" id="0x55AB" type="uinteger" range="0-1" maxOccurs="1">
-    <documentation lang="en" purpose="definition">Set if that track is suitable for users with hearing impairments. (1 bit)</documentation>
+    <documentation lang="en" purpose="definition">Set to 1 if that track is suitable for users with hearing impairments, set to 0 if it is unsuitable for users with hearing impairments.</documentation>
     <extension webm="0"/>
     <extension divx="0"/>
   </element>
   <element name="FlagVisualImpaired" path="\Segment\Tracks\TrackEntry\FlagVisualImpaired" id="0x55AC" type="uinteger" range="0-1" maxOccurs="1">
-    <documentation lang="en" purpose="definition">Set if that track is suitable for users with visual impairments. (1 bit)</documentation>
+    <documentation lang="en" purpose="definition">Set to 1 if that track is suitable for users with visual impairments, set to 0 if it is unsuitable for users with visual impairments.</documentation>
     <extension webm="0"/>
     <extension divx="0"/>
   </element>
   <element name="FlagTextDescriptions" path="\Segment\Tracks\TrackEntry\FlagTextDescriptions" id="0x55AD" type="uinteger" range="0-1" maxOccurs="1">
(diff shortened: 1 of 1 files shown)


mbunkus (approved):
I'm fine with the changes, both the documentation and removing their default values. Thanks.

ebml_matroska.xml
```diff
@@ -308,28 +308,28 @@ this can be used for tracks containing only translations of foreign-language aud
...
-    <documentation lang="en" purpose="definition">Set if that track is in the content's original language (not a translation). (1 bit)</documentation>
+  <element name="FlagOriginal" path="\Segment\Tracks\TrackEntry\FlagOriginal" id="0x55AE" type="uinteger" range="0-1" maxOccurs="1">
+    <documentation lang="en" purpose="definition">Set to 1 resp. 0 if that track is in the content's original language resp. is a translation.</documentation>
     <extension webm="0"/>
     <extension divx="0"/>
   </element>
-  <element name="FlagCommentary" path="\Segment\Tracks\TrackEntry\FlagCommentary" id="0x55AF" type="uinteger" range="0-1" default="0" maxOccurs="1">
-    <documentation lang="en" purpose="definition">Set if that track contains commentary. (1 bit)</documentation>
+  <element name="FlagCommentary" path="\Segment\Tracks\TrackEntry\FlagCommentary" id="0x55AF" type="uinteger" range="0-1" maxOccurs="1">
+    <documentation lang="en" purpose="definition">Set to 1 resp 0 if that track contains resp. does not contain commentary.</documentation>
```
hubblec4, line 332: here is resp written without a dot.
In all other cases you use "resp."

hubblec4:
Removing default values:

Yes this is for the moment maybe a good solution because of the problem with the attributes Mandatory and Default.

But default values are very useful, to save space and reading/writing processes. Especially with Boolean types  like the flags elements.

But the default values of 0 are not really useful.
For example the FlagCommentary element.
To signal that a track has not commentaries, we can omit(not write) the `FlagCommentary` and it is not necessary to write the elements header (of course we can omit the value, because at the moment is the default value 0).
Only if we want to signal there are commentaries include in the track we will use the `FlagCommentary` element.
And now it makes more sence to have a default value of 1. Only the element header must be written and this saves space of 1 byte.

For me personally more elements should have a default value.

ebml_matroska.xml
mkver: I just force-pushed a fix for this. Thanks.

mkver:
> Removing default values:
> 
> Yes this is for the moment maybe a good solution because of the problem with the attributes Mandatory and Default.
> 
> But default values are very useful, to save space and reading/writing processes. Especially with Boolean types like the flags elements.
> 
> But the default values of 0 are not really useful.
> For example the FlagCommentary element.
> To signal that a track has not commentaries, we can omit(not write) the `FlagCommentary` and it is not necessary to write the elements header (of course we can omit the value, because at the moment is the default value 0).
> Only if we want to signal there are commentaries include in the track we will use the `FlagCommentary` element.
> And now it makes more sense to have a default value of 1. Only the element header must be written and this saves space of 1 byte.

I agree, yet there are two compatibility problems with this approach. The first is with software that still uses the old "default value always present even if absent" semantics (libavformat's demuxer is probably the only one compliant to the new spec in this regard); the second is with files which use a zero-sized element for one of the new flags. While it is unlikely that they already exist, they would be broken by this.

hubblec4:
> While it is unlikely that they already exist, they would be broken by this.

Yes, that's happens also when we remove default values, and for some cases is the file structure broken/illegal.
When the element is written with a zero-length and there is no default value, that is more bad as to change the default value.

I guess not much files have this new element inside so the "damage" will be not so much, if we change the default value to 1.
But with removing the default value, the files will be broken.

mbunkus:
That assumes that a) there is a muxer that actually writes zero-length elements for elements with default values, b) that said muxer already knows about these elements and c) that the calling application/the user wanted to make use of these new elements.

`mkvmerge` satisfies b) but not a).

I'm very much of the opinion that this is a case where we can assume the amount of existing files to broken to be zero or near enough to justify the change.

hubblec4:
It now feels a bit like everyone is afraid of using default values.  :-)

Yeah I know that a looong time the usage of Default and Mandatory was not really clear and it was used in "wrong" way.
But we should look ahead and use the correct usage for Default and Mandatory because it has many advantages.

mbunkus:
In general… well maybe. In this particular case they don't gain us anything except for confusion. These are seldom-used track flags, elements that might be written less than ten times per file. The gains are around… ten bytes.

I'm for using default values where it matters: for elements that have to be written time and again in each file.

hubblec4:
> These are seldom-used track flags, elements that might be written less than ten times per file. The gains are around… ten bytes.

Of course, but for 1000 files (I have more) it will be save 10000 bytes.

For me it is OK (but not really good) when we remove the default values to avoid confusion in this "transition period" until the correct behavior of Mandatory and Default is used widely.

It is never a problem to add a Default value later, this will don't break existing files.

mkver:
> > While it is unlikely that they already exist, they would be broken by this.
> 
> Yes, that's happens also when we remove default values, and for some cases is the file structure broken/illegal.
> When the element is written with a zero-length and there is no default value, that is more bad as to change the default value.
> 

You are mistaken: Removing the default value of a non-mandatory element with default value of 0 does not break anything, because any empty integer/float element without a mandatory value will be read as if it had the value 0. That is

> I guess not much files have this new element inside so the "damage" will be not so much, if we change the default value to 1.
> But with removing the default value, the files will be broken.

No. Changing the default value might break files already making use of this feature (but I doubt they exist). Furthermore, explicitly making use of the zero-length = default value feature will make it harder for old software to support these elements properly (but it might also spur them to support the zero-length feature...).

hubblec4:
> because any empty integer/float element without a mandatory value will be read as if it had the value 0.

OK, that's new to me. Is this declared anywhere in the specs?

mkver:
"If the EBML Element is not defined to have a default value, then an Unsigned Integer Element with a zero-octet length represents an integer value of zero." (From [here](https://github.com/ietf-wg-cellar/ebml-specification/blob/master/specification.markdown#unsigned-integer-element); signed integers, floats and strings have equivalent text.)

hubblec4:
OK, thanks.

ready to merge?

hubblec4:
> (There is btw an element where adding a default value can't break any spec-compliant existing files: CueTrack. It is mandatory, so any valid CueTrackPositions has it; zero is not an allowed range, so no valid file can currently use an empty element for it. Adding a default value of 1 would therefore be harmless, yet such files would not be backwards compatible.)

Interesting, and a similar problem for the `CueClusterPosition` element. The value of 0 is allowed but a Cluster can never have a position of 0.

mkver:
> > (There is btw an element where adding a default value can't break any spec-compliant existing files: CueTrack. It is mandatory, so any valid CueTrackPositions has it; zero is not an allowed range, so no valid file can currently use an empty element for it. Adding a default value of 1 would therefore be harmless, yet such files would not be backwards compatible.)
> 
> Interesting, and a similar problem for the `CueClusterPosition` element. The value of 0 is allowed but a Cluster can never have a position of 0. Or I'm wrong?

A cluster can have a position of zero (it means that the cluster is the very first element in the segment). But there just isn't a sensible default value for `CueClusterPosition`, so it doesn't matter anyway.

hubblec4:
> A cluster can have a position of zero (it means that the cluster is the very first element in the segment). But there just isn't a sensible default value for CueClusterPosition, so it doesn't matter anyway.

OK. 

> "If the EBML Element is not defined to have a default value, then an Unsigned Integer Element with a zero-octet length represents an integer value of zero." (From here; signed integers, floats and strings have equivalent text.)

This means automatically that we have ever an auto-default value, except for the Binary and Master element.
Is this really a good idea?

mkver:
> > "If the EBML Element is not defined to have a default value, then an Unsigned Integer Element with a zero-octet length represents an integer value of zero." (From here; signed integers, floats and strings have equivalent text.)
> 
> This means automatically that we have ever an auto-default value, except for the Binary and Master element.
> Is this really a good idea?

Please file an issue against the EBML specification if you want to discuss this. This PR is not the proper place for it.

ebml_matroska.xml
```diff
@@ -308,28 +308,28 @@ this can be used for tracks containing only translations of foreign-language aud
 See (#default-track-selection) for more details. (1 bit)</documentation>
     <extension cppname="TrackFlagForced"/>
   </element>
-  <element name="FlagHearingImpaired" path="\Segment\Tracks\TrackEntry\FlagHearingImpaired" id="0x55AB" type="uinteger" range="0-1" default="0" maxOccurs="1">
-    <documentation lang="en" purpose="definition">Set if that track is suitable for users with hearing impairments. (1 bit)</documentation>
+  <element name="FlagHearingImpaired" path="\Segment\Tracks\TrackEntry\FlagHearingImpaired" id="0x55AB" type="uinteger" range="0-1" maxOccurs="1">
+    <documentation lang="en" purpose="definition">Set to 1 resp. 0 if that track is suitable resp. unsuitable for users with hearing impairments.</documentation>
```
robUx4, line 312: Not a fan of `resp.`, I had to look up online to find out what you mean. I think it's `respectively`. We don't need to compress words here.

robUx4: Even with the words expanded the text is still hard to process. I would prefer something like

> Set to 1 if that track is suitable for users with hearing impairments, 0 otherwise.

Or even

> Set to 1 if that track is suitable for users with hearing impairments, 0 if unsuitable for users with hearing impairments.

Which clearly says if the value is explicitly set to 0 it's not suitable.

robUx4 (changes requested):
In favor of this change although I'd prefer clearer text.

Also since they are new elements it's even less likely there are use of them in the wild that might be misinterpreted. It fixes #453 for those. I hope Github doesn't close the issue once this is merged.

ebml_matroska.xml
mbunkus: > Set to 1 if that track is suitable for users with hearing impairments, 0 if unsuitable for users with hearing impairments.

I like this wording very much as it's much clearer that 0 is different from the element not being present. @mkver please adjust all descriptions this way. Thanks.

mkver: I force-pushed a more explicit version. (Apologies for not knowing that "resp." is not a well-known abbreviation. It is in German.)

robUx4 (approved):
LGTM

PR #457, opened 2021-03-07
Make ordered chapters flag mandatory

robUx4:
An edition can be either ordered or not ordered, but there is no "undefined" value.

If it's not written or with a zero-length interpreters already assume it's no ordered.
Most likely, they don't even know about ordered chapters anyway.


hubblec4:
OK

mbunkus (approved):

hubblec4 (approved):

PR #459, opened 2021-03-07
Remove the "1 bit" mention in the specification

robUx4:
We do not store one bit. It was just meant to be only 0 and 1 should be used.
The range of these elements already specifies that.

Use the "Set to 1" phrasing to be coherent with the rest of the spec. Here the
"0" value is implied since they are mandatory elements (except for EditionFlagOrdered
but see #457)


hubblec4:
LGTM

retokromer (approved):

PR #461, opened 2021-03-07
Move control track related content in a separate document

### LINKED ISSUE #456: Remove mentions of Control Track in the main spec
So far there is no player (not even VLC with DVD menus) that can handle Control tracks. There is also no tool to create them.

This can be added in a separate document with its own RFC. We just need to keep the value in the `TrackEntry\TrackType` as reserved.

We can probably do the same with `complex` (mixed DV), `logo` (unused AFAIK). Not sure about `metadata`.

robUx4:
The new document is in a dirty state and we may never use it. But it's a way to clean the Control Track reference (an under specified and unused feature) out of the main document.

Fixes #456

### DIFF
--- a/chapters.md
+++ b/chapters.md
@@ -35,21 +35,21 @@ Edition   | FlagDefault | Default Edition
 Edition 1 | true        | X
 Edition 2 | true        |
 Edition 3 | true        |
-Table: Default edition, all default{#defaultEditionSomeVisibleAllDefault}
+Table: Default edition, all default{#defaultEditionAllDefault}
 
 Edition   | FlagDefault | Default Edition
 :---------|:------------|:---------------
 Edition 1 | false       | X
 Edition 2 | false       |
 Edition 3 | false       |
-Table: Default edition, no default{#defaultEditionAllHiddenNoDefault}
+Table: Default edition, no default{#defaultEditionNoDefault}
 
 Edition   | FlagDefault | Default Edition
 :---------|:------------|:---------------
 Edition 1 | false       |
 Edition 2 | true        | X
 Edition 3 | false       |
-Table: Default edition, with default{#defaultEditionAllHiddenWithDefault}
+Table: Default edition, with default{#defaultEditionWithDefault}
 
 ### EditionFlagOrdered
 
--- a/chapters.md
+++ b/chapters.md
@@ -17,30 +17,6 @@ An `Edition` contains a set of `Edition` flags and **MUST** contain at least one
 Chapters are always inside an `Edition` (or a Chapter itself part of an `Edition`).
 Multiple Editions are allowed. Some of these Editions **MAY** be ordered and others not.
 
-### EditionFlagHidden
-
-When the `EditionFlagHidden` flag is set to `false` it means the `Edition` is visible and selectable
-in a `Matroska Player`.
-All `ChapterAtoms Elements` **MUST** be interpreted with their own `ChapterFlagHidden` flags.
-
-ChapterFlagHidden | False | True | visible
-:-----------------|:------|:-----|:-------
-Chapter 1         |   X   |      | yes
-Chapter 2         |       | X    | no
-Table: ChapterAtom visibility to the user{#chapterVisibility}
-
-When the `EditionFlagHidden` flag is set to `true` the `Edition` is hidden and **SHOULD NOT** be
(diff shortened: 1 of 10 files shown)


hubblec4:
> move ChapterTrack and ChapterTrackUID in the Control Tracks spec
Is this really only for Control Tracks?

A good player could read the ChapterTrack and uses only the listed tracks (ordered chapters are required).

robUx4:
`ChapterTrack` is useless without a `ChapterTrackUID`, because it means "all tracks", just as if it doesn't exist.

It's true that the `ChapterTrackUID` spec currently says
> In the absence of a control track, choosing this chapter will select the listed Tracks and deselect unlisted tracks.

But just like control track it's not currently used so I'd rather move it out for now than delaying the specifications for something noone has ever used. Neither vlc nor libavformat support these elements.

robUx4:
BTW, the specs contained as many features as we could think of, even not correctly/properly defined because the idea was to have as much "prior art" as possible. So if anyone tries to patent a new feature they can't because we already thought of it. Even the chapter codecs fall in this category except it got to a proof of concept in VLC.

We don't need these extra parts in the v1 to v4 specs which is meant to specify what is currently used in Matroska.

chapters.md
```diff
@@ -339,8 +320,7 @@ of them contain another splitting.
           <ChapLanguage>eng</ChapLanguage>
         </ChapterDisplay>
         <ChapterFlagHidden>0</ChapterFlagHidden>
-        <ChapterFlagEnabled>1</ChapterFlagEnabled>
-      </ChapterAtom>
+        </ChapterAtom>
```
mbunkus, line 323: This change in indentation is wrong.

chapters.md
```diff
@@ -350,8 +330,7 @@ of them contain another splitting.
...
```
mbunkus, line 333: This change in indentation is wrong.

chapters.md
```diff
@@ -361,8 +340,7 @@ of them contain another splitting.
...
```
mbunkus, line 343: This change in indentation is wrong.

chapters.md
```diff
@@ -372,10 +350,8 @@ of them contain another splitting.
...
```
mbunkus, line 353: This change in indentation is wrong.

index_control.md
```diff
@@ -0,0 +1,70 @@
...
+stream = "IETF"
+status = "informational"
+value = "@BUILD_VERSION@"
+
+[[author]]
+initials="S."
+surname="Lhomme"
+fullname="Steve Lhomme"
+ [author.address]
+  email="slhomme@matroska.org"
```
mbunkus, line 22: inconsistent indentation (should probably be two spaces for both lines)

index_control.md
```diff
@@ -0,0 +1,70 @@
...
+
+[[author]]
+initials="M."
+surname="Bunkus"
+fullname="Moritz Bunkus"
+  [author.address]
+ email="moritz@bunkus.org"
```
mbunkus, line 29: inconsistent indentation (should probably be two spaces for both lines)

chapters.md
```diff
@@ -24,35 +24,21 @@ set to `true` and the second with `ChapterFlagHidden` either set to `false` or n
...
 
 
 #### EditionFlagDefault
 
 It is **RECOMMENDED** that no more than one `Edition` have an `EditionFlagDefault Flag`
-set to `true`. The first `Edition` with both the `EditionFlagDefault Flag` set to `true`
-and the `EditionFlagHidden Flag` set to `false` is the Default Edition. When all
-`EditionFlagDefault Flags` are set to `false`, then the first `Edition` with the
-`EditionFlagHidden Flag` set to `false` is the Default Edition. The Default Edition
+set to `true`. The first `Edition` with the `EditionFlagDefault Flag` set to `true`
```
mbunkus, line 38: Is the trailing space here intentional? Trailing spaces are significant for Markdown (two spaces == forced line break),
though this instance looks more like an accident.

mbunkus (changes requested):

robUx4:
Rebased after the #342 merge which had more text based on now-moved elements. As a result some tables and text have been simplified.

robUx4:
Rebase after #469 merged.

PR #464, opened 2021-03-14
The chapter end must be higher to the start value

### LINKED ISSUE #114: Chapter times logic and nested Chapters
I found in DVDMenuXtrator's(DMX) output-xml files, smaller end times than start times.
I have try to understand the Matroska-DVD-Menu logic but it is very hard. I think the end times in DVD-Matroska Menu are not every time needed. 
[TNG-S6-D1.zip](https://github.com/Matroska-Org/matroska-specification/files/844061/TNG-S6-D1.zip)
I hope Steve Lohmme can explain this a bit.

I know, DMX is not really in use and the support in VLC is not 100% working. 
But we should mention such things to complete the specs.

Chapter times are dependent on two things:
1. ordered edition
  - without an ordered edition, only start times are used and relevant (end times can be written but without an effect to the mkv/splitter).

2. linking system (we have 3 linking systems (hard; soft; medium))
 - hard linking: this linking is independent from the chapter times and all end times must be greater than
    the start times. In my tests with hard linking, the hard linking system will be "overwritten" 
    by medium linking when ordered editions are defined. 

- soft linking: this linking is used by Matroska-DVD-Menu system and the chapter times seems to have
   her own rules.

- medium linking: this linking is used by Chapter-Segment-Linking with an ordered edition
  chapter time rules are: 
  - end times greater than start times (normal behavior)
  - end times equal to start times: in this case a nested chapter with a start time equal to the parent
    start/end time sould be used otherwise is such a chapter only a chapter mark.
    With such a behaviour is it possible to build more complex medium-linking structures and in
    combination with the Matroska-Native-Menu(and VOB-Butons) we could have almost unlimited
    possibility for own Matroska menus.

robUx4:
Fixes #114

### DIFF
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -1253,7 +1253,8 @@ Use for WebVTT cue identifier storage [@!WebVTT].</documentation>
     <extension type="webmproject.org" webm="1"/>
   </element>
   <element name="ChapterTimeEnd" path="\Segment\Chapters\EditionEntry\+ChapterAtom\ChapterTimeEnd" id="0x92" type="uinteger" maxOccurs="1">
-    <documentation lang="en" purpose="definition">Timestamp of the end of Chapter (timestamp excluded, not scaled).</documentation>
+    <documentation lang="en" purpose="definition">Timestamp of the end of Chapter (timestamp excluded, not scaled).
+The value **MUST** be equal or higher than the `ChapterTimeStart` of the same `ChapterAtom`.</documentation>
     <extension type="webmproject.org" webm="1"/>
   </element>
   <element name="ChapterFlagHidden" path="\Segment\Chapters\EditionEntry\+ChapterAtom\ChapterFlagHidden" id="0x98" type="uinteger" range="0-1" default="0" minOccurs="1" maxOccurs="1">
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -1254,7 +1254,7 @@ Use for WebVTT cue identifier storage [@!WebVTT].</documentation>
   </element>
   <element name="ChapterTimeEnd" path="\Segment\Chapters\EditionEntry\+ChapterAtom\ChapterTimeEnd" id="0x92" type="uinteger" maxOccurs="1">
     <documentation lang="en" purpose="definition">Timestamp of the end of Chapter (timestamp excluded, not scaled).
-The value **MUST** be equal or higher than the `ChapterTimeStart` of the same `ChapterAtom`.</documentation>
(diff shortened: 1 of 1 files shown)


hubblec4:
LGTM

hubblec4:
Mmmh...
about "equal": that means the chapter duration is 0(zero), which is illigal, because such chapters are ignored for building the virtual timeline.

Maybe it's better to say the end time MUST be greater as the start time.

robUx4:
Good catch. There doesn't seem to be anything restricting the duration to be non-zero. It could be possible to have a Chapter in an ordered edition with no duration just to do some things with `ChapProcess`. But there is already a `ChapProcessTime` for entering and leaving so whatever Chapter that is before or after this one could have this command.

The incoherence is with the definition of `ChapterTimeEnd` itself. It says the timestamp of the `ChapterTimeEnd` is excluded. If `ChapterTimeStart` and `ChapterTimeEnd` are equal the timestamp is both included and excluded.

IMO the exclusion of the timestamp in `ChapterTimeEnd` must be kept. Given a timestamp you should be able with certainty which Chapter it belongs to, especially in ordered editions. That's key to making the transition from one chapter to another at the right time.
Also chapters with a duration of zero are useless. In non-ordered editions you don't have to use a `ChapterTimeEnd` if you just want a marker in the file. In ordered editions, as said above, you can merge things you need for that chapter in neighbor chapters. Only a marker right at the end of the file may not be possible but I'm not sure it's of any use.

retokromer (approved):

hubblec4:
Then we agree on this.
Yes, the exclusion of the timestamp in `ChapterTimeEnd` is very important and for me totally logic and useful.

ebml_matroska.xml
```diff
@@ -1253,7 +1253,8 @@ Use for WebVTT cue identifier storage [@!WebVTT].</documentation>
     <extension type="webmproject.org" webm="1"/>
   </element>
   <element name="ChapterTimeEnd" path="\Segment\Chapters\EditionEntry\+ChapterAtom\ChapterTimeEnd" id="0x92" type="uinteger" maxOccurs="1">
-    <documentation lang="en" purpose="definition">Timestamp of the end of Chapter (timestamp excluded, not scaled).</documentation>
+    <documentation lang="en" purpose="definition">Timestamp of the end of Chapter (timestamp excluded, not scaled).
+The value **MUST** be strictly higher than the `ChapterTimeStart` of the same `ChapterAtom`.</documentation>
```
JeromeMartinez, line 1257: higher --> greater.

robUx4: Fixed

Issue #330, opened 2019-07-22
[wish] image flipping


retokromer:
As [requested](https://github.com/cellar-wg/matroska-specification/pull/326#issuecomment-513852464) by @dericed:

I suggest to add the handling of specular images (vertical and/or horizontal flipping). For the people implementing Matroska into players it’s of the same family than rotation.

**Context:** One film reel can be digitised emulsion to sensor or base to sensor, from platter A to platter B or from platter B to platter A, and there are four combinations for the rotation of the platters. In case of highly damaged films, it’s actually the film itself who decides which configuration is chosen, not the operating person (no joke!). It’s useful when the raw scan can be played correctly without having to modify the player’s parameters.

robUx4:
As said in #326 I think `ProjectionPoseYaw` and `ProjectionPosePitch` can be used for horizontal and vertical flipping respectively. We just need to document the special case of 180 degree for each.

robUx4:
Duplicate of #326

PR #466, opened 2021-03-14
documentation image flipping through ProjectionPoseYaw/ProjectionPosePitch

### LINKED ISSUE #327: document image flipping
see https://github.com/cellar-wg/matroska-specification/pull/326#issuecomment-511088866

robUx4:
We don't need to mention what values the other elements must be. But default the `ProjectionType` is rectangular which corresponds to 2D rendering.

Fixes #327.


ebml_matroska.xml
```diff
@@ -885,7 +885,9 @@ redundant framing information while preserving versioning and semantics between
 
 Value represents a clockwise rotation, in degrees, around the up vector. This rotation must be applied
 before any `ProjectionPosePitch` or `ProjectionPoseRoll` rotations.
-The value of this field should be in the -180 to 180 degree range.</documentation>
+The value of this field should be in the -180 to 180 degree range.
+
+Setting `ProjectionPoseYaw` to 180.0 degrees flips the image horizontally.</documentation>
```
mbunkus, line 890: In general I'm fine with the change. But what's the difference between -180 and 180? Why recommend 180 but not -180?

In addition: please use `180` instead of `180.0`. The prior sentence states that `180` is the limit anyway; therefore any decimal part than 0 would be outside the range anyway.

ebml_matroska.xml
```diff
@@ -894,7 +896,9 @@ The value of this field should be in the -180 to 180 degree range.</documentatio
 
 Value represents a counter-clockwise rotation, in degrees, around the right vector. This rotation must be applied
 after the `ProjectionPoseYaw` rotation and before the `ProjectionPoseRoll` rotation.
-The value of this field should be in the -90 to 90 degree range.</documentation>
+The value of this field should be in the -90 to 90 degree range.
+
+Setting `ProjectionPosePitch` to 180.0 degrees flips the image vertically.</documentation>
```
mbunkus, line 901: This is… confusing. We give the expected range and, at the same moment, tell the user to use a value completely outside the range.

Apart from that the same two comments as above (why only positive, not negative, and `180` instead of `180.0`).

mbunkus (changes requested):

ebml_matroska.xml
hubblec4: Yeah, that confuses me too.

hubblec4:
We can/should mention that a value of -180 has the same effect as a value of 180.

mcr:
It's Tau/2 day.  Shouldn't we be using radians? :-)

ebml_matroska.xml
robUx4: Indeed, I didn't notice that. I'll check where this -90,90 limit comes from.

robUx4:
> It's Tau/2 day. Shouldn't we be using radians? :-)

It's too late for that and the values are in the same format as `ISOBMFF Equirectangular Projection Box`.

ebml_matroska.xml
robUx4: Fixed by allowing negative values and removing the decimals.

ebml_matroska.xml
robUx4: replaced the flip vertical with a ProjectionPoseRoll+ProjectionPoseYaw operation.

ebml_matroska.xml
```diff
@@ -885,7 +885,9 @@ redundant framing information while preserving versioning and semantics between
 
 Value represents a clockwise rotation, in degrees, around the up vector. This rotation must be applied
 before any `ProjectionPosePitch` or `ProjectionPoseRoll` rotations.
-The value of this field should be in the -180 to 180 degree range.</documentation>
+The value of this field should be in the -180 to 180 degree range.
+
+Setting `ProjectionPoseYaw` to 180 or -180 degrees flips the image horizontally.</documentation>
```
robUx4, line 890: Given the vertical flip also involves the same value, we might mention it's a horizontal flip if `ProjectionPoseRoll` is 0 and the Pitch is 0.

ebml_matroska.xml
```diff
@@ -903,7 +905,9 @@ The value of this field should be in the -90 to 90 degree range.</documentation>
 
 Value represents a counter-clockwise rotation, in degrees, around the forward vector. This rotation must be applied
 after the `ProjectionPoseYaw` and `ProjectionPosePitch` rotations.
-The value of this field should be in the -180 to 180 degree range.</documentation>
+The value of this field should be in the -180 to 180 degree range.
+
+Setting `ProjectionPoseRoll` to 180 or -180 degrees and the `ProjectionPoseYaw` to 180 or -180 degrees flips the image vertically.</documentation>
```
robUx4, line 910: Should mention that the pitch should be 0 for this to apply.

robUx4:
Fixes #330

ebml_matroska.xml
```diff
@@ -882,7 +882,9 @@ redundant framing information while preserving versioning and semantics between
 
 Value represents a clockwise rotation, in degrees, around the up vector. This rotation must be applied
 before any `ProjectionPosePitch` or `ProjectionPoseRoll` rotations.
-The value of this field should be in the -180 to 180 degree range.</documentation>
+The value of this field should be in the -180 to 180 degree range.
+
+Setting `ProjectionPoseYaw` to 180 or -180 degrees, with the `ProjectionPoseYaw` and `ProjectionPosePitch` set to 0 degrees flips the image horizontally.</documentation>
```
mbunkus, line 887: One of the two `ProjectionPoseYaw` should be `ProjectionPoseRoll`, doesn't it?

mbunkus (changes requested):
Apart from that one oversight: yeah, this clashes with #481. I don't care the order in which we apply things.

ebml_matroska.xml
robUx4: yes, good catch

robUx4: Fixed

retokromer (approved):

hubblec4 (approved):

Issue #467, opened 2021-03-21
Move Chapter codec definitions in separate documents


robUx4:
Unlike chapters which are a core feature, each "codec" has its own definition that doesn't need to be set in the main document.

Just like we have a separate document for audio+video+subtitle codecs, we should have one for chapters (and even one per codec IMO).

These sections should move out of chapters.md:
```
### Matroska Script (0)
### DVD menu (1)
```

hubblec4:
Yes I agree with this.
The Chapter codec should have its own document, because it is a own "world".
Many features are in my mind and waiting :-)

PR #469, opened 2021-03-21
Add a separate document that will receive all the Chapter Codecs parts removed

robUx4:
Implements the changes for #467

### DIFF
--- a/.github/workflows/generate-rfcs.yaml
+++ b/.github/workflows/generate-rfcs.yaml
@@ -44,6 +44,7 @@ jobs:
           mv draft-ietf-cellar-matroska-??.xml artifacts
           mv draft-ietf-cellar-codec-??.xml artifacts
           mv draft-ietf-cellar-tags-??.xml artifacts
+          mv draft-ietf-cellar-chapter-codecs-??.xml artifacts
 
       - name: XML Artifact
         uses: actions/upload-artifact@master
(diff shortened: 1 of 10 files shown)


robUx4:
There might also be legal issues mentioning the DVD specs/format which is not public or open. In the end it might be a RFC document but something we do on the side.

hubblec4:
LGTM

> There might also be legal issues mentioning the DVD specs/format which is not public or open.

There are many web sites which shows DVD-specs in more details as Matroska does.

JeromeMartinez:
> There are many web sites which shows DVD-specs in more details as Matroska does.

It does not mean that it is legal ;-).
But generally speaking I guess that this is part of fair use.

SpencerDawkins:
Could I ask what the proper reference for DVD format is that you're thinking of?

I'm happy to chase this with ADs and IETF lawyers if that's helpful.

retokromer (approved):

robUx4:
The DVD specs are found in "DVD Books" https://web.archive.org/web/20100501012533/http://www.mpeg.org/MPEG/DVD/General/Books.html
They must be bought from the [DVD Forum](http://www.dvdforum.org).

JeromeMartinez:
I suggest to merge this PR (technical task) and open an issue for tracking the question about potential legal issue in putting a (very small) part from the DVD spec.

PR #471, opened 2021-03-28
Add subtitle->audio track linking

rcombs:
Useful for cases where the appropriate tracks are otherwise ambiguous; see added notes.md section for a few examples.

Also clarified that `TrackOverlay` refers to tracks by their UIDs rather than their numbers; as far as I can tell this is the intent, but I can't find anything existing confirming this either way (nor any implementation code), so I could be wrong, but it seems consistent with other elements.

Previously it was suggested that `TrackLink` should go in a dedicated container element with maxOccurs=1, but I don't see any particular technical reason for that, and `TrackOverlay` already allows multiple instances without a container, though I can add one if that's a policy for newly-added elements or something.

### DIFF
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -448,7 +448,7 @@ see [@!I-D.ietf-cellar-codec] for more info.</documentation>
     <extension type="webmproject.org" webm="0"/>
   </element>
   <element name="TrackOverlay" path="\Segment\Tracks\TrackEntry\TrackOverlay" id="0x6FAB" type="uinteger">
-    <documentation lang="en" purpose="definition">Specify that this track is an overlay track for the Track specified (in the u-integer).
+    <documentation lang="en" purpose="definition">Specify that this track is an overlay track for the Track UID specified (in the u-integer).
 That means when this track has a gap, see (#silenttracks-element) on SilentTracks,
 the overlay track **SHOULD** be used instead. The order of multiple TrackOverlay matters, the first one is the one that **SHOULD** be used.
 If not found it **SHOULD** be the second, etc.</documentation>
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -454,6 +454,12 @@ the overlay track **SHOULD** be used instead. The order of multiple TrackOverlay
 If not found it **SHOULD** be the second, etc.</documentation>
     <extension type="webmproject.org" webm="0"/>
   </element>
+  <element name="TrackLink" path="\Segment\Tracks\TrackEntry\TrackLink" id="0x6FAC" type="uinteger">
+    <documentation lang="en" purpose="definition">Specify that subtitle track is suitable for use with the audio Track UID specified (in the u-integer).
+Multiple subtitle tracks may link to the same audio track, and the same subtitle track may link to multiple audio tracks by providing multiple TrackLink elements.
+</documentation>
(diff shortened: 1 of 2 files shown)


ebml_matroska.xml
```diff
@@ -454,6 +454,12 @@ the overlay track **SHOULD** be used instead. The order of multiple TrackOverlay
 If not found it **SHOULD** be the second, etc.</documentation>
     <extension type="webmproject.org" webm="0"/>
   </element>
+  <element name="TrackLink" path="\Segment\Tracks\TrackEntry\TrackLink" id="0x6FAC" type="uinteger">
+    <documentation lang="en" purpose="definition">Specify that subtitle track is suitable for use with the audio Track UID specified (in the u-integer).
+Multiple subtitle tracks may link to the same audio track, and the same subtitle track may link to multiple audio tracks by providing multiple TrackLink elements.
```
robUx4, line 459: It should mention that the TrackEntry it belongs to `**MUST**` be a subtitle track, in a normative way.

notes.md
```diff
@@ -327,6 +327,27 @@ Overlay tracks **SHOULD** be rendered in the same channel as the track its linke
...
+  A subtitle track containing a transcription of the dub, if present, should be linked to the dub audio track.
+- A forced subtitle track containing onscreen text consistent with the dialogue in an audio track
+  of the same language should be linked to that track, but should not be linked to another audio track
+  featuring a different translation into the same language.
+- A subtitle track containing a transcription of an audio commentary track should be linked to
+  that audio commentary track, but not to other tracks featuring separate commentary recordings.
+
+Track links are only necessary when the set of tracks available creates ambiguity about which
+subtitles are appropriate for which audio, and function as a hint to the player to improve
+its automatic track selection behavior. In cases where there is no such ambiguity, there is
```
robUx4, line 348: There should be a link to the `(#default-track-selection)` section here. And probably augment the "Subtitle selection" section on how to use the `TrackLink` to solve such ambiguity.

robUx4 (changes requested):

robUx4:
The `TrackOverlay` refers to the SilentTrack feature which itself has `SilentTrackNumber`. This is a track number.

Given there's probably 0 use anywhere in the wild of this feature I don't mind the change. We may also change the name to `TrackUIDOverlay` or `OverlayTrackUID`.

Now given all of these are unused we might as well mark them as deprecated so they are not used while not specified properly or are of any use to anyone. This feature is basically used to replace parts of a track by another track the container level. This can be useful in video to encode a whole movie normally and have some parts with translated text on the video encoded in different tracks. But then it says the overlay picks the first available track, so I don't know how this can be of any use.

Maybe silently replacing parts of a video that have been removed by another part without reencoding, maybe even using a different codec. In this scenario the multiple `TrackOverlay` and the order to pick them would correspond to different edits of the video file.
You should move the `TrackOverlay` change in a separate PR as it's not used at all by the `TrackLink` feature.

For the `TrackLink` part, I think all the values should be held in a master element. When translating the spec to code, if the generated code has hierarchy (ie not a flat structure) it's easier to look for the master element and then its children, rather than iterating through all the elements of the `TrackEntry` to find all the `TrackLink` values.

rcombs:
Forgot about this PR for (checks notes) almost 4 years, but I've rebased this and addressed all the comments.

PR #473, opened 2021-04-04
make ChapterTimeEnd mandatory for ordered chapters

### LINKED ISSUE #470: Should we make ChapterTimeEnd mandatory for ordered chapters ?
Right now it's never mandatory but in ordered chapters we need to the have the duration of each parts to play.

robUx4:
Fixes #470

### DIFF
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -1252,6 +1252,7 @@ Use for WebVTT cue identifier storage [@!WebVTT].</documentation>
   <element name="ChapterTimeEnd" path="\Segment\Chapters\EditionEntry\+ChapterAtom\ChapterTimeEnd" id="0x92" type="uinteger" maxOccurs="1">
     <documentation lang="en" purpose="definition">Timestamp of the end of Chapter (timestamp excluded, not scaled).
 The value **MUST** be strictly greater than the `ChapterTimeStart` of the same `ChapterAtom`.</documentation>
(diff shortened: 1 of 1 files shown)


hubblec4 (approved):
LGTM

retokromer (approved):
Nit: there are two spaces at `(#editionflagordered),  then`, otherwise LGTM.

hubblec4:
> Looking at my old DVDMenuxtractor output there are cases where only the lower nested ChapterAtom had a ChapterTimeEnd (it's legal since the element is not mandatory). The parent ChapterAtom's assumed the end from the children end.

So we can't make this element mandatory.

robUx4:
I don't consider the output of DVDMenuxtractor as a normative. The only reason it works in VLC is because to find the end of a chapter (in ordered editions) it considers the "current" chapter as the lowest level Chapter within the start/end of the current playing time. So it's just by luck that it never needs to check the parent element. But IMO that is not clean. If an intermediate chapter level has some command with a `ChapProcessTime` set to "2 (after playback of the chapter)" it **MUST** have a `ChapterTimeEnd`. It should not rely on its child(ren) element(s) having one. That makes the handling of chapterprocess more complicated for nothing. 

Another example  is if the child chapter(s) data may be damaged or voided. The non broken level should still be usable to determine its ChapterTimeEnd needed for proper ordered chapter playback.

In the end it's easier to make it mandatory for ordered chapters than refining the case where it MUST be and the ones that may not be necessary.

mbunkus (approved):

PR #475, opened 2021-04-04
make BlockGroup\Slices max version 1

robUx4:
All child elements are either maxver 0 or 1. We don't need to allow it any further.


hubblec4 (approved):
I fully agree. This Master element has no sub-element in use.

retokromer (approved):

PR #476, opened 2021-04-04
Only use the webm extension to say an element is supported

robUx4:
An element is either supported or not. We only need one version in our file.
And it makes more sense to tell the elements that are in WebM than list all the
ones that are not.


retokromer (approved):

hubblec4 (approved):
Yeah that makes totally sense.

Issue #305, opened 2019-01-29
Reference deprecated IDs separately


robUx4:
These elements have nothing to do in the main spec. On top of that the spec is not available from DivX anymore.

They should be an extension, expect the existing files using this (trick play) feature don't have the EBML extension header in them.

robUx4:
Also the `divx` attribute we use in the XML Schema is not a valid attribute to define EBML.

hubblec4:
I have never seen a file with DivX elements. I agree with the removal.

mkver:
I also agree with the removal and have two more questions: Were these elements just used by DivX without consulting the Matroska guys? And if these elements are removed, does this mean that their `Element ID` will be available again and might be reused for something completely different? (The DivX guys used one octet `Element IDs` for elements that only exist once in a `TrackEntry` of a track using these DivX extensions. What a waste!)

dericed:
https://github.com/cellar-wg/matroska-specification/pull/325 is related to this issue as it moves the divx attributes however the related elements still are there.

robUx4:
The DivX elements were only created by DivX tools AFAIK and only handled by their player(s). They put the documentation online but never consulted with us.
Now the specification is gone.

IMO the IDs should not be reused to maintain compatiblity with such files (if the same ID was used in the future with a different type or different meaning it might lead to serious issues). Again, the IDs alone could be found in different parts of the file and never really collide. So IMO we should keep track of the path theses IDs were used in the past and not reuse them there. We can keep the elements in the spec and set `maxver=0` meaning they can never be used.

dericed:
In the EBML spec we say that maxver MUST >= than minver. In some cases like 'SimpleBlock' there is a minver=2. IMHO to update that to say minver=2 and maxver=0 is invalid. I'd suggest that use maxver=2 in this case and perhaps add a comment to say that they are not recommended.

robUx4:
Given there are not official we could say both `minver=0` and `maxver=0`. A validator will consider existing files invalid. But the elements would still be known and documented.

dericed:
Does that imply that there is a Matroska version 0?

robUx4:
This is already how we marked elements many elements as existing but should not be used (`BlockVirtual`, `FrameNumber`, `EncryptedBlock`, etc).

Also all the DivX specific elements already have `minver=0` and `maxver=0` so we don't even need to do anything special.

robUx4:
[...]

PR #477, opened 2021-04-04
Only use the divx extension to say an element is supported

robUx4:
An element is either supported or not. We only need one version in our file.
And it makes more sense to tell the elements that are specific to DiVX than
list all the ones that are not.

The DivX-only element are not part of the Matroska spec either but here are placeholders/explanation if they are ever encountered. See #305


hubblec4 (approved):
Yeah that makes totally sense.

retokromer (approved):

PR #479, opened 2021-04-05
mark strings with no default value as not usable if empty

robUx4:
Wether it's a mandatory or non-mandatory element, if a string has no default
value, and it's written in the file it should have a value, otherwise don't write
it.

For mandatory an empty string could be a way cancel the default value. But we
allow it for elements with a default value.

It may be legal in EBML to write empty strings for some reason, but
we don't have the use case in Matroska.

### DIFF
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -385,7 +385,7 @@ the value refers to the BlockAddID ((#blockaddid-element)), value being describe
 To keep MaxBlockAdditionID as low as possible, small values **SHOULD** be used.</documentation>
     <extension type="webmproject.org" webm="0"/>
   </element>
-  <element name="BlockAddIDName" path="\Segment\Tracks\TrackEntry\BlockAdditionMapping\BlockAddIDName" id="0x41A4" type="string" maxOccurs="1" minver="4">
+  <element name="BlockAddIDName" path="\Segment\Tracks\TrackEntry\BlockAdditionMapping\BlockAddIDName" id="0x41A4" type="string" length="&gt; 0" maxOccurs="1" minver="4">
     <documentation lang="en" purpose="definition">A human-friendly name describing the type of BlockAdditional data,
 as defined by the associated Block Additional Mapping.</documentation>
     <extension type="webmproject.org" webm="0"/>
@@ -410,13 +410,13 @@ see (#language-codes) on language codes.
 This Element **MUST** be ignored if the LanguageIETF Element is used in the same TrackEntry.</documentation>
     <extension type="libmatroska" cppname="TrackLanguage"/>
   </element>
-  <element name="LanguageIETF" path="\Segment\Tracks\TrackEntry\LanguageIETF" id="0x22B59D" type="string" minver="4" maxOccurs="1">
+  <element name="LanguageIETF" path="\Segment\Tracks\TrackEntry\LanguageIETF" id="0x22B59D" type="string" length="&gt; 0" minver="4" maxOccurs="1">
     <documentation lang="en" purpose="definition">Specifies the language of the track according to [@!BCP47]
 and using the IANA Language Subtag Registry [@!IANALangRegistry].
 If this Element is used, then any Language Elements used in the same TrackEntry **MUST** be ignored.</documentation>
     <extension type="webmproject.org" webm="0"/>
   </element>
-  <element name="CodecID" path="\Segment\Tracks\TrackEntry\CodecID" id="0x86" type="string" minOccurs="1" maxOccurs="1">
+  <element name="CodecID" path="\Segment\Tracks\TrackEntry\CodecID" id="0x86" type="string" length="&gt; 0" minOccurs="1" maxOccurs="1">
     <documentation lang="en" purpose="definition">An ID corresponding to the codec,
 see [@!I-D.ietf-cellar-codec] for more info.</documentation>
   </element>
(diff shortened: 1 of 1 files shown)


robUx4:
The same might be done for all other types as well.

robUx4:
Rebased and use `not 0` as the length cannot be negative.

Related to #501 as somehow any string should be NOT empty. Maybe we should also mention it in the Matroska Design as well. Not there can be a case where we want to write an empty string. It's definitely not possible with elements with a default value as it would be interpreted as the default value (which can't be empty in the EBML Schema).

mkver:
I see that you left `FileName` out of this patch. Is this intended? By the wording that accompanies the patches you intend it to be included (yet forgot it because it is an utf8-element?). Yet there is a usecase for empty `FileName` elements: It allows to convert attachments from other formats that don't provide a `FileName` while still being in line with the specs.

robUx4:
I missed the `utf-8` strings in the search & replace. Fixed in new push.

The case you describe for `FileName` sounds more like it should not be mandatory, rather than writing a fake empty value (which can never be written on a filesystem). But it's been mandatory since the dawn of time.

ebml_matroska.xml
```diff
@@ -1200,24 +1200,24 @@ Absence of this Element indicates that the Chapter **SHOULD** be applied to any
...
     <extension type="webmproject.org" webm="1"/>
   </element>
-  <element name="ChapString" path="\Segment\Chapters\EditionEntry\+ChapterAtom\ChapterDisplay\ChapString" id="0x85" type="utf-8" minOccurs="1" maxOccurs="1">
+  <element name="ChapString" path="\Segment\Chapters\EditionEntry\+ChapterAtom\ChapterDisplay\ChapString" id="0x85" type="utf-8" length="not 0" minOccurs="1" maxOccurs="1">
     <documentation lang="en" purpose="definition">Contains the string to use as the chapter atom.</documentation>
     <extension type="webmproject.org" webm="1"/>
     <extension type="libmatroska" cppname="ChapterString"/>
   </element>
-  <element name="ChapLanguage" path="\Segment\Chapters\EditionEntry\+ChapterAtom\ChapterDisplay\ChapLanguage" id="0x437C" type="string" default="eng" minOccurs="1">
+  <element name="ChapLanguage" path="\Segment\Chapters\EditionEntry\+ChapterAtom\ChapterDisplay\ChapLanguage" id="0x437C" type="string" length="not 0" default="eng" minOccurs="1">
```
robUx4, line 1208: We could be even stricter on the length and only a size of 3

ebml_matroska.xml
```diff
@@ -347,29 +347,29 @@ to define how the BlockAdditional data should be handled.</documentation>
...
   </element>
-  <element name="Language" path="\Segment\Tracks\TrackEntry\Language" id="0x22B59C" type="string" default="eng" minOccurs="1" maxOccurs="1">
+  <element name="Language" path="\Segment\Tracks\TrackEntry\Language" id="0x22B59C" type="string" length="not 0" default="eng" minOccurs="1" maxOccurs="1">
     <documentation lang="en" purpose="definition">Specifies the language of the track in the Matroska languages form;
 see (#language-codes) on language codes.
 This Element **MUST** be ignored if the LanguageIETF Element is used in the same TrackEntry.</documentation>
     <extension type="libmatroska" cppname="TrackLanguage"/>
   </element>
-  <element name="LanguageIETF" path="\Segment\Tracks\TrackEntry\LanguageIETF" id="0x22B59D" type="string" minver="4" maxOccurs="1">
+  <element name="LanguageIETF" path="\Segment\Tracks\TrackEntry\LanguageIETF" id="0x22B59D" type="string" length="not 0" minver="4" maxOccurs="1">
```
robUx4, line 360: We could add a restriction on the length which would be 3 for Matroska 1-3 and `>= 3 & <=6` for Matroska 4

ebml_matroska.xml
```diff
@@ -1356,18 +1356,18 @@ If set to any other value, it **MUST** match the `FileUID` value of an attachmen
     <extension type="webmproject.org" webm="1"/>
     <extension type="libmatroska" cppname="TagSimple"/>
   </element>
-  <element name="TagName" path="\Segment\Tags\Tag\+SimpleTag\TagName" id="0x45A3" type="utf-8" minOccurs="1" maxOccurs="1">
+  <element name="TagName" path="\Segment\Tags\Tag\+SimpleTag\TagName" id="0x45A3" type="utf-8" length="not 0" minOccurs="1" maxOccurs="1">
     <documentation lang="en" purpose="definition">The name of the Tag that is going to be stored.</documentation>
     <extension type="webmproject.org" webm="1"/>
   </element>
-  <element name="TagLanguage" path="\Segment\Tags\Tag\+SimpleTag\TagLanguage" id="0x447A" type="string" default="und" minOccurs="1" maxOccurs="1">
+  <element name="TagLanguage" path="\Segment\Tags\Tag\+SimpleTag\TagLanguage" id="0x447A" type="string" length="not 0" default="und" minOccurs="1" maxOccurs="1">
```
robUx4, line 1363: We could add a restriction on the length which would be 3 for Matroska 1-3 and `>= 3 & <=6` for Matroska 4

ebml_matroska.xml
mbunkus: That doesn't make sense.

1. LanguageIETF wasn't introduced until v4.
2. The minimum length for a valid IETF BCP 47 language tag is two characters as all ISO 639 alpha 2 codes are valid (e.g. a simple `en`).
3. There is no maximum length for valid IETF BCP 47 language tags. Due to their "extensions" and "private use" sub-parts, they can have arbitrary length.

For the sake of not confusing anyone, just leave out any length restriction. Just state that it must be a valid IETF BCP 47 language tag.

robUx4: Indeed, only the v4 constraint makes sense.

Having machine readable constraints make it easier to validate files on reading/writing (and in mkvalidator) without knowing too much of the underlying data (which would be a deeper analysis than the container level).

As you said, it can be very lengthy so only `not 0` is possible here.

robUx4:
Maybe we could create default values for mandatory elements. Like `default=""` it seems legal in RFC8794. This is line with #501 which tolerates 0/0x0p+0 default values. It corresponds to `NUL` in a string which also OK in legacy parsers.

That leaves the issue on non mandatory elements. Empty Elements (length=0) are either using the default value or an empty string ([section 6.1 of RFC8794](https://www.rfc-editor.org/rfc/rfc8794.html#name-data-size-format)). In the absence of a default value (this Pull Request), that means an empty string (or a string with the first char is `NUL`). This works with legacy parsers.

So both cases already work with legacy parsers. We don't actually need to change anything because the Empty Element works well with legacy parsers.

Forcing a new requirement on elements out of the blue doesn't seem like a good idea. So I'm closing this.

PR #481, opened 2021-04-11
add a stronger normative range for the ProjectionPose values

### LINKED ISSUE #480: clarify the range of Projection values
Right now there is a `should` in lowercase in the text. The range of each value is probably a `MUST` to respect and should be normative, either with the `range=` attribute of the element or with a `usage notes` documentation.

robUx4:
Fixes #480

Unfortunately the range attribute format may not be precise enough to declare which boundaries are included or excluded and might be tricky with signed values. Plus float in hexa is not very readable so we need a proper text value anyway.

Will collide with #466

### DIFF
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -881,8 +881,8 @@ redundant framing information while preserving versioning and semantics between 
     <documentation lang="en" purpose="definition">Specifies a yaw rotation to the projection.
 
 Value represents a clockwise rotation, in degrees, around the up vector. This rotation must be applied
-before any `ProjectionPosePitch` or `ProjectionPoseRoll` rotations.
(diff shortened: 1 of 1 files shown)


retokromer (approved):
I guess, it’s better this way.

retokromer (approved):

mbunkus:
I'm unclear why the one new sentence has to go into an implementation note. Doesn't it make just as much sense in the main description?

robUx4:
> 
> 
> I'm unclear why the one new sentence has to go into an implementation note. Doesn't it make just as much sense in the main description?

A lot of the `<implementation_note>` elements contain a MUST/SHOULD for the specific attribute they define. IMO it's more specific and clearer than the generic `<documentation>` that explains what the element means.

PR #487, opened 2021-04-11
Move the deprecated elements at the end of the specs

### LINKED ISSUE #305: Reference deprecated IDs separately

robUx4:
And move the IANA Considerations at the end.

Maybe we need to define what fields should be in the registry ? At least the ID, the name, and likely the path.
It might use the same registry fields as the regular/used Matroska IDs.

Fixes #305

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ bootstrap.mak
 ebml_matroska_elements.md
 ebml_matroska_elements4rfc.md
 control_elements4rfc.md
+*_deprecated4rfc.md
 matroska_tagging_registry.md
 control_xsd.xml
 matroska_xsd.xml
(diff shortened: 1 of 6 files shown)


robUx4:
Add some text explaning that "players don't have to support these deprecated elements".

mbunkus (approved):

hubblec4 (approved):

retokromer (approved):

mbunkus:
@robUx4 This branch has conflicts now, too. Please rebase & then merge. Thanks.

robUx4:
I had to rework the patch a bit as notes were no longer the last markdown file processed.
We have 37 deprecated elements !

PR #491, opened 2021-04-11
Deprecate internal signature on the content

### LINKED ISSUE #45: Remove encryption elements that are not used
Except for the ones used for DRM already [in use in WebM](https://www.webmproject.org/docs/webm-encryption/).

robUx4:
Unlike encryption/compression this is not used by anyone. (maybe in mkvtoolnix @mbunkus ?)

Partially fixes #45


mbunkus (approved):
`mkvinfo` can output the elements if they exist.

`mkvmerge` doesn't really use them, and in fact, I'll remove code using them later today.

PR #493, opened 2021-04-12
fix UTF-8 that appear in the TXT output

robUx4:
When uploading the XML it warns that some UTF8 chars are found and won't let me upload.
This should fix this warning+error.


JeromeMartinez:
Not possible to keep the UTF-8 version when not TXT?

robUx4:
I'm sending a version without the UTF-8 to see if it goes through.

robUx4:
It still tells me to fix things...

robUx4:
It went through without, so closing.
Thanks @dericed !

PR #494, opened 2021-04-18
make new track flags v4+

robUx4:
If they are found on older versions it's likely an error/bad data.

poke @rcombs


rcombs:
LGTM 👍.

retokromer (approved):

Issue #355, opened 2019-11-06
EBML lacing with only one frame


mkver:
Although it makes no sense to use lacing when a (Simple)Block is only supposed to contain a single frame, the specifications don't disallow this and the syntax doesn't make it impossible: The lacing head only contains the `number of frames in the lace - 1` and not `- 2`, which makes it possible to indicate that a lace only contains a single frame. In this case there is an ambiguity with [EBML lacing](https://github.com/cellar-wg/matroska-specification/blob/20e8df3465fbd23d861af15e3f7cc97b68c9912f/index_matroska.md#ebml-lacing): On the one hand, the first frame's size is coded via an ordinary unsigned EBML VINT; on the other hand, the size of the last frame is always derived from the total size of the block. So what should it be if the first frame is the last frame? I guess the latter should prevail or that such laces should be treated as invalid.

mkvmerge seems to treat such laces as invalid; FFmpeg meanwhile reads the beginning of the buffer as unsigned EBML VINT, checks whether it has read a valid EBML VINT and checks whether the size is not too big for the rest of the buffer and sets this to be the size of the first frame in the lace. And then it sets the size of the last frame in the lace to be the rest of the buffer which of course overwrites the size just set (and results in unaccounted space in the buffer).

I am not aware of any implementation writing such blocks.

robUx4:
In fixed lacing the `Lace-coded size of each frame` part doesn't exist. So we could say that is the same for other forms of lacing if there's only one frame in the lace. In that case the size of the last frame (which happens to be the first) is the size remaining in the lace data minus all the previous frames (none).

In general I think the lacing section should be reworked an maybe separated as it's the same data in a `Block` or a `SimpleBlock`. The blocks should be defined and then how to handle the different forms of lacing. The formating as flat text also makes it more confusing that the table it originates from https://matroska.org/technical/specs/index.html#simpleblock_structure

robUx4:
It seems OK to write a lace with only one frame. It's not optimal but a simple audio writer that doesn't deal with anything else wouldn't need a special case to handle very short files or the last frame of a long file. It can use the same lacing it uses for every other blocks.

robUx4:
It seems that libebml behaves differently with a single frame between Xiph/fixed-size and EBML lacing. In the case of EBML the size of the first frame is always written. In the others it's omitted. It's unfortunate and we can't change it now. So we'll have to make this explicit in the specs.

writing parts
```
  case LACING_XIPH:
[...]

PR #495, opened 2021-04-18
Reorder/Reparent sections in the Matroska RFC

robUx4:
The sections are currently hard to follow. The flow of information doesn't go logically.
I reordered a few sections to make things a little better (more improvements are probably needed).

This is needed before #355 where some parts of the Block sections need some work. It's better if they are in a finalize place/file in the document.

The EBML contraints from the Schema were duplicate and less clear than the manually created section.

**NO CONTENT WAS CHANGED,** only text moved and `#` levels edited.


robUx4:
I uploaded the [old TXT layout](https://github.com/robUx4/matroska-specification/blob/old-layout/draft-ietf-cellar-matroska-08.txt) (before this PR) and the [new TXT layout](https://github.com/robUx4/matroska-specification/blob/new-layout/draft-ietf-cellar-matroska-08.txt) (after this PR) so it's easier to [compare them](https://github.com/robUx4/matroska-specification/pull/1/files).

The summary/index is the part to focus on, to see where the section were/are. The text within these sections is untouched.
One of the main changes is that the elements sections in the summary are less details (not every single elements of SeekHead/Info/etc are listed). They are not needed in the summary/index.

mcr (approved):

retokromer (approved):

SpencerDawkins:
I made this comment on the mailing list, but then realize that I should have put it in github - and half of this is about the reordering topic. 

I see that Michael has already approved the PR in https://github.com/robUx4/matroska-specification/branches, but I did have one quick question, after a quick look through the table of contents. 

       9.1.4.  Tracks Element  . . . . . . . . . . . . . . . . . . .  36
       9.1.5.  Cues Element  . . . . . . . . . . . . . . . . . . . . 102

So, there's a 66-page gap between 9.1.4 and 9.1.5. The specification has plenty of 9.1.4.n subsections, but they don't show up in the TOC, because the "tocDepth" attribute isn't specified in the XML, and defaults to "3", so they won't be easily spotted unless you know what to search for in the text. 

Perhaps you could consider adding "tocDepth=4" to 

<rfc version="3" ipr="trust200902" docName="draft-ietf-cellar-matroska-08" sortRefs="true" submissionType="IETF" category="info" xml:lang="en" xmlns:xi="http://www.w3.org/2001/XInclude" consensus="true">

and see if the result is manageable?

SpencerDawkins:
While cross-checking the XML, I noticed that

<title abbrev="Matroska Format">Matroska Media Container Format Specifications</title><seriesInfo value="draft-ietf-cellar-matroska-08" stream="IETF" status="informational" name="Internet-Draft"></seriesInfo>

says that the specification is "informational", but the CELLAR charter at https://datatracker.ietf.org/wg/cellar/about/ says that the Matroska specifications would be Standards Track, so status="std". 

Would it make sense to make that change at the same time?

Best,

Spencer

robUx4:
> ```
>    9.1.4.  Tracks Element  . . . . . . . . . . . . . . . . . . .  36
>    9.1.5.  Cues Element  . . . . . . . . . . . . . . . . . . . . 102
> ```
> 
> So, there's a 66-page gap between 9.1.4 and 9.1.5. The specification has plenty of 9.1.4.n subsections, but they don't show up in the TOC, because the "tocDepth" attribute isn't specified in the XML, and defaults to "3", so they won't be easily spotted unless you know what to search for in the text.
> 
> Perhaps you could consider adding "tocDepth=4" to

I thought the index looked a bit nicer without all the small details, but a 66 pages gap is indeed not nice at all.

Looking at the previous system we had before this PR it wasn't much better though
```
     9.4.  Tracks Element  . . . . . . . . . . . . . . . . . . . . .  44
       9.4.1.  TrackEntry Element  . . . . . . . . . . . . . . . . .  44
     9.5.  Cues Element  . . . . . . . . . . . . . . . . . . . . . . 109
       9.5.1.  CuePoint Element  . . . . . . . . . . . . . . . . . . 109
```

So I'd rather fix it in a separate PR to discuss that specific issue.

robUx4:
> While cross-checking the XML, I noticed that
> <title abbrev="Matroska Format">Matroska Media Container Format Specifications</title>
> 
> says that the specification is "informational", but the CELLAR charter at https://datatracker.ietf.org/wg/cellar/about/ says that the Matroska specifications would be Standards Track, so status="std".
> 
> Would it make sense to make that change at the same time?

Addressed in #512

SpencerDawkins:
@robUx4 - both your responses sound perfect. Thanks for considering my comments.

SpencerDawkins:
Merged during interim meeting

PR #496, opened 2021-04-22
Rewrite the lacing section with more normative text and precise examples

robUx4:
The lacing text was not very well defined and left a lot to interpretation. The proposed changes fix that by having independent sections for each type of lacing, including no lacing. An example is provided for each to show how to interpret the data.

As discussed in #355 lacing is now forbidden when a single frame is used.

The value 0 for `FlagLacing` is now explicitly defined as forbidding the use of lacing in all blocks of that track.

This is currently a draft as it sits on top of #495.


robUx4:
More editing of the Block/SimpleBlock sections will be needed (like the octet offsets), but that will come later.

PR #501, opened 2021-04-23
forbid non mandatory default values that are not 0

### LINKED ISSUE #500: Add an EBML constraint about 0-length elements
Although EBML allows an element with a default value to be written with just the ID and a length of 0, historically this is not properly supported by EBML parsers.
Recently we have started making element with a default value mandatory or not have a default value at all, to avoid this issue (ie making rules that will make legit Matroska files unreadable by current code) we should state that the 0-length feature is not supported in Matroska. The savings are minimal compared to a mandatory element not written.

This should be done after #495 which cleans the EBML sections at the beggining.

robUx4:
Due to legacy code we can't use Empty Elements with non zero defaults. We can't make something legal now that will break older parsers.

This makes some element temporarily invalid.

Fixes #500

Draft until #495 is merged.

### DIFF
--- a/index_matroska.md
+++ b/index_matroska.md
@@ -132,7 +132,7 @@ As an EBML Document Type, Matroska adds the following constraints to the EBML sp
 
 ## Matroska Design
 
-All top-levels elements (Segment and direct sub-elements) are coded on 4 octets -- i.e. class D elements.
+The Root Element and all Top-Levels Elements use 4 octets for their EBML Element ID -- i.e. Segment and direct children of Segment.
 
 # Language Codes
 
--- a/index_matroska.md
+++ b/index_matroska.md
@@ -135,12 +135,12 @@ As an EBML Document Type, Matroska adds the following constraints to the EBML sp
 The Root Element and all Top-Levels Elements use 4 octets for their EBML Element ID -- i.e. Segment and direct children of Segment.
 
 Matroska writers **MUST NOT** use  EBML Empty Elements, elements present in the file but with a length of zero,
-if the element has a default value is not 0 for integers/dates and 0x0p+0 for floats.
+if the element has a default value that is not 0 for integers/dates and 0x0p+0 for floats.
 This is to preserve compatibility with existing parsers that didn't interpret this EBML feature properly.
-Therefore any element that is not mandatory **SHOULD NOT** have a default value.
+Therefore any element that is not mandatory **SHOULD NOT** have a default value, so that Empty Elements cannot be used.
 
 A default value of 0 for integers/dates and 0x0p+0 for floats are tolerated as a zero length is correctly handled by legacy parsers.
-In this case an Empty Element is both interpreted as the default value or 0 for numbers.
+In this case, an Empty Element is both interpreted as the default value and 0 for numbers.
 
 # Language Codes
 
--- a/index_matroska.md
+++ b/index_matroska.md
@@ -134,13 +134,18 @@ As an EBML Document Type, Matroska adds the following constraints to the EBML sp
 
 The Root Element and all Top-Levels Elements use 4 octets for their EBML Element ID -- i.e. Segment and direct children of Segment.
 
-Matroska writers **MUST NOT** use  EBML Empty Elements, elements present in the file but with a length of zero,
+Matroska writers **MUST NOT** use EBML Empty Elements, elements present in the file but with a length of zero,
 if the element has a default value that is not 0 for integers/dates and 0x0p+0 for floats.
 This is to preserve compatibility with existing parsers that didn't interpret this EBML feature properly.
 Therefore any element that is not mandatory **SHOULD NOT** have a default value, so that Empty Elements cannot be used.
 
 A default value of 0 for integers/dates and 0x0p+0 for floats are tolerated as a zero length is correctly handled by legacy parsers.
 In this case, an Empty Element is both interpreted as the default value and 0 for numbers.
+
+Strings cannot have an empty string as default value, due to a limitation of the EBML Schema.
+So, based on the rules above, any strings that is not mandatory **SHOULD NOT** have a default value, so that Empty Elements cannot be used.
+And Matroska writers **MUST NOT** use EBML Empty Elements if the element is a string element.
+In that case the writer **MUST** either omit the element if it is mandatory and unique or write it with its value.
 
 # Language Codes
 
--- a/index_matroska.md
+++ b/index_matroska.md
@@ -134,6 +134,13 @@ As an EBML Document Type, Matroska adds the following constraints to the EBML sp
 
 The Root Element and all Top-Levels Elements use 4 octets for their EBML Element ID -- i.e. Segment and direct children of Segment.
 
+Matroska doesn't allow using EBML Empty Elements, elements present in the file but with a length of zero.
+This is to preserve compatibility with existing parsers that didn't interpret this EBML feature properly.
+Therefore any element that is not mandatory **SHOULD NOT** have a default value.
+
+A default value of 0 for integers/dates is tolerated as a zero length is correctly handled by legacy parsers.
+In this case an Empty Element is both interpreted as the default value or 0 for numbers.
+
 # Language Codes
 
 Matroska from version 1 through 3 uses language codes that can be either the 3 letters
--- a/index_matroska.md
+++ b/index_matroska.md
@@ -134,7 +134,8 @@ As an EBML Document Type, Matroska adds the following constraints to the EBML sp
 
 The Root Element and all Top-Levels Elements use 4 octets for their EBML Element ID -- i.e. Segment and direct children of Segment.
 
-Matroska doesn't allow using EBML Empty Elements, elements present in the file but with a length of zero.
+Matroska writers **MUST NOT** use  EBML Empty Elements, elements present in the file but with a length of zero,
+if the element has a default value is not 0 for integers/dates and 0x0p+0 for floats.
 This is to preserve compatibility with existing parsers that didn't interpret this EBML feature properly.
 Therefore any element that is not mandatory **SHOULD NOT** have a default value.
 
--- a/index_matroska.md
+++ b/index_matroska.md
@@ -138,7 +138,7 @@ Matroska doesn't allow using EBML Empty Elements, elements present in the file b
 This is to preserve compatibility with existing parsers that didn't interpret this EBML feature properly.
 Therefore any element that is not mandatory **SHOULD NOT** have a default value.
 
-A default value of 0 for integers/dates is tolerated as a zero length is correctly handled by legacy parsers.
+A default value of 0 for integers/dates and 0x0p+0 for floats are tolerated as a zero length is correctly handled by legacy parsers.
 In this case an Empty Element is both interpreted as the default value or 0 for numbers.
 
 # Language Codes
--- a/index_matroska.md
+++ b/index_matroska.md
@@ -134,18 +134,15 @@ As an EBML Document Type, Matroska adds the following constraints to the EBML sp
 
 The Root Element and all Top-Levels Elements use 4 octets for their EBML Element ID -- i.e. Segment and direct children of Segment.
 
-Matroska writers **MUST NOT** use EBML Empty Elements, elements present in the file but with a length of zero,
-if the element has a default value that is not 0 for integers/dates and 0x0p+0 for floats.
-This is to preserve compatibility with existing parsers that didn't interpret this EBML feature properly.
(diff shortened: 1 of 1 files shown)


index_matroska.md
```diff
@@ -136,7 +136,14 @@ As an EBML Document Type, Matroska adds the following constraints to the EBML sp
 
 ## Matroska Design
 
-All top-levels elements (Segment and direct sub-elements) are coded on 4 octets -- i.e. class D elements.
+The Root Element and all Top-Levels Elements use 4 octets for their EBML Element ID -- i.e. Segment and direct children of Segment.
+
+Matroska doesn't allow using EBML Empty Elements, elements present in the file but with a length of zero.
```
mbunkus, line 141: I'm a bit hesitant to forbid zero-length elements. I'm thinking mostly about user-provided data: chapters & tags. To me it seems quite possible that people might want to include a string-tag but not set a value yet. This wouldn't be allowed with this change.

hubblec4: > I'm thinking mostly about user-provided data: chapters & tags. To me it seems quite possible that people might want to include a string-tag but not set a value yet. This wouldn't be allowed with this change.

Very good catch.

robUx4: The wording is wrong, the goal is to forbid empty elements *with default value that is not zero/empty string*. User values like chapter names or tag values don't have a default value (even less non-zero).

robUx4: Actually that's a **SHOULD NOT** and not **MUST NOT** with the accepted cases being described in the next paragraph. It is missing strings though.

robUx4: And we do not have a way to express default values as "" for strings, so strings are excluded from Empty Elements not matter what.

robUx4: I turned this into a more specific MUST NOT

hubblec4: > the goal is to forbid empty elements with default value that is not zero/empty string.

I don't really understand why should this be banned.

If a reader reads a zero-length(empty) EBML-Integer element with a default value of 2, the value 2 is used and not the type default value 0.

robUx4: The reason is given in the next sentence

> This is to preserve compatibility with existing parsers that didn't interpret this feature properly

See #500

hubblec4: OK. But like you wrote

> This makes some element temporarily invalid.

And I ask me what is worse:
(maybe) a break of old (very old, no longer maintained) parsers
or make existing parameters invalid.

robUx4: It's neither nor.

As of today libebml has no fix (and no PR AFAIK) to fix this issue. It will likely involve some ABI changes that will take more time to implement and propagate on Linux distributions.

The only invalid parameter is `ChapterLanguage`, every thing is or has been cleaned. And for that field players already assume the default value no matter what.

hubblec4: > The only invalid parameter is ChapterLanguage,

The `ChapLanguage` element is not a non-mandatory element and so this is not the case for this element.
If the `ChapLanguage` element is not present or written with a zero-length of data then "eng" is used.

Here a list of non-mandatory not deprecated elements with a default value not equal to 0

* `MatrixCoefficients`
* `TransferCharacteristics`
* 'Primaries'
* 'CueBlockNumber'
* 'TargetTypeValue'

In this list is no String-element!

There are many non-mandatory elements with a default value of 0. All this elements must be written, and actually is the element header enough, because there is a default value which can be used if the data size is zero.
But when the zero-length feature is forbidden then the entire element must be written and the "default  atritubute" is more or less useless or can be omit.

> As of today libebml has no fix (and no PR AFAIK) to fix this issue.

Yes that's true, but such a fix is trivial from my humble point of view.
And when I understand the libebml code right then breaks this fix nothing.

robUx4: Added some more details, especially on strings.

* integer/float with default value = 0, Empty Element is allowed
* integer/float with default value != 0, Empty Element is NOT allowed
* string with default value = "", impossible in EBML Schema
* string with default value != "", Empty Element is NOT allowed

robUx4: For `MatrixCoefficients`, `TransferCharacteristics`, `Primaries`, `CueBlockNumber` and `TargetTypeValue`, they are in line with the rule 
> any element that is not mandatory **SHOULD NOT** have a default value

It's a "SHOULD NOT" rule, not a "MUST NOT". Meaning in some case they might have a default value. Nevertheless they won't be using Empty Elements if the spec is followed

> Matroska writers **MUST NOT** use EBML Empty Elements, elements present in the file but with a length of zero,
if the element has a default value that is not 0 for integers/dates and 0x0p+0 for floats.

robUx4:
https://www.w3.org/ is currently returning an error 503, so the build check is failing...

mkver:
I do not see why the current inability of several readers to handle empty elements with default values (other than zero) should mean that the whole feature should be abandoned. The proposed changes signal to anyone implementing a new EBML/Matroska parser that this EBML feature is unused and will be unused in Matroska, so that there is no incentive to support this, even when it would be easy to implement at an early stage of the design.

And this sentence is particularly weird: "Therefore any element that is not mandatory **SHOULD NOT** have a default value." I always thought that the intended audience for the spec are prospective implementators of said spec. Yet this sentence is not for them; it is for the people writing the spec.

My solution for this mess would be as follows:

1. Forbid writing mandatory elements with default value as empty elements (maybe with an exception if said default value is zero). It makes no sense to write them that way anyway.
2. Bring all the current elements in line with how they have been used the whole time (i.e. make elements with default values mandatory).
3.  Don't create non-mandatory elements with default value as long as parsers are unable to understand them. (Luckily, given that old parsers have to skip new elements anyway we do not need to think of them when deciding whether this feature can now be safely used. It will be enough if the newest versions of all tools support it.)

The first of these points is directed at implementators (and should be added to the spec) and the other two at the spec writers.

robUx4:
> so that there is no incentive to support this

Indeed, in the current state of Matroska parsers, it won't work. A future spec adding new elements could reevaluate this and consider the rule is not necessary anymore. It's possible for new elements in the distant future where parsers have been fixed. On the other hand, it may also encourage new implementations not to bother supporting this EBML rule properly.

> thought that the intended audience for the spec are prospective implementators of said spec. Yet this sentence is not for them; it is for the people writing the spec.

The spec also has to set rules on how the format can be expanded, so current readers can read future files. We could also set some guidelines on how to pick IDs which is useless for implementors.

#1 This is proposed in this PR 
> Matroska writers **MUST NOT** use EBML Empty Elements, elements present in the file but with a length of zero,
if the element has a default value that is not 0 for integers/dates and 0x0p+0 for floats.

#2 there are still elements that have a default value without being mandatory:
* `TrackEntry\CodecDelay`
* `TrackEntry\Video\StereoMode`
* `TrackEntry\Video\AlphaMode`
* `TrackEntry\Video\PixelCropBottom`
* `TrackEntry\Video\PixelCropTop`
* `TrackEntry\Video\PixelCropLeft`
* `TrackEntry\Video\PixelCropRight`
* `TrackEntry\Video\DisplayUnit`
* `TrackEntry\Video\AspectRatioType`
* `TrackEntry\Video\Colour\MatrixCoefficients` (default = 2 = unspecified)
* `TrackEntry\Video\Colour\BitsPerChannel`
* `TrackEntry\Video\Colour\ChromaSitingHorz`
* `TrackEntry\Video\Colour\ChromaSitingVert`
* `TrackEntry\Video\Colour\Range`
* `TrackEntry\Video\Colour\TransferCharacteristics` (default = 2 = unspecified)
* `TrackEntry\Video\Colour\Primaries` (default = 2 = unspecified)
* `ContentEncoding\ContentEncryption\ContentSigAlgo`
* `ContentEncoding\ContentEncryption\ContentSigHashAlgo`
* `CuePoint\CueTrackPositions\CueCodecState`
* `Chapters\EditionEntry\EditionFlagOrdered`
* `Tag\Targets\TargetTypeValue` (default = 50)
* `Tag\Targets\TagTrackUID`
* `Tag\Targets\TagEditionUID`
* `Tag\Targets\TagChapterUID`
* `Tag\Targets\TagAttachmentUID`

Unless, specified, the default value is 0 for all the elements. The deprecated elements (maxver="0") are not included.

I think they can become mandatory and it won't make a difference (in all files they are likely written and, if not, the value is assumed to be that default value anyway). Even `Tag\Targets\TagXXXUID` elements which mean 0 applies to all XXX elements in the segment (if not set it also applies to the all elements anyway).

Some element have non-zero default value, but it means `unspecified`, which is similar to an absent element in EBML/Matroska. So whether we write a "unspecified" value or don't write a value that is interpreted as "unspecified" (if default) it makes no difference.

#3 is trickier, in particular with libebml/libmatroska. Adding elements in libmatroska is easy (a few lines added), but that doesn't mean the default value would be interpreted correctly by libeml.

IMO it makes more sense to have default values mean "unspecified"/"unused", and thus allowing elements to be default, than having elements be interpreted differently if they are present or not (the only case where using Empty Elements has any real use). This is actually why all the elements above can be easily switched to default, because their default value really represent their default/natural state. If it's not possible, they just don't have a default value.

robUx4:
(rebased on master where `CueBlockNumber` doesn't have a default value anymore)

robUx4:
Oops, it turns out `TrackEntry\Video\AlphaMode` doesn't work like that :

>Presence of this Element indicates that the BlockAdditional Element could contain Alpha data

This comes from WebM. IMO it should not have a default value. In the [WebM spec](https://www.webmproject.org/docs/container/#AlphaMode), there's no default value mentioned. But technically writing 0 in the file means there's alpha data expected (in BlockAdditional). An Empty Element would also do the same. All known implementations would not have a problem with that because it's 0.

That's the only non-mandatory, with a default value, element that can't be made mandatory.

robUx4:
BTW, `AlphaMode` shows why it's important to have rules how to create new elements. The presence of a flag with no value or 0 meaning the same as 1 or any other value is not right.

robUx4:
Only these elements remain problematic:
* `Tag\Targets\TagTrackUID`
* `Tag\Targets\TagEditionUID`
* `Tag\Targets\TagChapterUID`
* `Tag\Targets\TagAttachmentUID`

Making this PR a Draft until #569 is merged/solved.

robUx4:
Actually all `Tag\Targets\TagxxxUID` elements have a default value of 0. So they are tolerated to be stored as Empty Element, to mark their presence with their default value. The rules layed out in this PR still work regardless of the outcome of #569 (become mandatory or not)

index_matroska.md
```diff
@@ -132,7 +132,20 @@ As an EBML Document Type, Matroska adds the following constraints to the EBML sp
...
+if the element has a default value that is not 0 for integers/dates and 0x0p+0 for floats.
+This is to preserve compatibility with existing parsers that didn't interpret this EBML feature properly.
+Therefore any element that is not mandatory **SHOULD NOT** have a default value, so that Empty Elements cannot be used.
+
+A default value of 0 for integers/dates and 0x0p+0 for floats are tolerated as a zero length is correctly handled by legacy parsers.
+In this case, an Empty Element is both interpreted as the default value and 0 for numbers.
+
+Strings cannot have an empty string as default value, due to a limitation of the EBML Schema.
+So, based on the rules above, any strings that is not mandatory **SHOULD NOT** have a default value, so that Empty Elements cannot be used.
+And Matroska writers **MUST NOT** use EBML Empty Elements if the element is a string element.
```
robUx4, line 147: This is wrong. An empty element for strings is OK. It means an empty string:

> If an Empty Element has no default value declared, then the EBML Reader
   MUST use the value of the Empty Element for the corresponding EBML
   Element Type of the Element ID, 0 for numbers and an empty string for
   strings.

index_matroska.md
```diff
@@ -132,7 +132,20 @@ As an EBML Document Type, Matroska adds the following constraints to the EBML sp
...
+In that case the writer **MUST** either omit the element if it is mandatory and unique or write it with its value.
```
robUx4, line 148: > write it with its value

We can't do that with a string that is legitimately empty, ie `""`.

robUx4 (commented):
We need to rework the Strings part, it's not THAT bad.

index_matroska.md
robUx4: This is similar to [what Moritz said](https://github.com/ietf-wg-cellar/matroska-specification/pull/501#discussion_r642107231).

robUx4:
Reworked the paragraph:

* Empty Elements are OK as long as the default value is 0
* Strings can be Empty Elements if they are not mandatory (and if they are mandatory their default value can't be `""` by design, so fall in the rule above)
* non-mandatory Elements can be no default value or a 0 default value (so for strings only no default is possible)

PR #502, opened 2021-04-23
make some Video\Colour elements mandatory

### LINKED ISSUE #310: Non-mandatory elements with default values
I have just made a surprising observation: The `Language` (`TrackLanguage`) element isn't mandatory, although it has a default value (namely "eng") and although most tools (including MKVToolNix and the foundation-source tools and ffmpeg/libavformat) treat it as if this element were mandatory with default value "eng". Given the abundance of files that were created with this interpretation, the specs should be changed to really make the element mandatory.

And then I looked a bit deeper into this topic and found that there are 41 non-mandatory elements with default values:

- Elements that IMO should be made mandatory (7): Language, PixelCrop* (4x) (If these values aren't mandatory, then the PixelCrop* values are completely indetermined. But they may still be used for the evaluation of the default value of DisplayWidth/Height, so they need to be mandatory.), DisplayUnit (mkvmerge, ffmpeg and mkclean treat files as if DisplayUnit were mandatory with default value 0), ContentEncAlgo (see #303),
- Deprecated and irrelevant elements (9): LaceNumber, FrameNumber, BlockAdditionID, Delay, SliceDuration, TrackOffset, TrickTrackFlag, CueRefNumber, CueRefCodecState
- Special elements (3): BlockDuration: Whether it is mandatory depends upon other elements (whether the track has a `DefaultDuration`). This looks like a violation of the principle of separation of layers (EBML and Matroska) to me; but this is unavoidable now.
  DisplayWidth/Height: The description here says that the default value only applies if DisplayUnit is 0. But given that these elements are not mandatory, the only way for this default value to apply would be if there were empty DisplayWidth/Height elements present in the file (and if DisplayUnit were 0). But the common tools (mkvmerge, mkclean, ffmpeg) all treat DisplayWidth/Height as mandatory (at least if DisplayUnit == 0). Therefore I think that one should make this element mandatory with its current default value if DisplayUnit == 0. (Yes, this breaks the separation between EBML and Matroska, too.)
- The rest (22): CodecDelay, Stereomode, Alphamode, AspectRatioType, OutputSamplingFrequency, MatrixCoefficients, BitsPerChannel, ChromaSitingHorz, ChromaSitingVert, Range, TransferCharacteristics, Primaries, ContentSigAlgo, ContentSigHashAlgo, CueBlockNumber, CueCodecState, EditionFlagOrdered, TargetTypeValue, TagTrackUID, TagEditionUID, TagChapterUID, TagAttachmentUID

robUx4:
They have a default value set to "undefined". Some have a non zero default value which should not be used (#500). So we change all of them at once.

This semantic change should have no impact on reading or writing since a non presence should already be assumed to be undefined. And no knowledge from the source should either write undefined or write nothing.

Partially fixes #310

### DIFF
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -626,7 +626,7 @@ This is additive with ChromaSubsamplingVert.</documentation>
     <extension type="webmproject.org" webm="1"/>
     <extension type="libmatroska" cppname="VideoCbSubsampVert"/>
   </element>
-  <element name="ChromaSitingHorz" path="\Segment\Tracks\TrackEntry\Video\Colour\ChromaSitingHorz" id="0x55B7" type="uinteger" minver="4" default="0" maxOccurs="1">
+  <element name="ChromaSitingHorz" path="\Segment\Tracks\TrackEntry\Video\Colour\ChromaSitingHorz" id="0x55B7" type="uinteger" minver="4" default="0" minOccurs="1" maxOccurs="1">
     <documentation lang="en" purpose="definition">How chroma is subsampled horizontally.</documentation>
     <restriction>
       <enum value="0" label="unspecified"/>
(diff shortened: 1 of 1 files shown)


mbunkus (approved):

hubblec4 (approved):
LGTM

retokromer (approved):

Issue #419, opened 2020-08-29
Specify requirements around ReferenceBlock elements and how BlockGroups signal random access points


mjbshaw:
The [W3C WebM spec states](https://w3c.github.io/media-source/webm-byte-stream-format.html#webm-random-access-points):

> Either a SimpleBlock element with its Keyframe flag set, or a BlockGroup element having no ReferenceBlock elements, signals the location of a random access point for that track. The order of multiplexed blocks within a media segment MUST conform to the WebM Muxer Guidelines.

It's possible I've overlooked this in the MKV specs, but I believe the following is missing:

- Are ReferenceBlock elements mandatory for non-RAPs? Are BlockGroup elements required to specify *all* references to other blocks? Or just some? Or none?
- Does MKV agree with WebM in that a lack of ReferenceBlock elements signals that a BlockGroup is a RAP?
- If so, how should a non-RAP I-frame be muxed? Would you need to add a fake ReferenceBlock element to signal it's not a RAP (e.g., a ReferenceBlock referring to itself)?
- How should rolling intra-refresh frames be muxed in a BlockGroup? These can be treated as RAPs (assuming you drop enough frames), but they can also have dependencies on other frames. If ReferenceBlock is used to report all its dependencies then how can the player know it's a RAP?
- Similar to the above, what is the guidance for audio, e.g. Opus? Any block in an audio stream is effectively a RAP, though there's the seek pre-roll you need to take into account. I don't think it really makes sense for an audio block to use ReferenceBlock elements, even if the block technically builds upon prior blocks. Can we specify which types of tracks should/shouldn't use ReferenceBlock elements?

robUx4:
>     * Are ReferenceBlock elements mandatory for non-RAPs? Are BlockGroup elements required to specify _all_ references to other blocks? Or just some? Or none?

ReferenceBlock is not mandatory in any case.
BlockGroups are not required to reference any other Block/SimpleBlock. I suppose they usually do. An analysis of existing files might give a better idea of the common practice. And if they usually do that's probably a rule we want to add.

>     * Does MKV agree with WebM in that a lack of ReferenceBlock elements signals that a BlockGroup is a RAP?

The ReferenceBlock means than the reference Block/SimpleBlock is needed to decode this BlockGroup, nothing more. That Block may itself depend on another Block, and so on.

[...]

PR #503, opened 2021-04-23
Random Access Point clarifications

robUx4:
Following the discussions from #419, we need some explanations on RAP.

Historically the keyframe flag was not always set for audio blocks even though each audio (Simple)Block could be seeked to. So there's some room in the rules for that.

In the thread we also proposed a new flag (in the Simple|Block or in the Track) to specify how strict it is to the rule. With such a flag we could make rules stricter for future files as long as they use that flag.

Another rule that may not correctly apply to existing files is the fact the `ReferenceBlock` must list all the frames it references. I'm not sure this is the case. We should study what mkvmerge/libavformat/libwebm do.

This patches sits on top of #495 so it's a draft for now.


robUx4:
* [libavformat](https://github.com/FFmpeg/FFmpeg/blob/master/libavformat/matroskaenc.c#L2119) is only writing one `ReferenceBlock` and it does not always use the proper value:
```c
    if (blockid == MATROSKA_ID_BLOCK && !keyframe)
        put_ebml_sint(pb, MATROSKA_ID_BLOCKREFERENCE, track->last_timestamp - ts);
    track->last_timestamp = ts;
```
The referenced frame may not always the previous one. And for a B-frame there are 2 frames to reference.

* libwebm also only [writes a single reference](https://aomedia.googlesource.com/aom/+/refs/heads/master/third_party/libwebm/mkvmuxer/mkvmuxerutil.cc#139), but that is normal since VP8 and VP9 don't have B-frames. It also gets the [actual reference frame timestamp](https://aomedia.googlesource.com/aom/+/refs/heads/master/third_party/libwebm/mkvmuxer/mkvmuxerutil.cc#69), but it seems to write as an absolute value instead of a relative value.

* mkvmerge seems to [reference the proper frames](https://gitlab.com/mbunkus/mkvtoolnix/-/blob/main/src/merge/libmatroska_extensions.cpp#L53) and it seems to [store them as relative timestamps](https://gitlab.com/mbunkus/mkvtoolnix/-/blob/main/src/merge/cluster_helper.cpp#L500).

robUx4:
This is actually a duplicate of #555 which more comments and more up to date text...

PR #504, opened 2021-05-02
Clean chapterend section

### LINKED ISSUE #478: fix ChapterTimeEnd section
It's currently giving examples on how some chapters end up with a duration of 0, but we have now established it's illegal:
`The value **MUST** be strictly greater than the ChapterTimeStart of the same ChapterAtom`

robUx4:
The text is inconsistent with the `ChapterTimeEnd` definition in the EBML Schema.

Fixes #478.

Draft until #495 is merged.

### DIFF
--- a/chapters.md
+++ b/chapters.md
@@ -134,7 +134,6 @@ For `Simple Chapters` this is the position of the chapter markers in the timelin
 ### ChapterTimeEnd
 The timestamp of the end of `Chapter` with nanosecond accuracy, not scaled by TimestampScale.
 The timestamp defined by the `ChapterTimeEnd` is not part of the `Chapter`.
-The end timestamp is used when the `EditionFlagOrdered` flag of the `Edition` is set to `true`.
 A `Matroska Player` calculates the duration of this `Chapter` using the difference between the
 `ChapterTimeEnd` and `ChapterTimeStart`.
 The end timestamp **MUST** be strictly greater than the start timestamp.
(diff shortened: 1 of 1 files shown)


chapters.md
```diff
@@ -128,25 +128,23 @@ The `ChapterAtom` is also called a `Chapter`.
...
-A not scaled timestamp of the start of `Chapter` with nanosecond accuracy.
+The timestamp of the start of `Chapter` with nanosecond accuracy, not scaled by TimestampScale.
 For `Simple Chapters` this is the position of the chapter markers in the timeline.
 
 ### ChapterTimeEnd
-A not scaled timestamp of the end of `Chapter` with nanosecond accuracy.
-The end timestamp is used when the `EditionFlagOrdered` flag of the `Edition` is set to `true`.
+The timestamp of the end of `Chapter` with nanosecond accuracy, not scaled by TimestampScale.
 The timestamp defined by the `ChapterTimeEnd` is not part of the `Chapter`.
+The end timestamp is used when the `EditionFlagOrdered` flag of the `Edition` is set to `true`.
```
hubblec4, line 137: On the one hand, you have "fought" to ensure that the `ChapterTimeEnd` is used(can be used) even without an ordered edition.
 #440
 removing the `ChapterEndTime` from the EditionFlagOrdered table.

And now is the element used if the `EditionFlagOrdered` flag of the `Edition` is set to `true`.

What is now correct?

robUx4: This is not a new sentence I added, only the text around has been moved.
I agree there is a contradiction here with #440 and in that case #440 is right.
I'll remove the sentence entirely.

mbunkus (approved):

hubblec4 (approved):

retokromer (approved):

PR #510, opened 2021-05-09
Add the 0x44B4 bogus TagDefault ID in the XSD

### LINKED ISSUE #508: Add the bogus TagDefault value in the RFC
`TagDefault` has an EBML ID `0x4484`, but some files exist with a value of `0x44B4`, referred as `MATROSKA_ID_TAGDEFAULT_BUG` in libavcodec.

We should probably mention this value in the specs. Probably in the section of deprecated elements that is introduced by #487.

robUx4:
Since we're supposed to describe existing files, this value is found in the wild.
It's parsed by libavformat as well.

The element is marked as deprecated since version 0, so should not be used.

Fixes #508


robUx4:
This should be merged after #487 which makes these elements informative rather than normative.

mbunkus (approved):

hubblec4 (approved):

retokromer (approved):

PR #511, opened 2021-05-09
move back the control track IDs in the main schema file

### LINKED ISSUE #509: add the control track elements in main Schema file
The Schema file we maintain here has more information (via extensions) than the actual EBML Schema that is used to generate the RFC text and is following strictly the EBML Schema standard. It is transformed to that file using [schema_clean.xsl](https://github.com/ietf-wg-cellar/matroska-specification/blob/master/transforms/schema_clean.xsl).

Our EBML Schema is used to generate code, so it's better if it includes all the information in one file. Just like we have DivX elements, for example. Since #461 some of the elements that were generating code are not found in that file anymore. We could put them back marking them with an extension, and filter them out in `schema_clean.xsl`.

robUx4:
It's easier to have all the information in one file.
We just exclude elements with the "other document" extension from the main RFC.

Fixes #509


ebml_matroska.xml
```diff
@@ -1340,4 +1340,25 @@ If this Element is used, then any TagLanguage Elements used in the same SimpleTa
     <documentation lang="en" purpose="definition">The values of the Tag, if it is binary. Note that this cannot be used in the same SimpleTag as TagString.</documentation>
     <extension type="webmproject.org" webm="1"/>
   </element>
+  <element name="EditionFlagHidden" path="\Segment\Chapters\EditionEntry\EditionFlagHidden" id="0x45BD" type="uinteger" range="0-1" default="0" minOccurs="1" maxOccurs="1">
```
hubblec4, line 1343: Why is this section added to the end?
Wouldn't it be better to add this to the chapter section?

robUx4: I can do that

hubblec4 (approved):
LGTM

robUx4:
Merging this so we can use the Schema file to generate code again

hubblec4:
Yes please merge.
For my own EBML and Matroska project I use this xml files to generate code.

PR #512, opened 2021-05-23
mark the markdown/xml/output as IETF standards

robUx4:
We don't have the RFC numbers yet but that's the goal for these documents.

Fixes issue raised in https://github.com/ietf-wg-cellar/matroska-specification/pull/495#issuecomment-846230313


SpencerDawkins:
Merged on the call.

PR #513, opened 2021-05-23
Expand the tocdepth to 4

robUx4:
The default `tocDepth` is 3. 4 might be a bit verbose but for the Matroska Elements that gives a better granularity.

The issue is mostly with `Tracks` + `TrackEntry` that needs to use finer granularity when 3 is used. Maybe merging these levels into one using XSLT might give nicer results without forcing everything else to have a `tocDepth` of 4 and give too many details in the index/Table Of Content.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,8 +70,7 @@ $(OUTPUT_CONTROL).md: index_control.md control.md control_elements4rfc.md menu.m
 	             -e "s/@BUILD_VERSION@/$(OUTPUT_CONTROL)/" > $@
 
 %.xml: %.md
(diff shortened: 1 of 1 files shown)


robUx4:
This changes also affects the other RFC documents we're working on (tags, codec and possibly control tracks). The impact on those is minimal (and no change for tags).

In the end the change in the main Matroska document does add some granularity in other places where it's welcome. There are some occasional subsections that only contain one item (`SilentTrackNumber Element`, `Ordered-Edition and Matroska Segment-Linking` and `The Micronauts "Bleep To Bleep"`). These can be fixed by not using a sub section in the text.

`SilentTracks` are a feature that is likely not used anywhere, we may deprecate it (only to bring it back later if needed).

**Conclusion**: this PR solves the big gap in the TOC correctly.

mbunkus (approved):

retokromer (approved):

PR #514, opened 2021-05-30
make ChapterDisplay use only one language at a time

robUx4:
Following the discussion in https://github.com/ietf-wg-cellar/ebml-specification/issues/409
It could be assumed that English is always present because ChapLanguage is
mandatory and has a default value.

To avoid this interpretation we should have one language (and one IETF variant)
per ChapterDisplay item. We already have to use many ones per language. The merge
of different values because they have the same value is unlikely (but it may
exist in the real world).

### DIFF
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -1165,14 +1165,14 @@ see (#physical-types) for a complete list of values.</documentation>
     <extension type="webmproject.org" webm="1"/>
     <extension type="libmatroska" cppname="ChapterString"/>
   </element>
-  <element name="ChapLanguage" path="\Segment\Chapters\EditionEntry\+ChapterAtom\ChapterDisplay\ChapLanguage" id="0x437C" type="string" default="eng" minOccurs="1">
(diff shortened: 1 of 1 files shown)


robUx4:
libavformat doesn't use the element either [when reading](https://github.com/FFmpeg/FFmpeg/blob/master/libavformat/matroskadec.c#L626) and always [writes a single `"und"` value](https://github.com/FFmpeg/FFmpeg/blob/master/libavformat/matroskaenc.c#L1687).

libwebm [reads a single value](https://aomedia.googlesource.com/aom/+/refs/heads/master/third_party/libwebm/mkvparser/mkvparser.cc#3665) (possibly leaking data if more than one value is found) and [writes a single value](https://aomedia.googlesource.com/aom/+/refs/heads/master/third_party/libwebm/mkvmuxer/mkvmuxer.cc#2072).

hubblec4:
No this is not a bug and I tottaly disagree with this change.

mbunkus (changes requested):
I've made my objections known in ietf-wg-cellar/ebml-specification#409 where we should keep the discussion about what to do until we reach consensus, I think.

robUx4:
OK with allowing multiple values although some clarifications are needed on the EBML side.

PR #515, opened 2021-05-30
Remove range attribute from element with a (large) restriction section

robUx4:
The restriction mentions the value that are allowed. The range gives the lower
and upper limits. But without an indication on what to do with values in the
range but with an unknown value we have more confusion than necessary.

If the restriction values are later expanded older parser will ignore the new
values no matter what. The range is only useful to tell if the value is legit
or not, but there's still plenty of changes it is not.

Somehow fixes the issue raised in https://github.com/ietf-wg-cellar/ebml-specification/issues/410

### DIFF
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -244,7 +244,7 @@ though the design allows an unlimited number).</documentation>
     <documentation lang="en" purpose="definition">A unique ID to identify the Track.
 This **SHOULD** be kept the same when making a direct stream copy of the Track to another file.</documentation>
   </element>
-  <element name="TrackType" path="\Segment\Tracks\TrackEntry\TrackType" id="0x83" type="uinteger" range="1-254" minOccurs="1" maxOccurs="1">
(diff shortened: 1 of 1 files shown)


hubblec4:
Yes we should remove the range attribute if a restriction is present.
We have more then these two elements: here a list:
`TrackType`
`FlagInterlaced`: the range matches the restriction but should also removed
'FieldOrder'
`ProjectionType`: the range matches the restriction but should also removed
`ContentEncodingScope`

mbunkus (changes requested):
I agree that the ranges should go if restrictions exist.

Please extend the patch to cover all such elements. I've looked through `ebml_matroska.xml`, too, and the list by @hubblec4 looks to be complete.

@hubblec4 I've edited your comment slightly for improved formatting only.

robUx4:
Done

hubblec4 (approved):

retokromer (approved):

PR #521, opened 2021-06-27
Rewrite timestamp intepretations using ticks

### LINKED ISSUE #517: rewrite general timestamp handling sections
The [TimestampScale Rounding](https://github.com/ietf-wg-cellar/matroska-specification/blob/master/notes.md#timestampscale-rounding) and [TrackTimestampScale](https://github.com/ietf-wg-cellar/matroska-specification/blob/master/notes.md#tracktimestampscale) sections provide somehow outdated explanations. If the goal is to have more precision on samples, TrackTimestampScale have to be used differently, as explained in #422.

WebM doesn't support `TrackTimestampScale` and a lot of readers don't use it either, so we should cover that case as well and explain the limits.

robUx4:
Draft: the Block, TrackTimestampScale are not rewritten yet

The notion of Matroska/Segment/Track ticks is introduced to make it easier to know what formula to apply. For example `Cluster\Timestamp` is using the `TimestampScale` but **NOT** the `TrackTimestampScale`.

The `Info\Duration` is expressed in Segment Ticks, but it's in floating point which is odd, only Track Ticks may be in floating point due to  `TrackTimestampScale`.

Fixes #517

### DIFF
--- a/notes.md
+++ b/notes.md
@@ -302,6 +302,7 @@ The elements storing values in Segment Ticks are:
 
 * `Cluster\Timestamp`; defined in (#timestamp-element)
 * `Info\Duration` is stored as a floating point but the same formula applies; defined in (#duration-element)
+* `CuePoint\CueTrackPositions\CueDuration`; defined in (#cueduration-element)
 
 ### Track Ticks
 
--- a/notes.md
+++ b/notes.md
@@ -358,88 +358,12 @@ During playback, when a frame has a negative timestamp, the content **MUST** be 
 ## TimestampScale Rounding
 
 The default Track Tick duration is one millisecond.
-When dealing with audio, this causes inaccuracy in timestamps.
 
-For instance, when storing a whole CD in a single track, a user will want to be able to seek
-to the exact sample that a song begins at. If seeking a few sample ahead or behind, a crack
-or pop may result as a few odd samples are rendered. Also, when performing precise editing,
-it may be useful to have the audio accuracy down to a single sample.
-
-When storing timestamps for an audio stream, the `TimestampScale Element` **SHOULD** have an accuracy
-of at least that of the audio sample rate, to minimize rounding errors that prevent users
-from knowing the precise location of a sample. Here's how a program **SHOULD** round each timestamp
-in order to be able to recreate the sample number accurately.
-
-Let's assume that the application has an audio track with a sample rate of 44100 Hz. As written
-above the `TimestampScale` **MUST** have at least the accuracy of the sample rate itself: 1,000,000,000 / 44100 = 22675.7369614512.
-This value **MUST** be rounded since `TimestampScale` is an integer.
-So in this example the application will use either "22675" or "22676" for the `TimestampScale`.
-
-Over time the rounding of the sampling period in nanosecond will accumulate,
-both in the `Cluster\Timestamp` and the Block/SimpleBlock timestamps which are both in Segment Tricks,
-assuming the `TrackTimestampScale` is kept at its default value of "1.0".
-
-To avoid this drift, the muxer **SHOULD** take in account how the `Matroska Reader` is going to compute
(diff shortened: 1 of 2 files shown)


robUx4:
Still missing the rewrite of the `TrackTimestampScale` section.

Using more of the Segment/Track Tick unit to make things easier to explain.

robUx4:
Ready for review: the `TrackTimestampScale` is removed as the default usage was wrong. The proper text will come in #437.

robUx4:
Cleaned up so all elements refer to a tick. We could remove the Matroska Ticks, just nanoseconds, to make this a bit lighter.

Depending if we chose to [keep `TrackTimestampScale` or not](https://github.com/ietf-wg-cellar/matroska-specification/pull/437), Track Ticks can be merged in Segment Ticks or not. In that case it's easier to remove Matroska Ticks and just keep the Segment Tick. (but I'm in favor or keeping `TrackTimestampScale` for good)

robUx4:
Rebased and fixed a sentence. This seems mergeable.

robUx4:
And remove the section that explained how to get finer frame precision as it's not the way we're going to do it. (Ref #422 )

notes.md
```diff
@@ -265,6 +265,36 @@ was called Timecode, the `TimestampScale Element` was called TimecodeScale, the
 `TrackTimestampScale Element` was called TrackTimecodeScale and the
 `ReferenceTimestamp Element` was called ReferenceTimeCode.
 
+## Raw And Scaled Timestamps
+
+There are 2 types of Timestamps in Matroska, the raw ones stored in nanoseconds and the scaled timestamps.
+
+The elements with raw timetamps in nanoseconds are:
+
+* `Cluster\Timestamp`; defined in (#timestamp-element)
```
mkver: This is not in nanoseconds at all. The same goes for CueTime below (and probably also for CueRefTime).

notes.md
```diff
@@ -328,23 +328,20 @@ So using a value other than "1.0" **MAY** not work in many places.
...
-The `Block`'s timestamp has a range of -32768 to +32767 units. When using the default value
-of the `TimestampScale Element`, each integer represents 1ms. The maximum time span of
-`Block Elements` in a `Cluster` using the default `TimestampScale Element` of 1ms is 65536ms.
+The `Block Element` and `SimpleBlock Element` store their timestamps as signed integers, relative
+to the `Cluster\Timestamp` value of the `Cluster` they are stored in.
+To get the timestamp of a `Block` or `SimpleBlock` in nanoseconds you have to use the following formula:
 
-If a `Cluster`'s `Timestamp Element` is set to zero, it is possible to have `Block Elements`
-with a negative `Raw Timestamp`. `Block Elements` with a negative `Raw Timestamp` are not valid.
+    Cluster\Timestamp + (signed timestamp * TimestampScale * TrackTimestampScale)
```
mkver: This formula is wrong, as Cluster\Timestamp is not in ns; so the real formula is (Cluster\Timestamp + signed timestamp) * TimestampScale * TrackTimestampScale.

notes.md
robUx4: You're right. The XML text is correct but this is not.

Nice catch !

robUx4: This was already fixed by a95611e50cd5a847d3faca393f437319a9314139 "merge Raw And Scaled Timestamps and Timestamp Ticks sections".

notes.md
robUx4: This one was fixed by e9a094248c7d12326986bd8d51548b2cf051b3c3 "rewrite the TimestampScale Rounding audio-only example". But the `TrackTimestampScale` doesn't apply to the `Cluster\Timestamp` value, only to the timestamp of the track, so the one found in blocks of that track.

So the formula is
```
(Cluster\Timestamp * TimestampScale) + (signed timestamp * TimestampScale * TrackTimestampScale)
```

PR #529, opened 2021-08-08
register MIME types in the IANA considerations

### LINKED ISSUE #525: Make the MIME types official
We do have MIME types mentioned in the document but they are mentioned as `not IETF endorsed`. Since we are not doing an IETF document, either we don't mention them or we make it official.

Following [this thread in the Mailing List](https://mailarchive.ietf.org/arch/msg/cellar/1-4Wwn7dL47sOOTkOGP0rOYd0c0/).

robUx4:
Names with the "x-" suffix should not be used according to RFC 6648, but
if it's the historical name it should be OK according to Appendix A of RFC 6838.

I'm not sure after 19 years adding video/matroska and audio/matroska will have
any significant use/impact. And it would be odd to make stereoscopic files (unlikely ever used) stand with a clean name.

Fixes #525


SpencerDawkins:
I might have opinions about this resolution, and they might even be correct, but they aren't well informed, so let's ignore that for now. 

But I DO try to be helpful, even if I don't understand everything that's going on. 

According to [the IANA Media Types registry](https://www.iana.org/assignments/media-types/media-types.xhtml), some guy named @mskucherawy is a backup reviewer for Media Types, and he's also the AD who would be doing initial AD review on the Matroska drafts. @mcr, do we need to ask him about this resolution? 

As part of IESG Evaluation, IANA is asked to review all documents for possible IANA actions, and I usually try to ask for early IANA review when registries are under Expert Review registration procedures, but since this resolution is an "IANA **non-action**",  I think it would be especially good to have that conversation directly with Murray before we get anywhere close to requesting publication.

robUx4:
Updated to reflect the conversation with [Murray S. Kucherawy on the Mailing List](https://mailarchive.ietf.org/arch/msg/cellar/w9Xx1OKD9JbyysfZNSxaSE99GSA/).

We keep the old names for legacy and prefer the new clean names.

SpencerDawkins:
Ready to merge? Please review and comment!

iana.md
```diff
@@ -4,6 +4,26 @@
...
 
+## MIME Types
+
+Matroska files and streams are found in three main forms: audio-video files, audio-only and occasionally with stereoscopic video tracks.
+
+The MIME types to use for each type is:
+
+* "video/matroska" for streams containing video tracks
+* "audio/matroska" for streams containing audio tracks with no video tracks
+* "video/matroska-3d" for streams containing some stereoscopic video tracks
```
retokromer, line 15: Is `some` necessary? (Also below.)

robUx4: `any` would work too. It also works with and without `some`. I have no preference.

If we want stricter wording we should say that the MIME types `SHOULD` or `SHOULD NOT` be used if there are(not) audio/stereoscopic tracks.

Then there's the case where the file has no audio or video (pure chapter files, as is possible with DVD "images"). It should be `video/matroska` but then it doesn't contain any video.

PR #530, opened 2021-08-08
Add some encryption/compression links

robUx4:
#423 will need to be updated with a similar link/explanation


robUx4:
Maybe mention not to use LZO as there's only GPL code to decode it.

JeromeMartinez (approved):

JeromeMartinez:
> Maybe mention not to use LZO as there's only GPL code to decode it.

As it is a different topic, I think it is not blocking this PR and I created a specific issue https://github.com/ietf-wg-cellar/matroska-specification/issues/538.

retokromer (approved):

PR #532, opened 2021-08-15
rename the "restrictions" field to "possible values"

robUx4:
There is no 'restriction' defined but just explanation on each values.

The generic field in the EBML is named that way, but that doesn't mean the generated
text should use that.

All enums should have definitions for each value anyway.

Alternatively "standard values" might be used instead of "possible values". Other value may be possible in the future.


retokromer:
>Alternatively "standard values" might be used instead of "possible values".

I am not an English speaker, but `valid` come to mind.

robUx4:
> > Alternatively "standard values" might be used instead of "possible values".
> 
> I am not an English speaker, but `valid` come to mind.

That would seem to imply that any other value is invalid, while the goal is to allow possible additions in the future.

retokromer:
>That would seem to imply that any other value is invalid, while the goal is to allow possible additions in the future.

Sorry, I am not sure I got your point. Then “possible values” would imply that any other value is impossible.

robUx4:
Not much of an English speaker either but it seems that possible sounds less normative than valid, so less blocking for the future. But that's really splitting hair.

Let's wait for native speakers point of view.

JeromeMartinez:
Not native speaker, but I think to "Defined values".

robUx4:
Looks like "defined values" is the consensus.

SpencerDawkins:
"Defined values" seems better than "standard values", if there are values that are not defined in an IETF standards-track document.  Colors, especially, may be used before they are standardized. This could be "valid values" if they are valid for a given version of Matroska. Possibly decide if a new value requires a new Matroska version or not. 

Going with defined values.

retokromer (approved):

SpencerDawkins (approved):
I think this is a good decision. Thanks for driving it to a conclusion.

PR #541, opened 2021-08-25
Clean segment linking explanations

### LINKED ISSUE #490: Rewrite parts of segment linking
At least the original medium linking section will need a bit of cleaning/rewriting. The text seemed clearer, in particular in relation to the relation with other kinds of linking.

> This kind of linking is a mix between hard and soft linking. Each segment linked is independant from the other (standalone, unlike hard linked ones). But it should be treated as a hard-link by the player. Medium linking is done through chapters using the ChapterSegmentUID element and only makes sense for ordered editions. The Segment matching the UID should be played as if it was part of the original segment (segment it's linked from) and then resume playback in the original segment. That means the timecodes of the following content should be shifted by the duration of the linked segment. As for hard-linking, the resulting segment edition should be played and considered as one.

robUx4:
* soft linking is not a real segment linking, it's a way to seek into other segments, there's no virtual segment involved (single presentation)
* SegmentFilename, PrevFilename and NextFilename are mostly for file recovery
* SegmentFamily is a generic thing usable by segment linking *and* chapter codec seeking
* mark the NextUID/PrevUID as invalid for the first/last linked segment
* make at least one NextUID/PrevUID mandatory per link node
* group/split rules that apply to hard linking, medium linking or both
* +add an example of chapter codec seeking/linking

Fixes #490

### DIFF
--- a/chapter_codecs.md
+++ b/chapter_codecs.md
@@ -4,7 +4,29 @@ title: Chapter Codecs
 
 # Matroska Chapter Codecs
 
-*TODO*
+Chapter codecs are a way to add more complex playback features than the usual linear playback.
+
+Some `ChapProcess` elements hold commands to execute when entering/leaving a chapter.
+
+When chapter codecs are used the `EditionFlagOrdered` of the edition they belong to **MUST** be set.
+
+## Segment Linking
+
+Chapter Codecs can reference another `Segment` and jump to that `Segment`.
+
+The Chapter Codecs **MAY** store the Segment information in their own format, possibly not using the `SegmentUID` format.
+The `ChapterTranslate` element and its child elements **SHOULD** be used
+to link the internal chapter codec representation, the chapter codec number and the actual Segment it represents.
+
+For example if a chapter codec of type "1" in SegmentA needs to link to SegmentB,
+it can store that information as "SegB" in its internal data.
+
+The translation `ChapterTranslate` in SegmentB would use the following elements:
+* `ChapterTranslate\ChapterTranslateCodec` = 1
+* `ChapterTranslate\ChapterTranslateID` = "SegB"
+
+The `Matroska Player` **MUST** use the `SegmentFamily` to find all Segments that need translation
+between the chapter codec values and the actual segment it targets.
 
 ## Matroska Script (0)
 
--- a/notes.md
+++ b/notes.md
@@ -590,12 +590,13 @@ in order to have seamless transition between segments.
 
 ## Hard Linking
 
-Hard Linking (also called splitting) is the process of creating a `Linked Segment`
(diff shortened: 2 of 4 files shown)


chapter_codecs.md
```diff
@@ -4,8 +4,32 @@ title: Chapter Codecs
 
 # Matroska Chapter Codecs
 
+Chapter codecs are way to add more complex playback features than the usual linear playback.
+
+Some `ChapProcess` elements hold commands to execute when entering/leaving a chapter.
+
+When chapter codecs are uses the `EditionFlagOrdered` of the edition they belong to **MUST** be set.
```
hubblec4, line 11: Google trasnslater suggested to use "used" instead of "uses".

> When chapter codecs are used the `EditionFlagOrdered` of the edition they belong to **MUST** be set.

notes.md
```diff
@@ -489,81 +489,96 @@ an `Element` is calculated by subtracting the position of the `Element Data` of
...
+
+With Hard Linking, the chapters of any `Segment` within the `Linked Segment` **MUST** only reference the current `Segment`.
+The `NextUID` and `PrevUID` reference the respective `SegmentUID` values of the next and previous `Segments`.
+
+The first `Segment` of a `Linked Segment` **MUST NOT** have a `PrevUID Element`.
+The last `Segment` of a `Linked Segment` **MUST NOT** have a `NextUID Element`.
+
+For each node of the chain of `Segments` of a `Linked Segment` at least one `Segment` **MUST** reference the other `Segment` of the node.
+
+In a chain of `Segments` of a `Linked Segment` the `NextUID` always takes precedence over the `PrevUID`.
```
hubblec4, line 530: This is not clear to me.
What is the reason for this? What should be achieved with it?

notes.md
```diff
@@ -489,81 +489,96 @@ an `Element` is calculated by subtracting the position of the `Element Data` of
...
+So if SegmentA has a `NextUID` to SegmentB and SegmentB has a `PrevUID` to SegmentC,
+the link to use is `NextUID`.
```
hubblec4, line 532: Also this is not clear to me.

notes.md
```diff
@@ -489,81 +489,96 @@ an `Element` is calculated by subtracting the position of the `Element Data` of
...
+If SegmentB has a `PrevUID` to SegmentA but SegmentA has no `NextUID`, then the Matroska Player
 **MAY** consider these two Segments linked as SegmentA followed by SegmentB.
```
hubblec4, line 534: This is not fully correct.

Hard-Linking is alwasy a question of "which Segment was started/used" by a user.

When a user starts SegmentA than is there no Hard-Linking because there is no `NextUID`.
Only if the user starts SegmentB than the player has to use Hard-Linking.

Issue #337, opened 2019-08-10
Colour Space values


gibix:
I want to parse `\Segment\Tracks\TrackEntry\Video\ColourSpace` and is not clear to me witch are all the possible values to handle by a complete mkv demuxer.

dericed:
I'm interested in making the description more specific to recommend an external authority or to possible list a set as should-be-supported; however at the moment I think the intent is that this field is intended to store the fourcc for uncompressed video as it would be in avi and I'm not sure if there is a decent authority list of those fourccs to cite.

lu-zero:
The AVI fourccs would clash with the `Colour` description, I'd deprecate the use (since the current description of the field does not even say that the AVI ones are supported).

Talking with @robUx4 I'll prepare a pull request to have a mapping for:
- `RGB`, 
- `YCbCr`
- `YCoCg`, 
- `ICtCp`, 
- `XYZ`
- `CMYK`
- `HSV`
- `LAB`

Expressed as `'M'`,`'0'`, `'0'`, `{numeric value matching the list order}`.

The `Colour` should contain all the other useful information.

kodabb:
I think `ColourSpace` is poorly named and mostly refers to how data is packed when it is RAW: for example it defines VYUY, and 2YUY which are packing modes, not related to color systems (which is what @lu-zero is referring to) or color spaces (which is what @dericed is referring to i think)

having a ColorSystem entity would be nice, I'm unsure of the usefulness of this ColourSpace entity

robUx4:
For the record, `ColourSpace` has been renamed `UncompressedFourCC` in d309e6828b92832f90c9b36bfc65e92cbf9040af and is published as such in RFC9559.

PR #548, opened 2021-08-26
Clarify the uncompressed fourcc element

robUx4:
a.k.a. `ColourSpace`

This element is only used with the `V_UNCOMPRESSED` codec (both in libavformat and mkvtoolnix). It's a FourCC that is usually found in AVI. It's goal, for uncompressed "codec" is to define how to interpret the "raw" data.

The element should only be used with `V_UNCOMPRESSED`. Nowadays we have good support for colorimetry (including colorspace) fields that are useful for uncompressed data. It may be lacking in the raw pixel format description.

fourcc.org has a good list of YUV and RGB formats that can be used in AVI and thus in this field. It even has packed formats like NV12. But not P010, P012, P014 and P016, its 10/12/14/16 bits counterparts.

A full/clean definition is out of the scope of the Matroska spec and should be done better in the codec spec.

Ref #337

### DIFF
--- a/codec_specs.md
+++ b/codec_specs.md
@@ -162,7 +162,7 @@ Codec ID: V_UNCOMPRESSED
 
 Codec Name: Video, raw uncompressed video frames
 
-Description: All details about the used color specs and bit depth are to be put/read from the `TrackEntry\Video\ColourSpace` elements.
+Description: All details about the used color specs and bit depth are to be put/read from the `TrackEntry\Video\UncompressedFourCC` elements.
 
 Initialization: none
 
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -554,10 +554,10 @@ PixelHeight - PixelCropTop - PixelCropBottom, else there is no default value.</i
     </restriction>
     <extension type="libmatroska" cppname="VideoAspectRatio"/>
   </element>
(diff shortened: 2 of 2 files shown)


JeromeMartinez:
https://web.archive.org/web/20220116212744/https://www.fourcc.org/ is the savior :).

ndufresne:
For the record, we wanted to find something that fit 10bit RGB as uncompressed in MTK, we have no idea were these are from, but we will opt for what FFMPEG do ('R' 'B' 'A' 64 and 'A' 'B' 'R' 64).

JeromeMartinez:
> 10bit RGB [...] 'R' 'B' 'A' 64

Storing 10-bit content in 16-bit storage is spending 50% additional storage cost than needed.

Unfortunately, RGB in AVI (used by Matroska for storing uncompressed content) expects 1,4,8,16,24,32 bits (only YUV has 10-bit, with `v210` fourcc).
But you may be interested in trying FFV1, slower (compression) but lossless and keep your 10-bit metadata and 50% less storage cost), with `ffmpeg -c:v ffv1`.

Now found a good reason to add a `V_DPX` mapping (10-bit RGB input --> DPX 10-bit --> MKV with DPX for uncompressed 10-bit) :-p.

robUx4:
Yes we use the Wayback Machine now since bb0e8797f2b9955c00424ff6f7643eb8662e3649

PR #549, opened 2021-08-26
forbid lacing when writing a single frame

### LINKED ISSUE #355: EBML lacing with only one frame

robUx4:
There is no existing (open source) library that can write a lace with a single
frame, especially not libmatroska (forces no lacing in that case) and libavformat
(doesn't write lacing at all).

Fixes #355

### DIFF
--- a/notes.md
+++ b/notes.md
@@ -79,11 +79,17 @@ Table: Block Header flags part{#blockHeaderFlags}
 ## Lacing
 
 Lacing is a mechanism to save space when storing data. It is typically used for small blocks
-of data (referred to as frames in Matroska). There are 3 types of lacing:
+of data (referred to as frames in Matroska). It packs multiple frames into a single `Block` or `SimpleBlock`.
+
+Lacing **MUST NOT** be used to store a single frame in a `Block` or `SimpleBlock`.
+
+There are 3 types of lacing:
 
 1. Xiph, inspired by what is found in the Ogg container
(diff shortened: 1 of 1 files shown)


mkver:
Just forbidding it seems indeed like the easiest solution. I am ok with this.

PR #552, opened 2021-08-27
allow chapter duration of 0

### LINKED ISSUE #542: really allow 0 duration chapters
In #464 I said it was not necessary. In retrospect I think a duration of 0 for some chapters is necessary.

>  In ordered editions, as said above, you can merge things you need for that chapter in neighbor chapters. Only a marker right at the end of the file may not be possible but I'm not sure it's of any use.

I found a simple DMX generated file (from 2005) which has a duration of 0:
```xml
<Chapters>
 <EditionEntry>
  <EditionUID>1676494232</EditionUID>
  <EditionFlagOrdered>1</EditionFlagOrdered>
  <!-- First Play PGC -->
  <ChapterAtom>
   <ChapterUID>1246888591</ChapterUID>
   <ChapterTimeStart>00:00:00.000000000</ChapterTimeStart>
   <ChapterTimeEnd>00:00:00.000000000</ChapterTimeEnd>

   <ChapterDisplay>
    <ChapterString>First Play PGC 1246888591</ChapterString>
   </ChapterDisplay>

   <ChapterProcess>
    <ChapterProcessCodecID>1</ChapterProcessCodecID>
    <ChapterProcessPrivate format="hex">30 00 00 00 </ChapterProcessPrivate>
    <!-- Pre commands -->
    <ChapterProcessCommand>
     <ChapterProcessTime>1</ChapterProcessTime>
     <ChapterProcessData format="hex">01 30 02 00 00 00 02 00 00 </ChapterProcessData>
    </ChapterProcessCommand>
   </ChapterProcess>
  </ChapterAtom>
 </EditionEntry>
</Chapters>
```

This is actually the equivalent of the entry point of the DVD (first play program chain). And because it uses DVD commands (when entering) it cannot be merged into another chapter that would also start at `00:00:00.000000000`, because it may also have some commands on entering. (I do have this case as well)

One reason not to allow 0 duration was the odd situation where the `ChapterTimeEnd` value is excluded from the chapter content to play (in ordered editions). And at the same time it's included in `ChapterTimeStart`. But we can just say that in this case the duration is 0. The player should just enter/leave such a chapter nonetheless during playback, in order to handle the `ChapProcess` content.

robUx4:
There can be segments with just chapters containing commands.

In ordered chapters the content will not play anything, it may seek to the
position for nothing but that's not a big deal.

Fixes #542

### DIFF
--- a/chapters.md
+++ b/chapters.md
@@ -109,6 +109,11 @@ A `Matroska Player` calculates the duration of this `Chapter` using the differen
 `ChapterTimeEnd` and `ChapterTimeStart`.
 The end timestamp **MUST** be greater than or equal to the start timestamp.
 
+When the `ChapterTimeEnd` timestamp is equal to the `ChapterTimeStart` timestamp,
+the timestamps is included in the `Chapter`. It can be useful to put markers in
+a file or add chapter commands with ordered chapter commands without having to play anything;
+see (#chapprocess-element).
+
 Chapter   | Start timestamp | End timestamp | Duration
 :---------|:----------------|:--------------|:-----
 Chapter 1 | 0               | 1000000000    | 1000000000
--- a/chapters.md
+++ b/chapters.md
@@ -107,13 +107,13 @@ The timestamp of the end of `Chapter` with nanosecond accuracy, not scaled by Ti
 The timestamp defined by the `ChapterTimeEnd` is not part of the `Chapter`.
 A `Matroska Player` calculates the duration of this `Chapter` using the difference between the
 `ChapterTimeEnd` and `ChapterTimeStart`.
-The end timestamp **MUST** be strictly greater than the start timestamp.
+The end timestamp **MUST** be greater than or equal to the start timestamp.
 
 Chapter   | Start timestamp | End timestamp | Duration
 :---------|:----------------|:--------------|:-----
 Chapter 1 | 0               | 1000000000    | 1000000000
 Chapter 2 | 1000000000      | 5000000000    | 4000000000
-Chapter 3 | 6000000000      | 6000000000    | Invalid (0)
+Chapter 3 | 6000000000      | 6000000000    | 0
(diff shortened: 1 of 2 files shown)


robUx4:
When exporting a DVD to Matroska the "first play PGC" often just contains a command. In Matroska chapter that corresponds to an empty chapter with an enter (or leave) command.

Here's an example output of [DvdMenuXtractor](https://github.com/Matroska-Org/dvdmenuxtractor/blob/c15d9fc4b91478d21d95ba7b54c361d71296e6f5/dmx/chaptermanager.cpp#L173):
```xml
		<!-- First Play PGC -->
		<ChapterAtom>
			<ChapterUID>1729839304</ChapterUID>
			<ChapterTimeStart>00:00:00.000000000</ChapterTimeStart>
			<ChapterTimeEnd>00:00:00.000000000</ChapterTimeEnd>

	<!-- Enter your text here -->
	<ChapterDisplay>
			<ChapterString>First Play PGC 1729839304</ChapterString>
	</ChapterDisplay>

			<ChapterProcess>
				<ChapterProcessCodecID>1</ChapterProcessCodecID>
				<ChapterProcessPrivate format="hex">30 00 00 00 </ChapterProcessPrivate>
				<!-- Pre commands -->
				<ChapterProcessCommand>
					<ChapterProcessTime>1</ChapterProcessTime>
					<ChapterProcessData format="hex">02 53 00 00 00 00 0f 00 00 30 06 00 00 00 42 00 00 </ChapterProcessData>
				</ChapterProcessCommand>
			</ChapterProcess>
		</ChapterAtom>
```

So it is needed in ordered chapters. It doesn't make a lot of sense in unordered chapters, but could be used as a marker of position (for example in an audio or video editor).

We do say that the end value is `timestamp excluded`. So we may clarify what to do with this when the start is equal to the end as it means the same timestamp is both included and excluded.

robUx4:
Fixed some typo for `greater than or equal to `.

PR #555, opened 2021-08-27
add a Random Access Point section after the blocks

robUx4:
This is line with #419 and the PR #554

* audio & subtitles are considered RAP by default (noone cared to set the proper flags in SimpleBlock)
* video are RAP if the keyframe flag is set or the BlockGroup has no ReferenceBlock.
* all video reference needed for a frame must be listed in a ReferenceBlock (if using a BlockGroup)
* intro-only (AV1) frames use a ReferenceBlock value of 0

### DIFF
--- a/notes.md
+++ b/notes.md
@@ -258,6 +258,51 @@ output sample is contiguous with a fixed frequency.
 For subtitles this is usually not the case so lacing **SHOULD NOT** be used.
 
 
+
+## Random Access Points
+
+Random Access Points (RAP) are positions where the parser can seek to and start playback without decoding
+of what was before. In Matroska `BlockGroups` and `SimpleBlocks` can be RAPs.
+To seek to these elements it is still necessary to seek to the `Cluster` containing them
+and start playback from the `BlockGroup` or `SimpleBlock` that is a RAP.
+
+Because a Matroska File is usually composed of multiple tracks playing at the same time
+-- video, audio and subtitles -- to seek properly to a RAP, each selected track must be
+taken in account. Usually all audio and subtitle `BlockGroup` or `SimpleBlock` are RAP.
+They are independent of each other and can be played randomly.
+
+Video tracks on the other hand often use references to previous and future frames for better
+coding efficiency. Frames with such reference **MUST** either contain one or more
+`ReferenceBlock` Elements in their `BlockGroup` or **MUST** be marked
+as non-keyframe in a `SimpleBlock`; see (#simpleblock-header-flags).
+
+Frames that are RAP -- i.e. they don't depend on other frames -- **MUST** set the keyframe
+flag if they are in a `SimpleBlock` or their parent `BlockGroup` **MUST NOT** contain
+a `ReferenceBlock`.
+
+There may be cases where the use of `BlockGroup` is necessary, as the frame may need a
+`BlockDuration`, `BlockAdditions`, `CodecState` or a `DiscardPadding` element.
+For thoses cases a `SimpleBlock` **MUST NOT** be used,
+the reference information **SHOULD** be recovered for non-RAP frames.
+
+When a frame in a `BlockGroup` is not a RAP, all references **SHOULD** be listed as a `ReferenceBlock`,
+at least some of them, even if not accurate, or one `ReferenceBlock` with the value "0" corresponding to a self or unknown reference.
+The lack of `ReferenceBlock` would mean such a frame is a RAP and seeking on that
+frame that actually depends on other frames **MAY** create bogus output or even crash.
+
+Intra-only video frames, such as the ones found in AV1 or VP9, can be decoded without any other
+frame, but they don't reset the codec state. So seeking to these frames is not possible
+as the next frames may need frames that are not known from this seeking point.
+Such intra-only frames **MUST NOT** be considered as keyframes so the keyframe flag
+**MUST NOT** be set in the `SimpleBlock` or a `ReferenceBlock` **MUST** be used
+to signify the frame is not a RAP. The timestamp value of the `ReferenceBlock` **MUST**
+be "0", meaning it's referencing itself.
+
+Because a video `SimpleBlock` has less references information than a video `BlockGroup`,
+it is possible to remux a video track using `BlockGroup` into a `SimpleBlock`,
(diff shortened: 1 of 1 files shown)


notes.md
```diff
@@ -190,6 +190,57 @@ For (possibly) Laced Data
...
+as non-keyframe in a `SimpleBlock`; see (#simpleblock-header-flags).
+
+Frames that are RAP -- i.e. they don't depend on other frames -- **MUST** set the keyframe
+flag if they are in a `SimpleBlock` or their parent `BlockGroup` **MUST NOT** contain
+any `ReferenceBlock`.
+
+For legacy reasons, audio and subtitle track **MAY** use a `SimpleBlock` without the keyframe
+flag set, even though they are RAPs. So for audio and subtitle tracks, the only way to specify
+a frame is not a RAP is to put it in a `BlockGroup` with a `ReferenceBlock`. In all other cases
+they **MUST** be considered as decodable on their own.
```
mbunkus, line 218: This interacts very, very badly with one of the most popular current audio codecs: TrueHD. TrueHD uses two different frame types: regular frames & major sync frames. Decoding can only start from major sync frames. `mkvmerge` and other software has therefore marked frames containing a TrueHD major sync frame with the "keyframe" flag and others without it.

Forcing muxers to use reference blocks here would be a significant increase in file overhead. TrueHD uses a high number of very small frames (both wrt. the number of samples it contains & the its byte size, often having a duration of less than 1ms with a size of around 100 bytes), and the overwhelming majority are regular sync frames, not major ones, in the range of 1 major sync frame every 125 regular frames.

All of those would have to use BlockGroups with reference blocks, which adds about 4% of container overhead for each TrueHD track.

In addition this would suddenly make all existing TrueHD-in-Matroska-with-Simpleblocks files invalid or at least bad to seek in.

What's the reasoning here? Why would we not keep using key frames for audio tracks correctly as there _are_ audio codecs that actually have such a distinction?

I can understand this with subtitle tracks as there's no subtitle codec I know of that uses such a distinction, but changing the semantics for audio tracks in this way is just plain wrong. I'm really against this change in its current form.

notes.md
```diff
@@ -190,6 +190,57 @@ For (possibly) Laced Data
...
+The lack of `ReferenceBlock` would mean all frames are RAP and seeking on
+frames that actually depend on other frames **MAY** create bogus output or even crash.
+
+Intra-only video frames, such as the ones found in AV1, can be decoded without any other
+frame, but they don't reset the codec state. So seeking to these frames is not possible
+as the next frames may need frames that are not known from this seeking point.
+Such intre-only frames **MUST NOT** be considered as keyframes so the keyframe flag
+**MUST NOT** be set in the `SimpleBlock` or a `ReferenceBlock` **MUST** be used
+to signify the frame is not a RAP. The timestamp value of the `ReferenceBlock` **MUST**
+be "0", meaning it's referencing itself.
```
mbunkus, line 235: This last change, especially the normative **MUST**, invalidates existing files. Being very aggressive here. What for?

mbunkus (changes requested):

notes.md
robUx4: > What's the reasoning here?

The reason is given in the first sentence: `For legacy reasons`. There are old files out there where the audio SimpleBlock never had the keyframe flag set. Looking at my old files it seems rare though. A deeper analysis is needed to find out what was writing these files and when it was fixed. It seems to be quite old. If we determine it's rare enough we can just add a mention that it exists and is bogus, the rule being to rely on the SimpleBlock flag for RAP, at least for audio.

There's also the question of lacing that I forgot to cover in #496. The keyframe flag **MUST** apply to all the frames in the lace (which are pretty much always audio). Are TrueHD files applying this rule or only the first frame of a lace has the keyframe flag ?

notes.md
robUx4: typo `intra-only`

This is a very particular case that only exist in AV1 yet AFAIK. The value "0" seems to be the correct/logical way. ie in a system that would interpret `ReferenceBlock` properly, it could cut the file accurately, keep that frame and also marking it a intra-only. Any other value (in the past) would mean it's referencing a frame which would need to be kept (and all other frames it depends on) when in fact it's not the case.

IMO for new features we should enforce clean rules, rather than

> listed as a `ReferenceBlock`, at least some of them, even not accurate

Also if we replace it with a **SHOULD** we have to explain why/when another could be used, but what would be the reason not to set 0 ?

notes.md
robUx4: After a look at (some of) my older files, I couldn't find audio tracks with SimpleBlock without the keyframe flag. BlockGroup don't have it so it's not an issue, they are all considered keyframes (I don't know about TrueHD). I'll have to dig deeper to find some samples and what/when they were produced (it may be easier to analyze of mkvmerge and libavformat code as well).

robUx4: Removed the legacy section. I could not find any old file that actually had this behavior.

notes.md
robUx4: Updated the paragraph:
* fixed typo
* added VP9 to codecs supporting intra-only frames

In libavcodec an MPEG video stream is also considered as intra-only when the GOP size is 1 (or 0).

robUx4: For `Rolling/Periodic Intra Refresh` mentioned in #419 (ie there's no actual keyframe, the keyframe parts are spread over frames, periodically) this is possible to tell the matroska demuxer how much data it needs to seek before the seeking point desired using `SeekPreRoll`. A track using `Periodic Intra Refresh` may have no actual keyframe, maybe not even the first frame. But with the `SeekPreRoll` it should be possible to set each frame as a keyframe or RAP, the frames will be fully usable only after `SeekPreRoll` has passed.

This is unrelated to intra-only which are decoded on their own but subsequent frames may depend on older frames. So they can't be considered RAP. Hence the use of `ReferenceBlock` with a special value, which is already defined as:

> The value "0" **MAY** also be used to signify this `Block` cannot be decoded on its own, but without knownledge of which `Block` is necessary. In this case, other `ReferenceBlock` **MUST NOT** be found in the same `BlockGroup`.

So this paragraph just expresses this in terms of RAP.

PR #556, opened 2021-09-06
Readd order of cropping and resizing

mkver:
Lost in 3e0347a7b4799087e8e1de309719609aeb9ce58c (PR #184).

### DIFF
--- a/notes.md
+++ b/notes.md
@@ -405,6 +405,9 @@ rows of pixels **SHOULD** be cropped from the top and bottom of the image (respe
  so that a `Matroska Player` **SHOULD** crop off 240 columns of pixels from the left and
  right of the encoded image to present the image with the pillar-boxes hidden.
 
+Cropping has to be performed before resizing and the display dimensions given by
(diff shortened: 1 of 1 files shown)


notes.md
```diff
@@ -405,6 +405,9 @@ rows of pixels **SHOULD** be cropped from the top and bottom of the image (respe
  so that a `Matroska Player` **SHOULD** crop off 240 columns of pixels from the left and
  right of the encoded image to present the image with the pillar-boxes hidden.
 
+Cropping has to be performed before resizing and the display dimensions given by
+ `DisplayWidth`, `DisplayHeight` and `DisplayUnit` apply to the already cropped image.
```
robUx4, line 409: IMO no need to mention `DisplayUnit` here. It's already defined as 

> How DisplayWidth & DisplayHeight are interpreted

It's more confusing because the cropping values **are** in pixel, regardless of the DisplayUnit.

Other than that LGTM.

robUx4: Actually it's a good idea to mention that DisplayUnit doesn't apply to cropping which should only be done on pixels.

PR #557, opened 2021-09-12
make (some) elements with a default value mandatory

robUx4:
These elements were already using the same "value" if the element wasn't present anyway.

* CodecDelay: nothing to add/remove to track timestamps
* PixelCropXXX: no cropping by default
* BitsPerChannel: "0 indicates that the BitsPerChannel is unspecified"
* CueTrackPositions: "0 means that the data is taken from the initial Track Entry"
* Tag\Targets\TargetTypeValue: **tag is not usable without this value**
* TagXXXUID: 0 = the tags apply to all XXX in the Segment

Following the discussion on #501 it would be more logical/easier to make elements mandatory when they have a default value.

Only `AlphaMode` is not included as the presence of the element signals the flag (!).

This doesn't include deprecated element (maxver=0).


JeromeMartinez:
> Only AlphaMode is not included as the presence of the element signals the flag (!).

I would have it too, after having defined 0 as "Unknown" (see my comment on the dedicated PR).

PR #558, opened 2021-09-12
make AlphaMode mandatory

robUx4:
This is a WebM thing loosely defined. I'm not sure where the default value
came from.

Having just the presence of the element present to signal something is at odd
with all other elements. The default value in this case has no use. And
following the discussions in #501 elements with a default value would preferably
be mandatory. But that's not possible with the "presence" signaling something.

So just remove the default value.

Both libavformat [1] and libwebm [2] write a value of 1 when they want to signal
it.

[1] https://github.com/FFmpeg/FFmpeg/blob/master/libavformat/matroskaenc.c#L1274
[2] https://aomedia.googlesource.com/aom/+/refs/heads/master/third_party/libwebm/mkvmuxer/mkvmuxer.cc#1522

### DIFF
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -484,7 +484,7 @@ with the top line of the top field stored first.</documentation>
     <extension type="libmatroska" cppname="VideoStereoMode"/>
   </element>
   <element name="AlphaMode" path="\Segment\Tracks\TrackEntry\Video\AlphaMode" id="0x53C0" type="uinteger" minver="3" default="0" minOccurs="1" maxOccurs="1">
-    <documentation lang="en" purpose="definition">Set to 1 if the BlockAdditional Element **MAY** contain Alpha data.</documentation>
+    <documentation lang="en" purpose="definition">Set to 1 if the BlockAdditional Element with BlockAddID of "1" contains Alpha data.</documentation>
     <extension type="webmproject.org" webm="1"/>
     <extension type="libmatroska" cppname="VideoAlphaMode"/>
   </element>
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -483,7 +483,7 @@ with the top line of the top field stored first.</documentation>
     <extension type="webmproject.org" webm="1"/>
     <extension type="libmatroska" cppname="VideoStereoMode"/>
   </element>
-  <element name="AlphaMode" path="\Segment\Tracks\TrackEntry\Video\AlphaMode" id="0x53C0" type="uinteger" minver="3" default="0" maxOccurs="1">
+  <element name="AlphaMode" path="\Segment\Tracks\TrackEntry\Video\AlphaMode" id="0x53C0" type="uinteger" minver="3" default="0" minOccurs="1" maxOccurs="1">
     <documentation lang="en" purpose="definition">Alpha Video Mode. Presence of this Element indicates that
 the BlockAdditional Element could contain Alpha data.</documentation>
     <extension type="webmproject.org" webm="1"/>
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -484,7 +484,17 @@ with the top line of the top field stored first.</documentation>
     <extension type="libmatroska" cppname="VideoStereoMode"/>
   </element>
   <element name="AlphaMode" path="\Segment\Tracks\TrackEntry\Video\AlphaMode" id="0x53C0" type="uinteger" minver="3" default="0" minOccurs="1" maxOccurs="1">
-    <documentation lang="en" purpose="definition">Set to 1 if the BlockAdditional Element with BlockAddID of "1" contains Alpha data.</documentation>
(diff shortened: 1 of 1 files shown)


JeromeMartinez:
For me a default of `0` is fine, and actually better. 0 could be defined as "Unknown".

robUx4:
But that's not compatible with what the [WebM spec](https://www.webmproject.org/docs/container/#AlphaMode) says:

> Alpha Video Mode. Presence of this element indicates that the BlockAdditional element could contain Alpha data. 

0, 1, 2, ..., 666, 667, ... all mean that the BlockAdditional element could contain alpha data. That would break compatibility with existing parsers.

robUx4:
BTW Firefox, one of the browsers that can handle alpha in WebM, [seems to use nestegg](https://bugzilla.mozilla.org/show_bug.cgi?id=1267492) to parse WebM. 
A value read as 0 doesn't mean the same as a value read as 1. The value of `alpha_mode` is 0 if the element is not present or if it was stored as 0.

https://github.com/mozilla/nestegg/blob/ec6adfbbf979678e3058cc4695257366f39e290b/src/nestegg.c#L2628
https://github.com/mozilla/nestegg/blob/ec6adfbbf979678e3058cc4695257366f39e290b/src/nestegg.c#L896

robUx4:
It seems to be the same in Chrome/chromium/Edge/etc. The internal read value is [initialized to `-1`](https://github.com/chromium/chromium/blob/72ceeed2ebcd505b8d8205ed7354e862b871995e/media/formats/webm/webm_video_client.cc#L60). And it's [compared to `1`](https://github.com/chromium/chromium/blob/72ceeed2ebcd505b8d8205ed7354e862b871995e/media/formats/webm/webm_video_client.cc#L148) to determine whether the video is opaque (`kIsOpaque`) or has alpha (`kHasAlpha`).

robUx4:
In both cases, the presence of the element is not sufficient to determine if the video track has an alpha channel. 

* a value of `0` means it has no alpha channel,
* a value of `1` means it has an alpha channel,
* a value of `2`, `3`, etc means it has an alpha channel for Firefox, but not for Chromium.

We have 2 choices to match the rules discussed #501, remove the default value, or make it mandatory. In both cases we have to explain than only a value of `1` means there is an alpha channel.

I'd rather make it mandatory as done for all other elements with a default value of 0 in #557.

robUx4:
Make mandatory as in #557 and define it as other flags (value 1 means defined).

ebml_matroska.xml
```diff
@@ -483,9 +483,8 @@ with the top line of the top field stored first.</documentation>
     <extension type="webmproject.org" webm="1"/>
     <extension type="libmatroska" cppname="VideoStereoMode"/>
   </element>
-  <element name="AlphaMode" path="\Segment\Tracks\TrackEntry\Video\AlphaMode" id="0x53C0" type="uinteger" minver="3" default="0" maxOccurs="1">
-    <documentation lang="en" purpose="definition">Alpha Video Mode. Presence of this Element indicates that
-the BlockAdditional Element could contain Alpha data.</documentation>
+  <element name="AlphaMode" path="\Segment\Tracks\TrackEntry\Video\AlphaMode" id="0x53C0" type="uinteger" minver="3" default="0" minOccurs="1" maxOccurs="1">
+    <documentation lang="en" purpose="definition">Set to 1 if the BlockAdditional Element with BlockAddID of "1" contains Alpha data.</documentation>
```
JeromeMartinez, line 487: Thank you for having digged in how it is implemented in main browsers. [code from Google](https://github.com/chromium/chromium/blob/72ceeed2ebcd505b8d8205ed7354e862b871995e/media/formats/webm/webm_video_client.cc#L148) does not match with [WebM specs from Google](https://www.webmproject.org/docs/container/) :-p.

I understand that it is only for BlockAdditional and that it must not (?) be used for something else. Checking FFmpeg, it is not used for e.g. FFV1 RGBA. I would be more explicit about that this element is not intended for alpha mode in general despite its name (I would change it but changing a name is maybe too much complicated compared to the low usage of this element) and the issue with values of 2 and more.

Also specification of the content is not indicated (I understand from BlockAddIDType definition that we could use different formats for alpha content or something else, not only the one understood by the decoder of the format defined in CodecID).

So suggestion:

```suggestion
    <documentation lang="en" purpose="definition">Indicate if the BlockAdditional Element with BlockAddID of 1 contains Alpha data conforming to the specification indicated by CodecID for alpha additional data passed to the codec. Undefined values SHOULD NOT be used as the behavior of known implementations is different (considered either as 0 or 1).</documentation>
    <restriction>
      <enum value="0" label="The BlockAdditional Element with BlockAddID of 1 does not exist or SHOULD NOT be considered as containing such data."/>
      <enum value="1" label="The BlockAdditional Element with BlockAddID of 1 contains such data."/>
    </restriction>
```
(BlockAddID is an uinteger so I would not use the quote for it, I intentionally removed quotes around its value).

Issue #414, opened 2020-08-23
Fix the IANA considerations for Matroska specs


robUx4:
In the Matroska, Codec and Tags documents they are set to `To be determined.`

- we probably want a registry of EBML IDs known in Matroska.
- we probably want a registry of codec IDs known in Matroska.
- we probably want a registry of tag names known in Matroska.

robUx4:
Split in different issues per project.

robUx4:
Should be done after #487 moves the sections.

robUx4:
I think the codec IDs and tag names should go in their respective document.

Not sure we should repeat the list of hundreds of EBML IDs found above in the document.

Issue #485, opened 2021-04-11
Fix the IANA considerations for Codec specs


robUx4:
Currently marked as TBD

robUx4:
Fixed in current master.

PR #564, opened 2021-09-26
add a codec ID IANA registry at the end of the codec document

robUx4:
This a crucial list of strings that need to be known by parsers to handle
Matroska files.

~Draft because it's missing the initial values.~

Ref: #414


robUx4:
Fxies #485

robUx4:
Move the text as the file was already created in 802e58b8a752c996fca74e82036382d333927cd2

retokromer (approved):

robUx4:
Once merge the codec of each category should be sorted alphanumerically so we can spot holes in the list (which is not auto generated).

retokromer (approved):

PR #578, opened 2022-01-16
Generate the initial IANA registry for Matroska IDs

### LINKED ISSUE #568: Generate the IANA list of Matroska IDs
To simplify the IANA registration, we can provide an XML file for the initial list. That will limit the human error factor (unless the generator) is bogus.

The format gathered by @mcr is on the [CELLAR mailing list](https://mailarchive.ietf.org/arch/msg/cellar/lqN1WX4HV7LEwBSHYy8sOgP9heI/).

Refs #563 / #414

robUx4:
Similar to the format of https://www.iana.org/assignments/ebml/ebml.xml

The IDs are sorted by length and ID value (like EBML ones).

The links to the actual RFC are temporary (no RFC numbers, section numbers can vary).

Fixes #568

### DIFF
--- a/transforms/ebml_schema2xml4iana_ids.xsl
+++ b/transforms/ebml_schema2xml4iana_ids.xsl
@@ -1,0 +1,46 @@
+<?xml version="1.0"?>
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" version="1.0" xmlns:ebml="urn:ietf:rfc:8794" xmlns="http://www.iana.org/assignments">
+  <xsl:output encoding="UTF-8" version="1.0" indent="yes"/>
+  <xsl:variable name="smallcase" select="'abcdefghijklmnopqrstuvwxyz'"/>
+  <xsl:variable name="uppercase" select="'ABCDEFGHIJKLMNOPQRSTUVWXYZ'"/>
+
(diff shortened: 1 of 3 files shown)


transforms/ebml_schema2xml4iana_ids.xsl
```diff
@@ -0,0 +1,46 @@
...
+    <registry id="matroska">
+      <title>Matroska</title>
+      <created>@BUILD_DATE@</created>
+
+      <!-- Element IDs -->
+      <registry id="matroska-element-ids">
+        <title>Matroska Element IDs</title>
+        <xref type="rfc" data="rfc8794"/>
+        <xsl:apply-templates select="//ebml:element[contains(@path,'\Segment') and string-length(@id)=4]">
+          <xsl:sort select="@id" />
```
retokromer, line 17: Nit: You can omit the space for all self-closing tags (before `/>`).

retokromer (approved):

robUx4:
It doesn't generate gaps between IDs, as found in the [EBML one](https://www.iana.org/assignments/ebml/ebml.xml), but that's still a good starting point with the correct values and correct pointers (to the draft for now and not the proper section numbers).

robUx4:
We may add the EBML for each ID as technically the same ID could be used in different places with different semantic meaning.

PR #579, opened 2022-01-23
fix the last sentence considered as a title

robUx4:
After joinging the files there seems to be an extra line needed by mmark.


retokromer (approved):
“Funny.”

PR #581, opened 2022-02-20
Misc ContentEncoding text and enum cleaning

### LINKED ISSUE #580: Describe how to use consecutive ContentEncoding
It is possible to use more than one `ContentEncoding` in a track. But there is nothing that describes how it should be applied. In particular in what order.

Also different `ContentEncoding` may have a different `ContentEncodingScope`. But that wouldn't work well if a different `ContentEncoding` is applied after or before. So there's probably some constraints needed there.

Noone probably ever used 2 `ContentEncoding` in a single track but the specs allow it. But it might even be useful (see [zstd Magic Number header stripping](https://github.com/ietf-wg-cellar/matroska-specification/pull/423#issuecomment-1037982014)). So we need to clarify this.

robUx4:
* reworked the ContentEncodingOrder text to clarify how to apply cascaded ContentEncoding
* turn enum values into simple name + description
* add AES cipher mode NIST link
* deprecate ContentEncodingScope `next` which doesn't make sense and is not used anywhere

Fixes #580


robUx4:
w3.org seems to be down but the schema is valid.

PR #583, opened 2022-03-14
bring back the CodecDelay default value

robUx4:
The default value was removed in 743317d06088e8a732a7b7f880386d8b0435142b but then the element was made mandatory in f81cb9d339574ce2b00faa288cd466908e31b797. Now files need to set a value even if they don't use the feature.

So we bring back the default value to fix this.

See issue found in libmatroska because of this https://github.com/Matroska-Org/libmatroska/pull/61.


mbunkus (approved):
LGTM. Thanks!

retokromer (approved):

PR #585, opened 2022-03-20
make Slices deprecated

robUx4:
In 7136c753fdecafbd81b11b0f31734ce836ce3ef8 all sub-elements had a maxver of
either 0 or 1 so we had to use 1.

Now all sub-elements have a maxver of 0 so it should be 0 as well. This is true
since 88adb292307bb5ca5df7b141e2df04cb1e16f1be.


robUx4:
Ref. #475 and #105

mbunkus (approved):

retokromer (approved):

PR #606, opened 2022-05-22
add EditionDisplay to give human readable names to chapter editions

### LINKED ISSUE #300: EditionName
To set a name for an edition you have to use Tags and there are you can set more as the name for example TargetTypeValue and TargetType. But for me are other values as the name total uninteresting.
When you delete the Tags all edition names are lost.
A parser must search the Tags additional to get the edition name(s).

It would be great to add the EditionName element.

robUx4:
When there are multiple editions it should make it easier to pick one for the user.

Otherwise tags with TagEditionUID as the target might be sufficient and more versatile.

Fixes #300


hubblec4:
Which edition name(string) should be used if there is an `EditionDisplay` and a `Tag`?
Has the `TagString` from the `SimpleTag` a higher priority?
Or should the `Tag` fully ignored if an `EditionDisplay` exists?

robUx4:
> Which edition name(string) should be used if there is an EditionDisplay and a Tag?
> Has the TagString from the SimpleTag a higher priority?
> Or should the Tag fully ignored if an EditionDisplay exists?

I created #655 to fix this issue which also exists for other elements and needs to be handled independently from this PR.

retokromer (approved):

robUx4:

PR #619, opened 2022-06-19
add a link to the Hard Linking explanations in Next/Prev UUID

robUx4:


retokromer (commented):
Are double parentheses necessary?

robUx4:
> Are double parentheses necessary?

One level is for the proper linking. Another is to put that link in parenthesis in the text. The "name" of the link is the text before the parenthesis. It gives something like

> that uses Hard Linking (<a href="#">Section 23.458.8</a>)

retokromer:
Thank you for the explanation!

retokromer (approved):

PR #639, opened 2022-06-26
Rearrange the Matroska versioning section

robUx4:
Move some text up and fix some linkes.


robUx4:
A lot of versioning and unknown elements is already defined in RFC8794

PR #648, opened 2022-07-06
fix UTF-8 that appear in the TXT output

robUx4:
The RFC generator complains some character at UTF-8.

### DIFF
--- a/chapters.md
+++ b/chapters.md
@@ -249,7 +249,7 @@ This would translate in the following matroska form, with the EBML tree shown as
         <ChapString>After the crime</ChapString>
       </ChapterDisplay>
       <ChapterDisplay>
-        <ChapString>Après le crime</ChapString>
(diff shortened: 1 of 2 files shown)


retokromer:
I am not familiar with the RFC generator, but could `&egrave;` etc. work or is only 7-bit ASCII allowed?

mcr:
Accents are supported on author's name, but not in the general text.
I think that xml2rfc will support &egrave:

retokromer:
Thank you for the information, @mcr

robUx4:
I tried the XML/HTML accent in the author, but idnits doesn't like it. And I found more hidden UTF-8 chars to fix.

retokromer:
Thanks for trying, @robUx4. It seems to be a RFC generator issue we cannot fix here.

retokromer (approved):

PR #649, opened 2022-07-06
complete media registration for matroska

mcr:
move file extensions into new media type registration in IANA Considerations section


robUx4 (approved):
LGTM

PR #654, opened 2022-07-10
Explain that Matroska itself doesn't create an Encryption scheme

robUx4:
We're just using pre-existing systems.

### DIFF
--- a/notes.md
+++ b/notes.md
@@ -586,6 +586,10 @@ Encryption in Matroska is designed in a very generic style to allow people to
 implement whatever form of encryption is best for them. It is possible to use the
 encryption framework in Matroska as a type of DRM (Digital Rights Management).
 
+This document does not specify any kind of standard for encrypting elements.
+The issue of key scheduling, authorisation, and authentication are out of scope.
+External entities have used these elements in proprietary ways.
+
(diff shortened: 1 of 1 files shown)


retokromer (approved):

notes.md
```diff
@@ -586,6 +586,9 @@ Encryption in Matroska is designed in a very generic style to allow people to
 implement whatever form of encryption is best for them. It is possible to use the
 encryption framework in Matroska as a type of DRM (Digital Rights Management).
 
+Matroska itself doesn't create its own encryption system. It uses other existing
+encryption systems, with their data wrapped inside Matroska Elements.
+
```
mcr, line 591:
```suggestion
This document does not specify any kind of standard for encrypting elements.
The issue of key scheduling, authorisation, and authentication are out of scope.
External entities have used these elements in proprietary ways.

```

mcr (changes requested):

notes.md
retokromer: The issues [plural] of key scheduling etc. are out of scope.

robUx4: Applied and updated the author.

retokromer (approved):

PR #657, opened 2022-08-07
exclude Matroska v5 elements from the v4 document

robUx4:
So we can start merging v5 elements and they won't affect the v4 document.

v5 elements will likely be part of a new document that may not include elements
from v1 to v4.

### DIFF
--- a/transforms/ebml_schema2markdown4iana_ids.xsl
+++ b/transforms/ebml_schema2markdown4iana_ids.xsl
@@ -6,7 +6,7 @@
   <xsl:template match="ebml:EBMLSchema">
     <xsl:text>Element ID | Element Name            | Reference&#xa;</xsl:text>
     <xsl:text>----------:|:------------------------|:-------------------------------------------&#xa;</xsl:text>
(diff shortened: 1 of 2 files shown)


robUx4:
#318 and #606 could be merged after this MR.

robUx4:
Ping @mbunkus I'd like to merge this so I can merge #318, #606 and #659. For the Matroska specs it will be transparent. But on the (de)muxer side that means you may start adding v5 support in you case you encounter those elements.

retokromer (approved):
LGTM

mbunkus (changes requested):
As far as I know, [XPath supports `<=`](https://www.w3.org/TR/1999/REC-xpath-19991116/#NT-RelationalExpr), so why not use the [`number` function](https://www.w3.org/TR/1999/REC-xpath-19991116/#function-number) and write this as `and (number(@minver)<=4)` (untested)? It'd be what we actually want to express, isn't it? Or maybe we'd have to handle `minver` not being present, too.

robUx4:
Done. We can now have v6, v7, v8, etc and the source will still produce the same v1 to v4 IETF document.

PR #659, opened 2022-08-07
Add an audio emphasis element (v5) to mark some tracks

### LINKED ISSUE #582: Add audio emphasis information per track
Vinyl records are recorded with an [emphasis filter](https://en.wikipedia.org/wiki/RIAA_equalization) to avoid some avoid having too much bass. Most commly the RIAA phono curve is applied and un-applied in phono pre-amplifiers.

CDs also have an [emphasis bit](https://www.studio-nibble.com/cd/index.php?title=Pre-emphasis_(release_list)) that tells a filter has been applied.

If one records/rips raw data from the medium, the proprer filter has to be applied during playback. So there should be an element in the audio track entry of such Matroska files. This can't be a tag which should have no impact on playback.

There doesn't seem to be such information in Broadcast Wave File. So I'm not sure there's an existing standard for the values that are commonly found. We may need to define these values ourselves. With at least the RIAA curve and the CD one. For old phono records each company had its own curve, so there might be a lot of possible values. And we need a reliable source about the curve to use, for each value we define.

robUx4:
Fixes #582 

* The reserved values is added so that the first 4 values match the [MPEG Audio header](http://www.mpgedit.org/mpgedit/mpeg_format/mpeghdr.htm).
* The phono values are taken from this detailed [document](https://wiki.audacityteam.org/wiki/78rpm_playback_curves). Only the curves using 3 time constants (as the RIAA) ones are kept (1950 and up, before it's too messy)
* I couldn't find a proper RIAA document with the equation but there's a NAB document with the same values and the equation available.
* The CCITT J.17 is actually T-REC-J.17
* I could not find a proper spec for the FM time constants, even [Wikipedia](https://en.wikipedia.org/wiki/FM_broadcasting#Pre-emphasis_and_de-emphasis) doesn't mention any.

### DIFF
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -1041,6 +1041,47 @@ Setting `ProjectionPoseRoll` to 180 or -180 degrees, with the `ProjectionPoseYaw
     <extension type="stream copy" keep="1"/>
     <extension type="webmproject.org" webm="1"/>
   </element>
+  <element name="Emphasis" path="\Segment\Tracks\TrackEntry\Audio\Emphasis" id="0x92F1" type="uinteger" minver="5" default="0" minOccurs="1" maxOccurs="1" >
+    <documentation lang="en" purpose="definition">Audio emphasis applied on audio samples. The player **MUST** apply the inverse emphasis to get the proper audio samples.</documentation>
+    <restriction>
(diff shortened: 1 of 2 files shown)


robUx4:

PR #661, opened 2022-08-15
Rework the element formatting for more consistency

robUx4:
Remove redundant information, regroup some information.

This saves 2 pages on the TXT output.

It turns
```
5.1.4.1.21.  Language Element

   name / type / id:  Language / string / 0x22B59C

   path:  "\Segment\Tracks\TrackEntry\Language"

   minOccurs - maxOccurs:  1 - 1

   default:  eng

   definition:  Specifies the language of the track in the Matroska
      languages form; see Section 12 on language codes.  This Element
      MUST be ignored if the LanguageBCP47 Element is used in the same
      TrackEntry.
```

into

```
5.1.4.1.21.  Language Element

   id / type / default:  0x22B59C / string / eng

   path:  "\Segment\Tracks\TrackEntry\Language"

   minOccurs / maxOccurs:  1 / 1

   definition:  Specifies the language of the track in the Matroska
      languages form; see Section 12 on language codes.  This Element
      MUST be ignored if the LanguageBCP47 Element is used in the same
      TrackEntry.
```

### DIFF
--- a/transforms/ebml_schema2markdown4rfc.xsl
+++ b/transforms/ebml_schema2markdown4rfc.xsl
@@ -53,7 +53,7 @@
     <xsl:if test="@minOccurs | ebml:implementation_note[@note_attribute='minOccurs']">
       <xsl:text>minOccurs</xsl:text>
       <xsl:if test="@maxOccurs">
-        <xsl:text> - maxOccurs</xsl:text>
+        <xsl:text> / maxOccurs</xsl:text>
       </xsl:if>
       <xsl:text>:&#xa;: </xsl:text>
       <xsl:choose>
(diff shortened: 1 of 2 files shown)


transforms/ebml_schema2markdown4rfc.xsl
```diff
@@ -54,9 +54,9 @@
       <xsl:value-of select="@range"/>
       <xsl:text>&#xa;&#xa;</xsl:text>
     </xsl:if>
-    <xsl:if test="@size">
-      <xsl:text>size:&#xa;: </xsl:text>
-      <xsl:value-of select="@size"/>
+    <xsl:if test="@length">
+      <xsl:text>length:&#xa;: </xsl:text>
+      <xsl:value-of select="@length"/>
```
robUx4: @mcr We should include this before we submit for more reviewing (unless it has been done already). Otherwise all the length fields are not mentioned in the spec, which doesn't include the non normative XML file.

retokromer:
nit: the official casing is FFmpeg as in the commit description (not FFMpeg as in the changes)

robUx4:
Oops, I missed your comment. Fixed silently in 56e8272641dd372599275ae185703e10651fee9b.

Thanks for watching out !

PR #673, opened 2022-09-22
fix element ID of Emphasis element

### LINKED ISSUE #672: invalid element ID for new element Emphasis
The new Emphasis element has an element ID of `0x92f1`. The first byte's most significant bit is set, though, which indicates that the ID is only one byte long.

I'll add a PR using a different ID in a moment.

mbunkus:
0x92F1 is invalid: the first byte's most significant bit is set, which indicates a one byte long element ID. However, 0x92F1 is two bytes long.

0x72F1 is not used yet.

Fixes #672.


ebml_matroska.xml
```diff
@@ -1041,7 +1041,7 @@ Setting `ProjectionPoseRoll` to 180 or -180 degrees, with the `ProjectionPoseYaw
     <extension type="stream copy" keep="1"/>
     <extension type="webmproject.org" webm="1"/>
   </element>
-  <element name="Emphasis" path="\Segment\Tracks\TrackEntry\Audio\Emphasis" id="0x92F1" type="uinteger" minver="5" default="0" minOccurs="1" maxOccurs="1" >
+  <element name="Emphasis" path="\Segment\Tracks\TrackEntry\Audio\Emphasis" id="0x72F1" type="uinteger" minver="5" default="0" minOccurs="1" maxOccurs="1" >
```
JeromeMartinez, line 1044:
```suggestion
  <element name="Emphasis" path="\Segment\Tracks\TrackEntry\Audio\Emphasis" id="0x52F1" type="uinteger" minver="5" default="0" minOccurs="1" maxOccurs="1" >
```

So we change the fewest count of bits needed to correctly indicate the length of the ID without changing other bits.
0x92F1 = **10**01 0010 1111 0001
0x52F1 = **01**01 0010 1111 0001

mbunkus: I'm fine with whatever 😁

mbunkus: Updated branch pushed.

JeromeMartinez (approved):

robUx4 (approved):
Yes, please merge. This won't affect the IETF documents, but it will be used by our code generators.

PR #674, opened 2022-09-27
Replace ISO3166-1 and IANADomains references by BCP47 two-letter region subtag

JeromeMartinez:
Normalize the text about languages and countries, always referencing the notes so definition is at one place only.
ISO3166-1 is replaced by [BCP47 two-letter region subtag](https://datatracker.ietf.org/doc/html/rfc5646#section-2.2.4).
IANADomains is removed because it is actually not (no more?) only about countries (it also trademarks or sponsored words), and items using it are nationalities (so pure ISO3166-1 is needed) or locations (tag may have additional info about state, city...)
IANALangRegistry is removed because [it is referenced by BCP47](https://datatracker.ietf.org/doc/html/rfc5646#section-3).

Note:  we may need to have 2 separate types Nationality and Country, as Country actually lists some regions. For example NC (New Caledonia) is ISO3166-1/IANADomains but is not a nationality (NC inhabitants have FR nationality). But may be too complex and we keep as is for now.

### DIFF
--- a/matroska_tags.xml
+++ b/matroska_tags.xml
@@ -32,9 +32,9 @@ targeted item taken from another work of art. All tags in this list can be used
     </tag>
     <tag name="COUNTRY" class="Nesting Information" type="UTF-8">
       <description lang="en">The name of the country
-that is meant to have other tags inside (using nested tags) to country specific information about the item.
+that is meant to have other tags inside (using nested tags) to country specific information about the item,
+in the Matroska countries form, i.e. [@!BCP47] two-letter region subtag, without the UK exception.
 All tags in this list can be used "under" the COUNTRY_SPECIFIC tag like LABEL, PUBLISH_RATING, etc.
-The country code uses the same 2 octets country-codes as in Internet domains [@!IANADomains] based on [@!ISO3166-1] alpha-2 codes.
 </description>
     </tag>
     <tag name="TOTAL_PARTS" class="Organization Information" type="UTF-8">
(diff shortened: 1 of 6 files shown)


matroska_tags.xml
```diff
@@ -32,9 +32,9 @@ targeted item taken from another work of art. All tags in this list can be used
     </tag>
     <tag name="COUNTRY" class="Nesting Information" type="UTF-8">
       <description lang="en">The name of the country
-that is meant to have other tags inside (using nested tags) to country specific information about the item.
+that is meant to have other tags inside (using nested tags) to country specific information about the item,
+in the Matroska countries form; see (#country-codes) on countries codes.
```
mbunkus, line 36: "countries codes" should be "country codes", just like "language codes": you only pluralize the second word.

matroska_tags.xml
```diff
@@ -252,25 +252,24 @@ an age in other countries or a URI defining a logo).</description>
       <description lang="en">Information on when the file was purchased; see also (#commercial) on purchase tags.</description>
     </tag>
     <tag name="RECORDING_LOCATION" class="Spatial Information" type="UTF-8">
-      <description lang="en">The location where the item was recorded. The countries corresponding to the string,
-same 2 octets country-codes as in Internet domains [@!IANADomains] based on [@!ISO3166-1] alpha-2 codes.
+      <description lang="en">The location where the item was recorded,
+in the Matroska countries form; see (#country-codes) on countries codes.
```
mbunkus, line 256: "countries codes" should be "country codes", just like "language codes": you only pluralize the second word.

matroska_tags.xml
```diff
@@ -252,25 +252,24 @@ an age in other countries or a URI defining a logo).</description>
...
 This code is followed by a comma, then more detailed information such as state/province, another comma, and then city. For example,
 "US, Texas, Austin". This will allow for easy sorting. It is okay to only store the country, or the country and the state/province.
 More detailed information can be added after the city through the use of additional commas. In cases where the province/state
 is unknown, but you want to store the city, simply leave a space between the two commas. For example, "US, , Austin".</description>
     </tag>
     <tag name="COMPOSITION_LOCATION" class="Spatial Information" type="UTF-8">
-      <description lang="en">Location that the item was originally designed/written. The countries corresponding to the string,
-same 2 octets country-codes as in Internet domains [@!IANADomains] based on [@!ISO3166-1] alpha-2 codes.
+      <description lang="en">Location that the item was originally designed/written,
+in the Matroska countries form; see (#country-codes) on countries codes.
```
mbunkus, line 264: "countries codes" should be "country codes", just like "language codes": you only pluralize the second word.

matroska_tags.xml
```diff
@@ -252,25 +252,24 @@ an age in other countries or a URI defining a logo).</description>
...
 For example, "US, Texas, Austin". This will allow for easy sorting. It is okay to only store the country, or the country and the state/province.
 More detailed information can be added after the city through the use of additional commas. In cases where the province/state is unknown,
 but you want to store the city, simply leave a space between the two commas. For example, "US, , Austin".</description>
     </tag>
     <tag name="COMPOSER_NATIONALITY" class="Spatial Information" type="UTF-8">
-      <description lang="en">Nationality of the main composer of the item, mostly for classical music.
-The countries corresponding to the string,
-same 2 octets country-codes as in Internet domains [@!IANADomains] based on [@!ISO3166-1] alpha-2 codes.</description>
+      <description lang="en">Nationality of the main composer of the item, mostly for classical music,
+in the Matroska countries form; see (#country-codes) on countries codes.</description>
```
mbunkus, line 272: "countries codes" should be "country codes", just like "language codes": you only pluralize the second word.

mbunkus (changes requested):
Apart from these nitpicks LGTM.

matroska_tags.xml
```diff
@@ -32,9 +32,9 @@ targeted item taken from another work of art. All tags in this list can be used
     </tag>
     <tag name="COUNTRY" class="Nesting Information" type="UTF-8">
       <description lang="en">The name of the country
-that is meant to have other tags inside (using nested tags) to country specific information about the item.
+that is meant to have other tags inside (using nested tags) to country specific information about the item,
+in the Matroska countries form; see (#country-codes) on country codes.
```
robUx4, line 36: `country-codes` is not part of the Tags document (yet)

JeromeMartinez: Possible to link from Tags document to the Matroska notes?

JeromeMartinez: Well, too difficult for me :). So as an intermediate step I just put full description in Tags document in order to avoid to block the merge of the PR for a file which is currently not intended for final review.
I hope that it is fine enough.

PR #675, opened 2022-09-27
Cleanup normative references

JeromeMartinez:
IMO normative references should be only for what is the most important to be supported by a parser/player, so "SHOULD" or "MUST". All with "MAY" could have informative references only.

Also the core document should be neutral about the transported formats, so no reference to a specific format in the element descriptions.

### DIFF
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -440,7 +440,7 @@ If this Element is used, then any Language Elements used in the same TrackEntry 
   </element>
   <element name="CodecID" path="\Segment\Tracks\TrackEntry\CodecID" id="0x86" type="string" minOccurs="1" maxOccurs="1">
     <documentation lang="en" purpose="definition">An ID corresponding to the codec,
-see [@!MatroskaCodec] for more info.</documentation>
+see [@?MatroskaCodec] for more info.</documentation>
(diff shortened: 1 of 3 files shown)


robUx4:
This should be split in several commits (or MR). For example turning a SHOULD into a MAY should be discussed on its own.

Also why remove the [@?DivXTrickTrack] which is not normative ?

JeromeMartinez:
WIP because some change may be to discuss

- DivX specific elements: IMO we should not have format specific Matroska elements, I suggest to make them neutral (no reference to DivX) and, if needed, move some details to the Codec doc saying how DivX use them.
- UncompressedFourCC should points to Codec doc with more details about this element.
- MasteringMetadata: added "MAY" in order to avoid to have CIE spec as normative, does it make sense?
- ContentEncodings: so many players don't support that, IMO it should be explicitly flagged as optional.

JeromeMartinez:
> This should be split in several commits (or MR). 

I kept there only the obvious (I hope!) changes.
- MatroskaCodec and MatroskaTags optional as discussed on the mailing list.
- WebVTT is as Divx --> specific format so optional.
- BZIP2 and LZO are with a "MAY" so optional.

robUx4 (approved):
LGTM

PR #676, opened 2022-09-27
Clarify ContentCompAlgo usage hints

JeromeMartinez:


ebml_matroska.xml
```diff
@@ -1191,7 +1191,8 @@ Each block **MUST** be decompressable even if no previous block is available in
     <documentation lang="en" purpose="definition">The compression algorithm used.</documentation>
     <documentation lang="en" purpose="usage notes">Compression method "1" (bzlib) and "2" (lzo1x) are lacking proper documentation on the format which limits implementation possibilities.
 Due to licensing conflicts on commonly available libraries compression methods "2" (lzo1x) does not offer widespread interoperability.
-Decoding implementations **MAY** support methods "1" and "2" as possible. The use of these compression methods **SHOULD NOT** be used as a default.</documentation>
+Encoding implementations **SHOULD NOT** use these compression methods by default.
+Decoding implementations **MAY** support methods "1" and "2" as possible, and **SHOULD** support other methods.</documentation>
```
robUx4, line 1195: `Matroska Reader` would be more appropriate than `Decoding implementations` as it has a definition at the beginning of the document. You could use `Matroska Writer` for the "encoding implementations" although it doesn't have a definition.

JeromeMartinez: Modified (and "Matroska Writer" definition added).

PR #677, opened 2022-09-27
Clarify ContentEncryption as optional

JeromeMartinez:
Not all players support ContentEncryption, let's avoid that players not supporting ContentEncryption are considered as non conforming.

### DIFF
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -1215,7 +1215,8 @@ the bytes that were removed from the beginning of each frames of the track.</doc
   </element>
   <element name="ContentEncryption" path="\Segment\Tracks\TrackEntry\ContentEncodings\ContentEncoding\ContentEncryption" id="0x5035" type="master" maxOccurs="1">
     <documentation lang="en" purpose="definition">Settings describing the encryption used.
-This Element **MUST** be present if the value of `ContentEncodingType` is 1 (encryption) and **MUST** be ignored otherwise.</documentation>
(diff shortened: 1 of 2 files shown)


ebml_matroska.xml
```diff
@@ -1215,7 +1215,8 @@ the bytes that were removed from the beginning of each frames of the track.</doc
   </element>
   <element name="ContentEncryption" path="\Segment\Tracks\TrackEntry\ContentEncodings\ContentEncoding\ContentEncryption" id="0x5035" type="master" maxOccurs="1">
     <documentation lang="en" purpose="definition">Settings describing the encryption used.
-This Element **MUST** be present if the value of `ContentEncodingType` is 1 (encryption) and **MUST** be ignored otherwise.</documentation>
+This Element **MUST** be present if the value of `ContentEncodingType` is 1 (encryption) and **MUST** be ignored otherwise.
+Decoding implementations **MAY** support encryption.</documentation>
```
robUx4, line 1219: Same remark, this should be `Matroska Readers`.

robUx4:
While it's debatable whether compression should be handled most of the time. Encryption is definitely optional, as 99% of existing players show.

ebml_matroska.xml
JeromeMartinez: I changed to `Matroska Player` as a `Matroska Reader` does not need to handle that (transporting the opaque packets, splitting them, etc, does not require the support of this feature).

robUx4 (approved):
LGTM

retokromer (approved):

PR #680, opened 2022-09-27
Add link to CIE 1931 Wikipedia page

JeromeMartinez:
Wikipedia page has more description about CIE-1931 than CIE web page.


robUx4 (approved):
LGTM

PR #681, opened 2022-10-05
Add ATRAC1 codec mapping

Sir68k:
This PR adds a mapping for ATRAC1 audio data. 

Related communication in the Celler mailing list: https://mailarchive.ietf.org/arch/msg/cellar/YWKreGPHFltjtps0Yz2JOILdWA4/.

### DIFF
--- a/codec_specs.md
+++ b/codec_specs.md
@@ -823,7 +823,7 @@ Codec ID: A_ATRAC/AT1
 
 Codec Name: Sony ATRAC1 Codec
 
-Description: The original ATRAC codec by Sony, mainly used in MiniDisc platforms. The core technical details on ATRAC1 can be found in [@!AtracAES]. An example encoder/decoder can be found at [@!atracdenc].
+Description: The original ATRAC codec by Sony, mainly used in MiniDisc platforms. The core technical details on ATRAC1 can be found in [@?AtracAES]. An example encoder/decoder can be found at [@?atracdenc].
 
 Initialization: None
 
--- a/codec_specs.md
+++ b/codec_specs.md
@@ -817,6 +817,16 @@ The `BlockAddId` of the `BlockMore` containing these data **MUST** be 1.
 
 Initialization: none
 
+### A_ATRAC/AT1
+
+Codec ID: A_ATRAC/AT1
+
+Codec Name: Sony ATRAC1 Codec
+
+Description: The original ATRAC codec by Sony, mainly used in MiniDisc platforms.
+
+Initialization: If the encoded audio track has a channel count other than `2` (STEREO), the `Private Data` contains a single unsigned byte that specifies the number of channels. If the channel count is `0` or the `Initialization` is set to `None`, the channel count will default to `2` (STEREO).
+
 ## Subtitle Codec Mappings
 
 ### S_TEXT/UTF8
--- a/codec_specs.md
+++ b/codec_specs.md
@@ -823,7 +823,7 @@ Codec ID: A_ATRAC/AT1
 
 Codec Name: Sony ATRAC1 Codec
 
-Description: The original ATRAC codec by Sony, mainly used in MiniDisc platforms. The core technical details on ATRAC1 can be found at https://www.minidisc.wiki/technology/atrac/aes. An example encoder can be found at https://github.com/dcherednik/atracdenc.
+Description: The original ATRAC codec by Sony, mainly used in MiniDisc platforms. The core technical details on ATRAC1 can be found in [@!AtracAES]. An example encoder/decoder can be found at [@!atracdenc].
 
 Initialization: None
 
--- a/codec_specs.md
+++ b/codec_specs.md
@@ -825,7 +825,7 @@ Codec Name: Sony ATRAC1 Codec
 
 Description: The original ATRAC codec by Sony, mainly used in MiniDisc platforms.
 
-Initialization: If the encoded audio track has a channel count other than `2` (STEREO), the `Private Data` contains a single unsigned byte that specifies the number of channels. If the channel count is `0` or the `Initialization` is set to `None`, the channel count will default to `2` (STEREO).
+Initialization: None
 
 ## Subtitle Codec Mappings
 
--- a/codec_specs.md
+++ b/codec_specs.md
@@ -823,7 +823,7 @@ Codec ID: A_ATRAC/AT1
 
 Codec Name: Sony ATRAC1 Codec
 
-Description: The original ATRAC codec by Sony, mainly used in MiniDisc platforms.
+Description: The original ATRAC codec by Sony, mainly used in MiniDisc platforms. The core technical details on ATRAC1 can be found at https://www.minidisc.wiki/technology/atrac/aes. An example encoder can be found at https://github.com/dcherednik/atracdenc.
 
 Initialization: None
 
--- a/rfc_backmatter_codec.md
+++ b/rfc_backmatter_codec.md
@@ -93,3 +93,21 @@
     <date day="04" month="April" year="2021" />
   </front>
 </reference>
+
+<reference anchor="AtracAES" target="https://www.minidisc.wiki/technology/atrac/aes">
+  <front>
+    <title>ATRAC: Adaptive Transform Acoustic Coding for MiniDisc</title>
+    <author>
+      <organization>Sony Corporate Research Laboratories</organization>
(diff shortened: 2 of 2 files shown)


codec_specs.md
```diff
@@ -817,6 +817,16 @@ The `BlockAddId` of the `BlockMore` containing these data **MUST** be 1.
...
 
+### A_ATRAC/AT1
+
+Codec ID: A_ATRAC/AT1
+
+Codec Name: Sony ATRAC1 Codec
+
+Description: The original ATRAC codec by Sony, mainly used in MiniDisc platforms.
+
+Initialization: If the encoded audio track has a channel count other than `2` (STEREO), the `Private Data` contains a single unsigned byte that specifies the number of channels. If the channel count is `0` or the `Initialization` is set to `None`, the channel count will default to `2` (STEREO).
```
JeromeMartinez, line 828: Channel count is already provided by Matrska dedicated element, why is it needed to repeat it in private data part?

mbunkus: I agree. If signalling the number of channels is the only thing it's used for, just leave it empty & use the existing element.

mbunkus (changes requested):

codec_specs.md
Sir68k: It seems I misunderstood the purpose of the existing `Channels` attribute. Just to confirm: the Tracks -> TrackEntry -> Audio -> Channels field can be used to express the number of channels encoded in the actual audio data (encoded in a way specific to the used codec), and does not refer to a Matroska-specific way of potentially encoding channels (e.g. using additional clusters/blocks)?

Sir68k: I've updated the PR to remove the Initialization as requested.

JeromeMartinez: > the Tracks -> TrackEntry -> Audio -> Channels field can be used to express the number of channels encoded in the actual audio data (encoded in a way specific to the used codec), and does not refer to a Matroska-specific way of potentially encoding channels (e.g. using additional clusters/blocks)?

Exactly. And it is mandatory (if not in the file it means that it is 1).

codec_specs.md
```diff
@@ -817,6 +817,16 @@ The `BlockAddId` of the `BlockMore` containing these data **MUST** be 1.
 
 Initialization: none
 
+### A_ATRAC/AT1
+
+Codec ID: A_ATRAC/AT1
+
+Codec Name: Sony ATRAC1 Codec
+
+Description: The original ATRAC codec by Sony, mainly used in MiniDisc platforms.
```
JeromeMartinez, line 826: May you add a link to a a page describing a bit the format? Even if not an official spec, at least it provides developers a starting point.

Sir68k: I added a link to the original ATRAC1 paper, and a reference encoder.

codec_specs.md
```diff
@@ -817,6 +817,16 @@ The `BlockAddId` of the `BlockMore` containing these data **MUST** be 1.
 
 Initialization: none
 
+### A_ATRAC/AT1
+
+Codec ID: A_ATRAC/AT1
+
+Codec Name: Sony ATRAC1 Codec
+
+Description: The original ATRAC codec by Sony, mainly used in MiniDisc platforms. The core technical details on ATRAC1 can be found at https://www.minidisc.wiki/technology/atrac/aes. An example encoder can be found at https://github.com/dcherednik/atracdenc.
```
robUx4, line 826: You should use a more fornal way to link to external resources, in the form of [@?AtracAes] and a description in `rfc_backmatter_codec.md`. The same can be done for the example. Although IMO it can be dropped.

Looking at the linked documents. They don't provide an actual description of the bitstreams that can be expected to be found in each `A_ATRAC/AT1` frame. Is there a more bitstream based documentation that we could link to ? For example the [code you mention on CELLAR](https://github.com/cybercase/netmd-js/blob/ada9272a6f670ff7813b4405fd8444809c5ceb67/src/utils.ts#L339) is closer to what could be used to decypher the bitstream but it's not a documentation.

codec_specs.md
robUx4: Where is the channel information found in the bitstream ? Theoretically we should show say that this information MUST (normative) match the channel field in Matroska. Although I think to save space we should have a general rule in the codec document that states such obvious thing. We still need to know where to find that information in the codec data (if there's one and not relying on a container header).

robUx4: It seems [this library](https://github.com/linux-minidisc/linux-minidisc/blob/43a01190f819e44c5709017b889be9bf8766b989/libhimd/trackindex.c#L134) has more information on all the bitstream (the TypeScript lib you mention seems to borrow from that project). But that's not exactly a documentation.

JeromeMartinez: > Where is the channel information found in the bitstream ?

My understanding, because the first version of the PR was containing the channel count in private data, is that this piece of information is not in the bitstream. Not a problem for the spec here, or hat do I miss?

>  Although I think to save space we should have a general rule in the codec document that states such obvious thing.

Well, it is the idea for all duplicated pieces of information, nothing specific here.

codec_specs.md
JeromeMartinez: > You should use a more fornal way to link to external resources,

Hmmm... We have a lot of this method in the doc, e.g. [for V_MS/VFW/FOURCC](https://github.com/ietf-wg-cellar/matroska-specification/blob/master/codec_specs.md#v_msvfwfourcc).

> For example the [code you mention on CELLAR](https://github.com/cybercase/netmd-js/blob/ada9272a6f670ff7813b4405fd8444809c5ceb67/src/utils.ts#L339) is closer to what could be used to decypher the bitstream but it's not a documentation.

We also have a lot of similar references lie that e.g. [for V_REAL/RV40](https://github.com/ietf-wg-cellar/matroska-specification/blob/master/codec_specs.md#v_realrv40).

IMO we should not request a lot more for just reserving an identifier, we understand well that this is ATRAC1 and player willing to support this format would have what they need for decoding it.

robUx4: > Hmmm... We have a lot of this method in the doc, e.g. [for V_MS/VFW/FOURCC](https://github.com/ietf-wg-cellar/matroska-specification/blob/master/codec_specs.md#v_msvfwfourcc).

The document is far from finished and this is the kind of thing we have to fix. So it's better if we don't introduce more of these.

> IMO we should not request a lot more for just reserving an identifier

IMO it's not good enough to have just an identifier and then good luck to find the relevant information to parse the data. This is especially true for rare codecs. The goal here is clearly sustainability and with a rare codec like that you can't expect people to easily be able to know how to use it.

robUx4: That being said, I'm fine with having a first step to reserve the codec ID so people are aware of it.

thinkbrown: Hi, I've been contributing to the reverse engineering project mentioned on the Cellar mailing list and can hopefully shed a little bit of light on where our understanding of ATRAC is. 

So far, our best sources of information have been the ATRAC decoder [included](https://github.com/FFmpeg/FFmpeg/blob/master/libavcodec/atrac1.c) in FFmpeg and the now expired Sony US patent [5301205](http://www.minidisc.org/patents/pdfs/US05301205.pdf). To the best of my knowledge, Sony never released a formal specification for the bitstream.

The ATRAC decoder in FFmpeg is - as far as we're aware - correct, and we'd like to make it easier to leverage that without the aforementioned issues with the AEA container.

Sir68k: Hi,

As thinkbrown has mentioned, there does not exist any technical documentation on the bitstream format that is publicly available (as far as we know). The closest is the decoder source code in ffmpeg and atracdenc. The goal of getting the codec ID is to make the format more accessible. Potentially we could write this documentation, given time, but right now our focus is on the reverse engineering process and audio archival.

I will take a look at the referencing system and try to submit an update as soon as possible.

Sir68k: Updated the references to conform to the suggested formatting.

codec_specs.md
Sir68k: Sorry, missed this, the information is indeed not present in the bitstream itself. It's something that needs to be encoded in the container.

robUx4:
The two `[@!]` (normative link) should be `[@?]` (non normative). Especially since these links are not exactly proper specs.

Sir68k:
Updated the links so that they are non-normative.

retokromer (approved):

PR #684, opened 2022-11-13
Support of an unique SMPTE ST 12 Timecode in track header

JeromeMartinez:
Several container formats, e.g. MXF or QuickTime, don't have SMPTE ST 12 Timecode values per frame, they have a start time code and a frame rate, then other time codes are computed from that.

This PR adds the support of a similar way to store timecodes in Matroska.

As SMPTE ST 12 does not indicate the corresponding frame rate, we provide it in the private data. frame rate num/den is preferred over an integer because we need the precise source (constant) frame rate for computing the corresponding timecode.

### DIFF
--- a/block_additional_mappings/smpte-st12-1-timecode.md
+++ b/block_additional_mappings/smpte-st12-1-timecode.md
@@ -53,4 +53,62 @@ The BlockAddIDName value reserved for timecode is "SMPTE ST 12-1 timecode".
 
 ### BlockAddIDExtraData
 
-BlockAddIDExtraData is unused within this block additional mapping.
+BlockAddIDExtraData is a sequence of EBML elements.
+
+#### Start Timecode
+
+id / type:
+    0x???? / binary
+
+path:
+    \\
+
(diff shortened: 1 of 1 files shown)


block_additional_mappings/smpte-st12-1-timecode.md
```diff
@@ -53,4 +53,62 @@ The BlockAddIDName value reserved for timecode is "SMPTE ST 12-1 timecode".
...
+    0x???? / binary
+
+path:
+    \\
+
+maxOccurs:
+    1
+
+definition:
+    If present, indicates that Block Additional Mapping is not present and that the stored bytes represent the start timcode.
```
retokromer, line 70: nit: timcode > timecode

JeromeMartinez:
> Should we first implement the EBML Rational type?

Would be better, for sure.

robUx4:
>  Should we first implement the EBML Rational type?

This might be a while. And since there's no agreed technical solution, you should not create such a file for now.

PR #700, opened 2022-12-29
our UUIDs are not RFC4122 UUIDs

robUx4:
It turns out UUIDs are not just random values, but a lot of bits have a meaning. That's not what we have used in Matroska so far.

Partially reverts #620 and #621. We keep the UUID names, for now.

### DIFF
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -27,7 +27,7 @@
     <extension type="webmproject.org" webm="1"/>
   </element>
   <element name="SegmentUUID" path="\Segment\Info\SegmentUUID" id="0x73A4" type="binary" range="not 0" length="16" maxOccurs="1">
-    <documentation lang="en" purpose="definition">A randomly generated unique ID to identify the Segment amongst many others (128 bits).</documentation>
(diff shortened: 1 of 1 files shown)


retokromer (approved):

hubblec4:
Revert everything fully. It is confusing to use UUID as name but the UID is not an UUID.

robUx4:
We have the same format/size as a regular UUID. So any code parsing/displaying actual UUIDs can be used with ours as well. They just don't have the same "meaning". I think it's good to keep the connection.

ebml_matroska.xml
```diff
@@ -27,17 +27,15 @@
     <extension type="webmproject.org" webm="1"/>
   </element>
   <element name="SegmentUUID" path="\Segment\Info\SegmentUUID" id="0x73A4" type="binary" range="not 0" length="16" maxOccurs="1">
-    <documentation lang="en" purpose="definition">A randomly generated unique ID to identify the Segment amongst many others (128 bits).
-It is effectively a Universally Unique IDentifier stored in binary form [@!RFC4122].</documentation>
+    <documentation lang="en" purpose="definition">A randomly generated unique ID to identify the Segment amongst many others (128 bits).</documentation>
```
mcr, line 30: make it UUID version 4.

robUx4: I kept the default to all random but allowed UUID v4 where some bits a not random too.

hubblec4 (approved):

retokromer (approved):

Issue #721, opened 2023-01-08
Replace the fourcc.org link


robUx4:
The domain has been stolen by an online casino.

PR #722, opened 2023-01-08
mention the UncompressedFourCC values don't have an official list

robUx4:
Link to fourcc.org via web.archive.org as the website is gone. Also add some more official MS values that can be found in AVI, as P010-P016 is missing from all online lists I could find.

Maybe we should start a IANA registry for that in the Codec Spec document. Describing each value might be tricky...


robUx4:
Fixes #721

PR #724, opened 2023-01-12
Add ARIB B-24 subtitle codec mapping

rcombs:
This adds a mapping for ISDB/ARIB STD-B24 caption data.

### DIFF
--- a/subtitles.md
+++ b/subtitles.md
@@ -677,3 +677,35 @@ in segment 7.2 "Syntax and semantics of the subtitling segment" of ETSI EN 300 7
 Each Matroska Block **SHOULD** have a Duration indicating how long the DVB Subtitle Segments
 in that Block **SHOULD** be displayed.
 
+## ARIB (ISDB) subtitles
+
+The specifications for the ARIB B-24 subtitle bitstream format (short: ARIB subtitles)
+and its storage in MPEG transport streams can be found in the documents
+[@!ARIB.STD-B24], [@!ARIB.STD-B10], and [@!ARIB.TR-B14].
+
+### Storage of ARIB subtitles
+
+#### CodecID
+
+The CodecID to use is `S_ARIBSUB`.
+
+#### CodecPrivate
+
+The CodecPrivate element is three bytes long and has the following structure:
+
+*    1 byte: component tag (bit string, left bit first)
+*    2 bytes: data component ID (bit string, left bit first)
+
+The semantics of the component tag are the same as those described in [@!ARIB.STD-B10], part 2, Annex J.
+The semantics of the data component ID are the same as those described in [@!ARIB.TR-B14], fascicle 2, Vol. 3, Section 2, 4.2.8.1.
+
+#### Storage of ARIB subtitles in Matroska Blocks
+
+Each Matroska Block consists of one or more ISDB Caption Data Groups as described
+in chapter 9 "Transmission of caption and superimpose" of [@!ARIB.STD-B24], volume 1, part 3.
+All of the Caption Statement Data Groups in a given Matroska Track **MUST** use the same language index.
+
+A Data Group is normally shown until a subsequent Group provides instructions to clear it.
+Therefore the Matroska Block **SHOULD NOT** have a Duration.
+A player **SHOULD** display a Data Group within a Matroska Block until its internal duration elapses,
+or until a subsequent Data Group removes it.
--- a/codec_specs.md
+++ b/codec_specs.md
@@ -925,6 +925,15 @@ Description: A subtitle format developed for ogg. The mapping for Matroska is de
 on the [Xiph wiki](http://wiki.xiph.org/index.php/OggKate#Matroska_mapping).
 As for Theora and Vorbis, Kate headers are stored in the private data as xiph-laced packets.
 
+### S_ARIBSUB
+
+Codec ID: S_ARIBSUB
+
+Codec Name: ARIB STD-B24 subtitles
+
+Description: This is the textual subtitle format used in the ISDB/ARIB broadcasting standard.
+For more information, see (#arib-isdb-subtitles) on ARIB (ISDB) subtitles.
+
 ## Button Codec Mappings
 
 ### B_VOBBTN
--- a/rfc_backmatter_codec.md
+++ b/rfc_backmatter_codec.md
@@ -106,8 +106,38 @@
 
 <reference anchor="atracdenc" target="https://github.com/dcherednik/atracdenc">
   <front>
-    <title>atracdenc - ATRAC1 and ATRAC3 Decoder/Encoder</title> 
+    <title>atracdenc - ATRAC1 and ATRAC3 Decoder/Encoder</title>
     <author initials="D." surname="Cherednik" fullname="Daniil Cherednik"></author>
     <date year="2022" month="October" day="12"/>
+  </front>
+</reference>
+
+<reference anchor="ARIB.STD-B24" target="https://www.arib.or.jp/english/std_tr/broadcasting/desc/std-b24.html">
+  <front>
+    <title>Data Coding and Transmission Specification for Digital Broadcasting</title>
+    <author>
+      <organization>ARIB</organization>
+    </author>
+    <date year="2022" month="October" day="6"/>
+  </front>
+</reference>
+
+<reference anchor="ARIB.STD-B10" target="https://www.arib.or.jp/english/std_tr/broadcasting/desc/std-b10.html">
+  <front>
+    <title>Service Information for Digital Broadcasting System</title>
+    <author>
+      <organization>ARIB</organization>
+    </author>
+    <date year="2019" month="December" day="5"/>
+  </front>
+</reference>
+
+<reference anchor="ARIB.TR-B14" target="https://www.arib.or.jp/english/std_tr/broadcasting/desc/tr-b14.html">
+  <front>
+    <title>Operational Guidelines for Digital Terrestrial Television Broadcasting</title>
+    <author>
+      <organization>ARIB</organization>
+    </author>
+    <date year="2022" month="October" day="6"/>
   </front>
 </reference>


subtitles.md
```diff
@@ -677,3 +677,36 @@ in segment 7.2 "Syntax and semantics of the subtitling segment" of ETSI EN 300 7
 Each Matroska Block **SHOULD** have a Duration indicating how long the DVB Subtitle Segments
 in that Block **SHOULD** be displayed.
 
+## ARIB (ISDB) subtitles
+
+The specifications for the ARIB B-24 subtitle bitstream format (short: ARIB subtitles)
+and its storage in MPEG transport streams can be found in the documents
+"ARIB STD-B24 - Data Coding and Transmission Specification for Digital Broadcasting",
+"ARIB STD-B10 - Service Information for Digital Broadcasting System",
+and "ARIB TR-B14 - Operational Guidelines for Digital Terrestrial Television Broadcasting".
```
robUx4, line 686: You should have normative links to these specifications ? They should be added to `rfc_backmatter_codec.md`.

The link being probably https://www.arib.or.jp/english/std_tr/broadcasting/desc/std-b24.html and such.

subtitles.md
```diff
@@ -677,3 +677,36 @@ in segment 7.2 "Syntax and semantics of the subtitling segment" of ETSI EN 300 7
...
+The semantics of the component tag are the same as those described in ARIB STD-B10, part 2, Annex J.
+The semantics of the data component ID are the same as those described in ARIB TR-B14, fascicle 2, Vol. 3, Section 2, 4.2.8.1.
+
+#### Storage of ARIB subtitles in Matroska Blocks
+
+Each Matroska Block consists of one or more ISDB Caption Data Groups as described
+in chapter 9 "Transmission of caption and superimpose" of ARIB STD-B24 volume 1 part 3.
+
+A Data Group is normally shown until a subsequent Group provides instructions to clear it.
+Therefore the Matroska Block **SHOULD** have no Duration.
```
robUx4, line 710: I think it's a **MUST** unless there is a case where you want it. It could be a SHOULD if it's added as an indication of where to expect the clear instruction. If so, it should be mentioned.

subtitles.md
```diff
@@ -677,3 +677,36 @@ in segment 7.2 "Syntax and semantics of the subtitling segment" of ETSI EN 300 7
...
+*    1 byte: component tag (bit string, left bit first)
+*    2 bytes: data component ID (bit string, left bit first)
+
+The semantics of the component tag are the same as those described in ARIB STD-B10, part 2, Annex J.
+The semantics of the data component ID are the same as those described in ARIB TR-B14, fascicle 2, Vol. 3, Section 2, 4.2.8.1.
+
+#### Storage of ARIB subtitles in Matroska Blocks
+
+Each Matroska Block consists of one or more ISDB Caption Data Groups as described
+in chapter 9 "Transmission of caption and superimpose" of ARIB STD-B24 volume 1 part 3.
```
robUx4, line 707: It sounds odd to have more than one data group in one Block, and by that I think you mean a frame. Does a "caption data group" just means it's a series of character ? Or they can be different languages ?

subtitles.md
```diff
@@ -677,3 +677,36 @@ in segment 7.2 "Syntax and semantics of the subtitling segment" of ETSI EN 300 7
...
+
+A Data Group is normally shown until a subsequent Group provides instructions to clear it.
+Therefore the Matroska Block **SHOULD** have no Duration.
+A player **SHOULD** display a Data Group within a Matroska Block until its internal duration elapses,
+or until a subsequent Data Group removes it.
```
robUx4, line 712: The problem is that Matroska on its own does not know about internals of codec, and should NOT. If you want to handle cutting files properly, that means relying on a Block duration. In that case the group removal MUST be translated to a Matroska duration. It's up to the player to transform it back to an actual format an ARIB B-24 parser understands, if necessary.

In that case there's no need for this sentence at all since it falls back to the general rule of Block duration.

robUx4 (changes requested):

subtitles.md
rcombs: "Caption data group" is defined in ARIB STD-B24 volume 1 part 3 chapter 9; an MPEGTS PES packet contains a "caption data", which consists of one or more "caption data groups", which are defined structures. I don't know why they call them that, but it's their terminology, not mine.

Technically different groups can be in different languages, but in practice there seems to always be one, with separate languages being separate PES streams (this is recommended by the B24 spec). The groups can be multiple caption statements, or a mix of "management" data (which configures display modes and such) and caption statements.

rcombs: Also, I got the "Each Matroska Block consists of one or more [...]" language from the DVB section of this file.

subtitles.md
rcombs: "Removing it" generally means that a later packet includes a command that clears all previous lines from the screen. Most groups don't have a defined duration at all. See also the PGS section of this file, which specifies that "A Segment is normally shown until a subsequent Segment is encountered."

subtitles.md
robUx4: A Matroska track cannot have multiple language. If there is more than one language then it needs to be split, or set to `undefined` language, which sucks when it comes to selecting proper subtitles.

There should be a description on how to handle the language mapping (they probably have some code, that have to be mapped to our strings) and what to do in case there's more than one language involved.

I know others don't do it, that's the legacy of what we had so far. But going forward we should do things correctly, at least for new things. For audio and video codecs there are a lot of fields that need to be mapped to Matroska (see the AV1 spec) at least for subtitles that's just the language and what the extra data contain.

rcombs: I could put a line in here that says that all data groups that contain caption statements must have the same language ID (or specifically language ID 0)? In practice, I doubt anything will enforce that (particularly since multilingual streams seem to just be a spec quirk that doesn't actually get used in the field), but I can stick it in there to make it clear that players don't have to support it (ffmpeg doesn't even support it in mpegts).

subtitles.md
robUx4: The PGS description has more details on what to do when the muxer chooses not put a duration. It is bogus though "a player MUST display a Segment within a Matroska Block until the next Segment is encountered.". The Segment boundaries are arbitrary, a remuxing of the same file will produce a different playback experience. I think the text should be "PGS Segment" and then it makes sense.

Then there's a proper explanation on what a "SHOULD have a duration" means, when there is one and when there isn't one. At least you should have something similar.

It's still equally broken when there is no duration. If you cut a portion of audio/video that also contained the Clear command, then your subtitle can be displayed for hours.

I understand it's easier to mux without changing anything of the original bits, but it produces bogus Matroska files. It would also be hard to find the duration preemptively without preparsing the original source first, or editing the duration of previous blocks while muxing. I don't know of any framework capable of doing that easily.

subtitles.md
robUx4: That would be nice. At least if the case happens that means the file is bogus, and coders don't have to go to great length to make it work. Or just by reading the specs wonder how they are supposed to handle such a case.

subtitles.md
robUx4: Not sure of mkvmerge, maybe it can preparse files to get a list of subtitle entries and their duration (can be done with PGS as well). cc @mbunkus 

I don't think libavformat would be capable of that.

rcombs: > It would also be hard to find the duration preemptively without preparsing the original source first, or editing the duration of previous blocks while muxing. I don't know of any framework capable of doing that easily.

Yeah, that's the thing; if I required a duration here, or explicitly said that packets without specified durations always end with the next packet, I'd be making it dramatically harder to simply remux (a common operation) in favor of making it simpler to clip and concatenate (a rarer operation). It would also be impossible to do in realtime, which is a requirement for some matroska applications.

Note that all of these issues also apply to ATSC closed captions, though those currently exist in Matroska largely by being carried within MPEG2, H.264, or HEVC bitstreams (e.g. as SEIs).

mbunkus: > Not sure of mkvmerge, maybe it can preparse files to get a list of subtitle entries and their duration (can be done with PGS as well). cc @mbunkus

`mkvmerge` cannot pre-parse files as that would require parsing multiple tens of GB of data in the case of MPEG TS files (Blu-rays), and that's the primary source of PGS subs. At the moment `mkvmerge` simply disregards this particular issue & doesn't care about the duration. Yes, I'm very aware that this is problematic, both with playback and operations such as splitting & appending. You could say `mkvmerge` doesn't support those fully at this time.

subtitles.md
jeeb: Asked around regarding if anyone had seen the usage of two subtitle languages in a single PID.

> I have added support for switching between the two available languages in my playback plugin, but I have not seen this being utilized.

> There are rumors regarding the feature being utilized by Disney Channel, but I've not been able to verify this.

So this is kind of similar to US style closed captions or DVB teletext, where you can have multiple different "alternatives" within a single stream, but at least its use seems to be quite rare.

rcombs:
I've re-pushed this with the following changes:
- Added spec backmatter links
- Indicated that all caption statements in a track must share a single language index
  - This probably won't be enforced in practice, but having the MUST in there at least indicates players don't have to care about it, which should be fine since it seems to be extremely rare
- Indicated that durations MUST be absent

I believe this leaves only the duration question; we seem to have concluded that it's not practical to expect players to precompute them up-front, so do we want to stick with the existing text there?

subtitles.md
robUx4: I guess that ship has sailed with PGS. However I don't think a **MUST** have no duration (should be **MUST NOT** have a duration is a appropriate. If some tools want to post-mux/remux the file to add durations to help editing/seeking, we should not forbid it.

I suggest: `**SHOULD NOT** have a duration, unless it is known at the time of muxing.`

subtitles.md
robUx4: The last edit where it says it should not happen seems OK. If it happens, people are on their own to deal with it, ie split the tracks.

robUx4 (changes requested):
almost there, with the MUST about the duration that I think should be turned into SHOULD NOT.

rcombs:
Changed to SHOULD NOT.

subtitles.md
robUx4: We'll probably have to edit again when we publish it, as SHOULD and SHOULD NOT rules have to explain in which it is not the case. But that's good enough for now.

robUx4 (approved):

robUx4:
Thanks for your work :)

PR #731, opened 2023-03-05
make top-level elements ID 4-octets long

robUx4:
cc @mbunkus as it will change the libmatroska code (unlikely to have any effect in the real world)


hubblec4 (approved):
LGTM

retokromer (approved):

PR #732, opened 2023-03-05
don't use the range attribute to forbid zero SegmentUID

### LINKED ISSUE #730: Binary element range values
Currently we have only two Binary elements which have a `range` restriction.

`SegmentUUID` = "not 0"
`ChapterSegmentUUID` = ">0"

EBML specs say that the `range` is used for numerical EBML types (Unsigned Integer, Signed Integer, Float, and Date).
This excludes the Binary element type.

It makes also no really sense to check a numerical value against  binary data.

The `range` value "not 0" is in my opinion not correct. It allows negative numbers.
OK, negative numbers for binary data makes also no sense.

And why are the `range` values of this both elements different?

For Binary `range` values I would prefer something like this:

`SegmentUUID` = "not all bits 0"
`ChapterSegmentUUID` = "not all bits 0"

And we have to change the EBML specs for the `range` attribute.

robUx4:
Fixes #730

### DIFF
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -26,9 +26,10 @@
     <documentation lang="en" purpose="definition">Contains general information about the Segment.</documentation>
     <extension type="webmproject.org" webm="1"/>
   </element>
-  <element name="SegmentUUID" path="\Segment\Info\SegmentUUID" id="0x73A4" type="binary" range="not 0" length="16" maxOccurs="1">
+  <element name="SegmentUUID" path="\Segment\Info\SegmentUUID" id="0x73A4" type="binary" length="16" maxOccurs="1">
     <documentation lang="en" purpose="definition">A randomly generated unique ID to identify the Segment amongst many others (128 bits). It is equivalent to a UUID v4 [@!RFC4122] with all bits randomly (or pseudo-randomly) chosen.  An actual UUID v4 value, where some bits are not random, **MAY** also be used.</documentation>
-    <documentation lang="en" purpose="usage notes">If the Segment is a part of a Linked Segment, then this Element is **REQUIRED**.</documentation>
+    <documentation lang="en" purpose="usage notes">If the Segment is a part of a Linked Segment, then this Element is **REQUIRED**.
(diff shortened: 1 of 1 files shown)


retokromer (approved):

ebml_matroska.xml
```diff
@@ -1452,7 +1453,7 @@ the last frame it includes, especially for the `ChapterAtom` using the last fram
 When disabled, the movie **SHOULD** skip all the content between the TimeStart and TimeEnd of this chapter; see (#chapter-flags) on Chapter flags.</documentation>
     <extension type="other document" spec="control-track"/>
   </element>
-  <element name="ChapterSegmentUUID" path="\Segment\Chapters\EditionEntry\+ChapterAtom\ChapterSegmentUUID" id="0x6E67" type="binary" range="&gt;0" length="16" maxOccurs="1">
+  <element name="ChapterSegmentUUID" path="\Segment\Chapters\EditionEntry\+ChapterAtom\ChapterSegmentUUID" id="0x6E67" type="binary" length="16" maxOccurs="1">
     <documentation lang="en" purpose="definition">The SegmentUUID of another Segment to play during this chapter.</documentation>
     <documentation lang="en" purpose="usage notes">The value **MUST NOT** be the `SegmentUUID` value of the `Segment` it belongs to.</documentation>
     <implementation_note note_attribute="minOccurs">ChapterSegmentUUID **MUST** be set (minOccurs=1) if ChapterSegmentEditionUID is used; see (#medium-linking) on medium-linking Segments.</implementation_note>
```
hubblec4, line 1459: I miss here the restriction that 1 bit must be 1.

robUx4: > The value **MUST NOT** be the `SegmentUUID` value of the `Segment` it belongs to.

This is a valid `SegmentUUID` elsewhere, so it has exactly the same constraints.

hubblec4 (approved):

ebml_matroska.xml
robUx4: I meant to quote that one

> The SegmentUUID of another Segment to play during this chapter

It's better to keep the constraint in fewer locations in case they need to be adjusted.

PR #733, opened 2023-03-05
Fix EBML signed VINT encoding description

### LINKED ISSUE #726: EBML Lacing details are incorrect
Hello! I'm opening this issue to describe a problem in the specification documentation related to this section: 

https://github.com/ietf-wg-cellar/matroska-specification/blob/bdb0b3c01a13537ad6f60d90ba80cf89ff2360c2/notes.md#ebml-lacing

The documentation states:
> The first frame size is encoded as an EBML Unsigned Integer Element value. The other frame sizes are encoded as a difference with the previous frame size as EBML Signed Integer Element values. That corresponds to an EBML Data Size Values with two’s complement notation with the leftmost bit being the sign bit as found in [@!RFC8794], giving this range of values:

The part I'd like to point out in the above excerpt is that subsequent frame sizes should be encoded as EBML Signed Integer Element values, which use *two's complement notation*.

The document then follows with a table showing an example encoding for a block with frame sizes of 800, 500, and 1000.  The relevant part here is the size of the second frame:

> 7-8 | 0x5E 0xD3 | Size of the second frame (500 - 800 = -300 = - 0x12C + 0x1FFF + 0x4000)

The size expressed as a decimal is definitely "-300".  However, the binary output shown **is not -300 in two's complement notation**, and the formula in parenthesis **does not result in two's complement notation**.  The correct binary for -300 (in two's complement notation using 14 bits, which is the number of usable bits for the value of the vint) is {11 1110 1101 0100}.  Prepending the vint marker {01} for two value bytes gives us: 0x7ED4.

Is there a consensus on how ebml lacing should actually be represented?  If so, it'd be great to update the documentation to reflect that correctly.

Handy calculator for two's complement: https://www.omnicalculator.com/math/twos-complement

robUx4:
Fixes #726

### DIFF
--- a/notes.md
+++ b/notes.md
@@ -208,21 +208,20 @@ The Block data with laced frames is stored as follows:
 * Lacing size of each frame except the last one.
 * Binary data of each frame consecutively.
 
-The first frame size is encoded as an EBML Variable-Size Integer value.
-The other frame sizes are encoded as a difference with the previous frame size as EBML Signed Integer Element values.
-That corresponds to an EBML Data Size Values with two's complement notation with the leftmost bit being the sign bit as found in [@!RFC8794],
-giving this range of values:
+The first frame size is encoded as an EBML Variable-Size Integer value, also known as VINT in [@!RFC8794].
+The remaining frame sizes are encoded as signed values using the difference between the frame size and the previous frame size.
+These signed values are encoded as VINT, with a mapping from signed to unsigned numbers.
+Decoding the unsigned number stored in the VINT to a signed number is done by subtracting 2^((7*n)-1)^-1, where `n` is the octet size of the VINT.
 
-Bit Representation                                                          | Value
-:---------------------------------------------------------------------------|:-------
-1xxx xxxx                                                                   | value -(2^6^-1) to 2^6^-1 (ie 0 to 2^7^-2 minus 2^6^-1, half of the range)
-01xx xxxx  xxxx xxxx                                                        | value -(2^13^-1) to 2^13^-1
-001x xxxx  xxxx xxxx  xxxx xxxx                                             | value -(2^20^-1) to 2^20^-1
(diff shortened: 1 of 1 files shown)


robUx4:
The bottom range value was actually correct. However the upper boundary seems to allow 1 more value that we used to allow. This is **not** what the code actually does https://github.com/ietf-wg-cellar/matroska-specification/issues/726#issuecomment-1445148868

But encoding 2<sup>6</sup> (64) fits in the 7 bits we have: 2<sup>6</sup> + (2<sup>6</sup>-1) = 127 = `0111 1111`.

robUx4:
That comes down to the libebml code that is not allowing this https://github.com/Matroska-Org/libebml/blob/a81936929f9c2c21e0ed0e11476df42a01111eb8/src/EbmlElement.cpp#L93

It should have been `if (Length > -64 && Length <= 64)`.

So we have to mention that in the text as well. Because libmatroska will not be able to read that `1111 1111` as 64.

robUx4:
Actually it cannot write that value, but it can read the value properly https://github.com/Matroska-Org/libebml/blob/a81936929f9c2c21e0ed0e11476df42a01111eb8/src/EbmlElement.cpp#L189. The size read will be 1 octet and so 63 will be substracted from 127.

This is also the case for libavformat where the substracted value depends on the size read, not the value itself: https://github.com/FFmpeg/FFmpeg/blob/3153c441e1d980ff9931ed26a0e01b4e366ac521/libavformat/matroskadec.c#L3249

So I'd just keep the spec as done here. `libmatroska` could be optimized to write 64 in 1 octet (and that code moved out of libebml for good)

hubblec4 (approved):

retokromer (approved):

robUx4:
Fixed the formula, it's actually 2<sup>(7*n)-2</sup>-1

robUx4:
Merging as the reading code (of libmatroska/libebml) has been verified to be able to read the upper boundaries we were not writing so far.

PR #736, opened 2023-03-05
Misc text fixes following the gen-art review

robUx4:


hubblec4 (approved):

rfc_backmatter_matroska.md
```diff
@@ -149,6 +149,17 @@
...
 
+<reference anchor="ISO9899" target="https://www.iso.org/standard/57853.html">
+  <front>
+    <title>Information technology -- Programming languages -- C</title>
+    <author>
+      <organization>International Organization for Standardization</organization>
+    </author>
+    <date year="2011" />
+  </front>
+  <seriesInfo name="ISO/IEC" value="9899:2011" />
```
retokromer, line 160: The current version is ISO/IEC 9899:2018, but possibly you need C11.

robUx4: I copied the reference from the EBML RFC. IMO it's better if they reference the same document. And soon there will be C23. We're not going to need to update again. What was right in 2011 is still OK now.

retokromer: Thank you! I understand that. I actually use C11 also for my own software packages …

retokromer (approved):

PR #739, opened 2023-03-12
Remove subtrack PART from TargetTypeValue 20

### LINKED ISSUE #738: TargetType has "PART" twice in the enum values
I can remember me that in the past the `label` values was the same like the `enum` vaules.

For example:
`enum`: COLLECTION -> `label`: Collection

Now is the `label`: TargetTypeVaule 70.

TargetTypeValue 70 should be set to the `documentation` -> `usage notes`.

But the question still is, can `enum` values be used? [String element range values](https://github.com/ietf-wg-cellar/matroska-specification/issues/737)

And a second issue is that the `enum` value "PART" occurs twice.
This value can be used if the `TargetTypeValue` value is 40 or 20.

I can not implement an Enum type with duplicated values (Lazarus), and I guess in c++ it is also not possible.

robUx4:
fixes  #738

### DIFF
--- a/ebml_matroska.xml
+++ b/ebml_matroska.xml
@@ -1604,7 +1604,7 @@ If empty or omitted, then the Tag describes everything in the Segment.</document
       <enum value="30" label="TRACK / SONG / CHAPTER">
         <documentation lang="en" purpose="definition">The common parts of an album or movie.</documentation>
       </enum>
-      <enum value="20" label="SUBTRACK / PART / MOVEMENT / SCENE">
+      <enum value="20" label="SUBTRACK / MOVEMENT / SCENE">
         <documentation lang="en" purpose="definition">Corresponds to parts of a track for audio (like a movement).</documentation>
       </enum>
       <enum value="10" label="SHOT">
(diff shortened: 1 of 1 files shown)


retokromer (approved):

hubblec4:
Was the value "PART" for "TargeTypeValue 20" invalid?

If not, then it should not removed, because all existing files which uses this combination of values(20 = PART), then are invalid.

I see no issues to have the value "PART" as valid value for two different `TargetTypeValue` values.
(It was easy to implement)

robUx4:
Technically it's only informational, but it would be wrong to have one name that means 2 different things for the user. The goal of the string is to be more meaningful that the numbers. But if it's not then it's useless.

IMO `TargetTypeValue` 40 is too loosely defined to be really useful. But it's the one that should have the `PART` name.

As for existing files I doubt `PART` is used in any existing file (not sure `TargetType` is even used much).

retokromer:
> All my files uses orderd chapters and therefore I need Tags for naming the Editions and all of these Tags uses `TargetTypeValue` and `TargetType`

Thanks for this piece of information. I still support this PR.

Issue #756, opened 2023-05-21
Fix Block text referring to a bit when there's 2 bits used


robUx4:
A table as in #755 might help clarify this as well.

PR #766, opened 2023-05-21
Rework Block/SimpleBlock definition using tables of bit

### LINKED ISSUE #755: Present Block/SimpleBlock bit layout in tables
This is how it's usually done with binary formats. It should be easier to read.

We have a variable length for the Track Number, but we only have 4 possible sizes. We can make a table for each.

robUx4:
This is much clearer regarding the variable length.

There is now a big text with explanations common to Block and SimpleBlock.

The lacing examples still need a bit of work.

Fixes #755 and #756

### DIFF
--- a/notes.md
+++ b/notes.md
@@ -98,6 +98,9 @@ The timestamp is expressed in Track Ticks; see (#timestamp-ticks).
 The value is stored as a signed value on 16 bits.
 
 ## Block Structure
+
+This section describes the binary data contained in the `Block` Element (#block-element). Bit 0 is the most significant bit.
+
 ### Block Header
 
 | Offset | Player | Description |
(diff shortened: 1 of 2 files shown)


mbunkus (approved):

notes.md
```diff
@@ -67,65 +67,159 @@ Examples:
...
 the frames referencing them. A consequence is that timestamps might not be consecutive.
 But a frame with a past timestamp **MUST** reference a frame already known, otherwise it's considered bad/void.
 
-## Block Header
+Matroska has two similar ways to store frames in a block:
+
+* in a `Block` which is contained inside a `BlockGroup`,
+* or in a `SimpleBlock` which is directly in the `Cluster`.
+
+The `SimpleBlock` is usually prefered unless some extra elements of the `BlockGroup` need to be used.
```
retokromer, line 81: prefered > preferred

retokromer (approved):
One typo, otherwise LGTM.

notes.md
robUx4: Good catch, thanks!

PR #769, opened 2023-05-29
Fix some list spacing and millisecond unit

robUx4:


robUx4:
This is due the the markdown processing. There must be a line after the ":" to do a proper list. And the space between lines is so that the spacing is humanly readable.

Check https://www.ietf.org/archive/id/draft-ietf-cellar-matroska-15.html#section-21.2-6 and https://www.ietf.org/archive/id/draft-ietf-cellar-matroska-15.html#section-26-2

mbunkus:
Like I said I'm fine with the blank lines before & after the lists, just wondering about the lines inside the lists. The generated HTML looks to be identical (one list of N entries, as it should be). So your argument is about making the Markdown variant easier to read for lists with long items, right? If so, that's fine by me, too. Just wanted to understand.

mbunkus (approved):

retokromer (approved):

robUx4:
> So your argument is about making the Markdown variant easier to read for lists with long items, right? If so, that's fine by me, too. Just wanted to understand.

They are both in the same document, not meant to be compared. Just to show how "bad" it currently is.

PR #770, opened 2023-05-29
Add a normative section about Tag level emptying

robUx4:
That's a rewritten section of a rule that exists in the Tags document.

Plus some cleaning.


tags-precedence.md
```diff
@@ -17,5 +19,12 @@ As the Tag element is optional, a lot of `Matroska Readers` do not handle it and
 So for maximum compatibility, it's usually better to put the strings in the `TrackEntry`, `ChapterAtom` and `Attachment`
 and keep the tags matching these values if tags are also used.
 
-See [@?MatroskaTags] for more details on how to use tags.
+## Tag Levels
+
+Tag elements allow tagging information on multiple levels, each level having a `TargetTypeValue` (#targettypevalue-element).
+An element for a given `TargetTypeValue` also applies to the lower levels, ie smaller `TargetTypeValue` values. If an upper value
```
retokromer, line 25: `ie` > `i.e.`

mbunkus: Agreed, and maybe even change the wording completely to something like "…also applies to lower levels denoted by smaller `TargetTypeValues` values."

tags-precedence.md
```diff
@@ -17,5 +19,12 @@ As the Tag element is optional, a lot of `Matroska Readers` do not handle it and
...
+doesn't apply to a level but the actual value to use is not known, 
+an empty `TagString` ((#tagstring-element)) or an empty `TagBinary` ((#tagbinary-element)) element **MUST** be used as the tag value for this level.
+
+See [@?MatroskaTags] for more details on common tags values.
```
mbunkus, line 29: Isn't "common tags values" wrong here? If you think about the Matroska tagging system as a listed of key/value pairs, the "value of a tag" is e.g. "Mozart", not "COMPOSER". I guess you're trying to refer to the common values of the `TagString` element here, but that's not what a lay person would understand to be the "value of a tag".

mbunkus (changes requested):

tags-precedence.md
robUx4: I reword it to be a little more precise on what to expect there.

mbunkus (approved):

retokromer (approved):

PR #779, opened 2023-05-29
Use file-like protocol for live streaming

robUx4:


streaming.md
```diff
@@ -23,8 +23,8 @@ Livestreaming is the equivalent of television broadcasting on the internet. Ther
 families of servers for livestreaming: RTP/RTSP and HTTP. Matroska is not meant to be
 used over RTP. RTP already has timing and channel mechanisms that would be wasted if doubled
 in Matroska. Additionally, having the same information at the RTP and Matroska level would
-be a source of confusion if they do not match. Livestreaming of Matroska over HTTP
-(or any other plain protocol based on TCP) is possible.
+be a source of confusion if they do not match.
+Livestreaming of Matroska over file-like protocols like HTTP, QUIC, FTP, etc is possible.
```
mbunkus, line 27: Personally I wouldn't even mention FTP here. FTP is an anachronism that isn't even widely used anymore (and really shouldn't from a security point of view). Just say "…over file-like protocols such as HTTP or QUIC is possible" and leave it at that.

mbunkus (changes requested):

streaming.md
retokromer: `etc` > `etc.`

I agree that FTP is an anachronism, but still I see it used quite often. This is just a remark and **not** an objection for deleting it here, of course.

robUx4: I added FTP as a replacement for the "TCP based" protocol. But then QUIC is over UDP. I'll just drop it.

mbunkus (approved):

retokromer (approved):

PR #780, opened 2023-05-29
Clarify the recommended limit for Clusters

robUx4:


matroska_implement.md
```diff
@@ -2,8 +2,8 @@
 
 ## Cluster
 
-It is **RECOMMENDED** that the size of each individual `Cluster Element` be limited to store
-no more than 5 seconds or 5 megabytes.
+It is **RECOMMENDED** that the size of each individual `Cluster Element` store no more than the lower of
+5 seconds or 5 megabytes.
```
mbunkus, line 6: That sounds just as wrong as the original. How about:

> It is **RECOMMENDED** that each individual `Cluster Element` contains no more than 5 seconds or 5 megabytes of content.

mbunkus (changes requested):

matroska_implement.md
robUx4: Yup, much better.

mbunkus (approved):

PR #783, opened 2023-05-29
Do not use the attachement if it can't be checked

robUx4:
This is just a recommandation for security

### DIFF
--- a/matroska_security.md
+++ b/matroska_security.md
@@ -6,7 +6,7 @@ Attacks on a `Matroska Reader` could include:
 
 * Storage of a arbitrary and potentially executable data within an `Attachment Element`.
   `Matroska Readers` that extract or use data from Matroska Attachments **SHOULD**
-  check that the data adheres to expectations.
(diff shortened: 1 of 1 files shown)


matroska_security.md
```diff
@@ -6,8 +6,8 @@ Attacks on a `Matroska Reader` could include:
 
 * Storage of a arbitrary and potentially executable data within an `Attachment Element`.
   `Matroska Readers` that extract or use data from Matroska Attachments **SHOULD**
-  check that the data adheres to expectations.
-* A `Matroska Attachment` with an inaccurate mime-type.
+  check that the data adheres to expectations or not us the attachement.
```
mbunkus, line 9: `us` → `use`

mbunkus (changes requested):

matroska_security.md
robUx4: Good catch!

mbunkus (approved):

PR #789, opened 2023-05-29
Misc grammar/typo/URL fixes

robUx4:


mbunkus (approved):

notes.md
```diff
@@ -650,16 +650,16 @@ During playback, when a frame has a negative timestamp, the content **MUST** be
 
 The default Track Tick duration is one millisecond.
 
-The `TimestampScale` is a floating value, which is usually 1.0. But when it's not, the multiplied
-Block Timestamp is a floating values in nanoseconds.
+The `TimestampScale` is a floating point value, which is usually 1.0. But when it's not, the multiplied
+Block Timestamp is a floating point value in nanoseconds.
```
retokromer, line 654: I would write `floating-point number` (with a dash and `number` rather than `value`), but I am not an English native.

robUx4: You are probably right: https://en.wikipedia.org/wiki/Floating-point_arithmetic. But I'm not an oiriginal English speaker either ;)

PR #792, opened 2023-07-02
Add a link for further BlockAdditional explanation

robUx4:
I had uploaded this change but never made a merge request


robUx4:
Actually that paragraph is in the codec spec...

PR #807, opened 2023-10-15
Remove tags Status Of This Document section

robUx4:
The IETF status doesn't matter. The document is found at the IETF anyway. We already describe the goal of the document in the Introdution.

And the EBML RFC doesn't has a status either. The Matroska document has one because it deals with versions of Matroska.


retokromer (approved):

mcr:
The RPC will put in the correct text there.

robUx4:
> The RPC will put in the correct text there.

So we can remove it or keep it as a placeholder ?

robUx4:
This was done by 3417432b06c30216f3fcbd6976b0be6c106c662d in the Matroska document. But there was already some text explaining what the document contains.

robUx4:
I messed up a push on this branch. It's merged plus other Matroska documents that don't need it.

Issue #375, opened 2020-04-29
Add Support for Closed Captions to .MKV container


sbshepherd:
I would like to transcode existing .MXF video files to FFV1/.MKV for long-term preservation but the closed caption streams get stripped out because the .MKV container can’t contain them.  

The codecs of my .MXF files differ, but the example I’ll use here is DNxHD. It contains six closed caption streams (different languages). Three are EIA-608 and three are EIA-708. These show in MediaInfo as “Text” streams, and they show in FFprobe as a single data stream with a data_type of “vbi_vanc_smpte_436M.” See screenshots below:

![image](https://user-images.githubusercontent.com/64480460/80605489-681b5700-89f0-11ea-943d-d780bae1faab.png)
![image](https://user-images.githubusercontent.com/64480460/80605529-76697300-89f0-11ea-8c9e-b1bc7e256f87.png)

If it’s helpful, here is the video codec information for this file:
<img width="466" alt="DNxHD_MXF" src="https://user-images.githubusercontent.com/64480460/80605577-81bc9e80-89f0-11ea-87d2-361aba05c696.PNG">

Ideally, these caption streams will carry over into the new .MKV file and be playable in a standard media player such as VLC. I should be able to turn on/off each language I want to view as it the video plays.

JeromeMartinez:
> and they show in FFprobe as a single data stream with a data_type of “vbi_vanc_smpte_436M.”

FYI "vbi_vanc_smpte_436M" in FFmpeg is called "Ancillary Data" in MediaInfo (608/708 captions are muxed in the "vbi_vanc_smpte_436M" which is muxed in the MXF).

> It contains six closed caption streams

For reference: 1 closed caption stream, format is CDP.
We could extract 608 from CDP, but 708 can not be alone (it needs CDP).
Here, I think we should convert CDP to "DTVCC Transport" (transport layer of 708 spec, and same features as CDP). this stream would transport 608 and 708 streams as in ATSC streams ("DTVCC Transport"  is the content in AVC or HEVC private element dedicated to captions).


Several steps here:
- defining 708 transport layer in a MKV track (extension document)
- implementing it in an encoder
- implementing it in VLC

sbshepherd:
There is a sample file at the following location if one is needed for testing: https://archive.org/details/xdcam_sample_with_caption_track

robUx4:
[...]

PR #823, opened 2024-04-10
add c608 mapping

dericed:
For discussion, here is a pull request to add storage of EIA-608 data to Matroska. It's defined in a manner similar to (QuickTime)[https://developer.apple.com/documentation/quicktime-file-format/closed_captioning_sample_data/]. The QuickTime version is a bit vague so I added something to the extra data based on DASH to define the name of the caption channel and its language.

### DIFF
--- a/block_additional_mappings/c608-captions.md
+++ b/block_additional_mappings/c608-captions.md
@@ -1,0 +1,26 @@
+## c608 Captions
+
+### c608 Captions Description
+
+CEA-608 captions can be stored in `BlockMore Element` to associate the content of a Matroska Block (typically a video frame) with closed captioning data. Storing CEA-608 caption data requires that the corresponding Block SHALL NOT use Lacing.
+
+The `BlockAdditional Element` stores the c608 data as a array of one or more octet pairs from one data channel of a CEA-608 data stream with each octet pair corresponding to a video frame. For more information about the content see [@!ANSI-CTA-608-E-S-2019].
+
+### BlockAddIDType
+
+The BlockAddIDType value reserved for CEA-608 captions is "608".
+
+### BlockAddIDName
+
+The BlockAddIDName value reserved for CEA-608 captions is "CEA-608 captions".
+
+### BlockAddIDExtraData
+
+BlockAddIDExtraData MAY contain a value as defined by Section 6.4.3.3 of [!@DASH-IF-IOP] that describes the caption services and language. An ABNF from [!@DASH-IF-IOP] for this value is restated here for convenience:
+
+```
+@value = (channel *3 [";" channel]) / (language *3[";" language])
+2 channel = channel-number "=" language
+3 channel-number = CC1 | CC2 | CC3 | CC4
+4 language = 3ALPHA ; language code per ISO 639.2/B
+```
--- a/rfc_backmatter_codec.md
+++ b/rfc_backmatter_codec.md
@@ -141,3 +141,23 @@
     <date year="2022" month="October" day="6"/>
   </front>
 </reference>
+
+<reference anchor="ANSI-CTA-608-E-S-2019" target="https://shop.cta.tech/products/line-21-data-services">
+  <front>
+    <title>Line 21 Data Services ANSI/CTA-608-E S-2019</title>
+    <author>
+      <organization>ANSI/CTA</organization>
+    </author>
+    <date year="2019" month="December"/>
+  </front>
+</reference>
+
+<reference anchor="DASH-IF-IOP" target="https://dashif.org/docs/DASH-IF-IOP-v4.3.pdf">
+  <front>
+    <title>Guidelines for Implementation: DASH-IF Interoperability Points</title>
+    <author>
+      <organization>DASH Industry Forum</organization>
+    </author>
+    <date year="2018" month="November" day="18"/>
+  </front>
+</reference>


robUx4:
I guess this is one correct/portable way of doing it. As discussed in #375 it may also be good to extract the data into proper subtitle tracks.

Also should we add constraints on what kind of track (video, with specific resolutions) it can be attached to ?

mbunkus:
I'm strongly against storing CC data in this way. Repeating what I wrote in #375:

> Personally I'd vote for "keep it subtitle, let CodecID speak for itself". We already integrate to many different formats under the type "subtitle", some text based, others are images. And closed captions fulfill largely the same role.

> I don't consider their traditional way of transportation (embedded in the video track) to be relevant for our decision.

and

> Well, what our track type "subtitles" transports can easily fill both roles, and it only depends on the content. Similar to how "audio" tracks can contain the whole dialog or only the director's comments. I don't see any reason to use a separate track type.

Block additions are meant to carry data required to or enhancing decoding the main block data in the same track. This means the video data (think of e.g. HDR information). However, CCs don't have anything to do with decoding video data at all.

There's no advantage to storing CC data alongside video data. All it does is creating new corner cases for software handling it. For example, mkvmerge would suddenly have to expose that data as some kind of virtual track outside of the regular track structure so that MKVToolNix GUI can present the CC data as what it is, continuous data timed to display at certain points = a track. Same for all plackbac programs out there.

My vote goes to storage as regular track with type `subtitle` & proper codec ID. I'd also be fine with storage as regular track with a new track type `closed_captions` or whatever.

dericed:
I'm adding a subtitle track definition for C608 for comment. See https://github.com/ietf-wg-cellar/matroska-specification/pull/823/commits/90e7b11401cae60a6a755c573082f93b0996b30f. If there's consensus about this approach, then I'll remove the BlockAdditionalMapping version and rebase.

For storing the channel number I wanted to make a recommendation and suggested the LABEL tag, but more appropriate suggestions are welcome. Or perhaps this could just go into the private data.

codec_specs.md
```diff
@@ -944,6 +944,14 @@ Codec Name: ARIB STD-B24 subtitles
 Description: This is the textual subtitle format used in the ISDB/ARIB broadcasting standard.
 For more information, see (#arib-isdb-subtitles) on ARIB (ISDB) subtitles.
 
+### S_C608
+
+Codec ID: S_C608
+
+Codec Name: CEA-608 Captions
+
+Description: Each Block of a S_C608 subtitle track stores the c608 data as a array of one or more octet pairs from one data channel of a CEA-608 data stream with each octet pair corresponding to a video frame. The "LABEL" Matroska tag is RECOMMENDED for storage the channel number of the CEA-608 (such as "CC1", "CC2", "CC3" or "CC4"). For more information about this format see [@!ANSI-CTA-608-E-S-2019].
```
JeromeMartinez, line 953: CC1 & CC2 share field 1 and can not be separated.
CC3 & CC4 share field 2 and can not be separated.

IMO it would be better to indicate the field, or even better to say that for frames there should be 4 bytes (2 bytes of field 1 then 2 bytes of field 2) so we avoid a have to manage tags.

JeromeMartinez, line 953: CC1 & CC3 share field 1 and can not be separated.
CC2 & CC4 share field 2 and can not be separated.

IMO it would be better to indicate the field, or even better to say that for frames there should be 4 bytes (2 bytes of field 1 then 2 bytes of field 2) so we avoid a have to manage tags.

JeromeMartinez: Actually we have also T1 to T4 and XDS, so the service name ("CC" thing) is really not relevant.

dericed: @JeromeMartinez I pushed an update.

JeromeMartinez:
> first channel number of the CEA-608 pair (expressed as either "CC1" or "CC2")

CC1 and CC2 are names of a service in the specs and it is technically possible to use e.g. only CC2 without using CC1 so I prefer just "field number of the CEA-608 pair (expressed as either "1" or "2")".

block_additional_mappings/c608-captions.md
```diff
@@ -0,0 +1,26 @@
...
+
+The BlockAddIDName value reserved for CEA-608 captions is "CEA-608 captions".
+
+### BlockAddIDExtraData
+
+BlockAddIDExtraData MAY contain a value as defined by Section 6.4.3.3 of [!@DASH-IF-IOP] that describes the caption services and language. An ABNF from [!@DASH-IF-IOP] for this value is restated here for convenience:
+
+```
+@value = (channel *3 [";" channel]) / (language *3[";" language])
+2 channel = channel-number "=" language
```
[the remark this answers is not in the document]
robUx4: To answer myself: yes, these number are line numbers in the PDF. They should be removed here.

block_additional_mappings/c608-captions.md
```diff
@@ -0,0 +1,26 @@
...
+
+The BlockAddIDType value reserved for CEA-608 captions is "608".
+
+### BlockAddIDName
+
+The BlockAddIDName value reserved for CEA-608 captions is "CEA-608 captions".
+
+### BlockAddIDExtraData
+
+BlockAddIDExtraData MAY contain a value as defined by Section 6.4.3.3 of [!@DASH-IF-IOP] that describes the caption services and language. An ABNF from [!@DASH-IF-IOP] for this value is restated here for convenience:
```
robUx4, line 19: MAY should be normative

```suggestion
BlockAddIDExtraData **MAY** contain a value as defined by Section 6.4.3.3 of [!@DASH-IF-IOP] that describes the caption services and language. An ABNF from [!@DASH-IF-IOP] for this value is restated here for convenience:
```

block_additional_mappings/c608-captions.md
```diff
@@ -0,0 +1,26 @@
...
+
+```
+@value = (channel *3 [";" channel]) / (language *3[";" language])
+2 channel = channel-number "=" language
+3 channel-number = CC1 | CC2 | CC3 | CC4
```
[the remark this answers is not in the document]
robUx4: Given the examples in the PDF they are litteral strings.

In a more recent version of the [split PDF](https://dashif.org/docs/IOP-Guidelines/DASH-IF-IOP-Part9-v5.0.0.pdf) it says:
> SCTE 214-1 [5], section 7.2.3, “Signalling CEA-608 caption service metadata”

In the [current version of SCTE 214-1](https://wagtail-prod-storage.s3.amazonaws.com/documents/ANSI_SCTE_214-1_2024.pdf) that's section 8.2.3 and the BNF has been fixed:

```
@value = language / channel 0*3 [";" channel]
channel = channel-number "=" language
channel-number = “CC1” / “CC2” / “CC3” / “CC4”
language = 3ALPHA ; language code per ISO 639-3
```

We should probably use that and also reference that source directly.

robUx4 (changes requested):

robUx4:
marking this a v5 as the codec is not supported yet (or defined).

codec_specs.md
```diff
@@ -944,6 +944,14 @@ Codec Name: ARIB STD-B24 subtitles
 Description: This is the textual subtitle format used in the ISDB/ARIB broadcasting standard.
 For more information, see (#arib-isdb-subtitles) on ARIB (ISDB) subtitles.
 
+### S_C608
+
+Codec ID: S_C608
+
+Codec Name: CEA-608 Captions
+
+Description: Each Block of a S_C608 subtitle track stores the c608 data as a array of one or more octet pairs from one data channel of a CEA-608 data stream with each octet pair corresponding to a video frame. The "LABEL" Matroska tag is RECOMMENDED for storage the first channel number of the CEA-608 pair (expressed as either "CC1" or "CC2"). For more information about this format see [@!ANSI-CTA-608-E-S-2019].
```
robUx4, line 953: RECOMMENDED should have "**" around it

block_additional_mappings/c608-captions.md
```diff
@@ -0,0 +1,26 @@
+## c608 Captions
+
+### c608 Captions Description
+
+CEA-608 captions can be stored in `BlockMore Element` to associate the content of a Matroska Block (typically a video frame) with closed captioning data. Storing CEA-608 caption data requires that the corresponding Block SHALL NOT use Lacing.
```
robUx4, line 5: SHALL NOT should have "**" around it.

PR #824, opened 2024-04-21
Apply some RFC9559 AUTH48 changes

robUx4:
Both formatting and text rules.

No actual text changes yet.


robUx4:
Some of the changes also apply to the other documents as well.

PR #826, opened 2024-04-29
Add remaining changes from AUTH48 review

robUx4:
These includes formatting, (US) English errors, use of sections in references, etc.

This more or less matches the XML file sent with the AUTH48 review. There is some kind of formatting that doesn't seem possible to do with Markdown/mmark (usually in sub lists). mmark also adds a lot of <t></t> where there is not in AUTH48.

This doesn't includes questions raised by the editor(s).

Some changes also impact the other Matroska CELLAR documents. We should regenerate them soon.

### DIFF
--- a/codec_specs.md
+++ b/codec_specs.md
@@ -7,7 +7,7 @@ Each TrackEntry used within Matroska **MUST** reference a defined `Codec Mapping
 `Codec ID` to identify and describe the format of the encoded data in its associated Clusters.
 This `Codec ID` is a unique registered identifier that represents the encoding stored within
 the Track. Certain encodings **MAY** also require some form of codec initialization
-in order to provide its decoder with context and technical metadata.
+to provide its decoder with context and technical metadata.
 
 The intention behind this list is not to list all existing audio and video codecs,
 but rather to list those codecs that are currently supported in Matroska and therefore
--- a/iana.md
+++ b/iana.md
@@ -6,7 +6,7 @@
 IANA has created a new registry called the "Matroska Chapter Codec IDs" registry.
 The values correspond to the unsigned integer `ChapProcessCodecID` value described in (#chapprocesscodecid-element).
 
-The following are needed to register a new Chapter Codec ID in this registry:
+To register a new Chapter Codec ID in this registry, one needs
 a Chapter Codec ID,
 a Change Controller (IETF or email of registrant), and
 an optional Reference to a document describing the Chapter Codec ID.
--- a/iana.md
+++ b/iana.md
@@ -13,7 +13,7 @@ an optional Reference to a document describing the Chapter Codec ID.
 
 The Chapter Codec IDs are to be allocated according to the "First Come First Served" policy [@!RFC8126].
 
-`ChapProcessCodecID` values of "0" and "1" are RESERVED to the IETF for future use.
+Values of "0" and "1" are reserved for future use (with the IETF as the Change Controller).
 
 ## Media Types
 
@@ -23,7 +23,7 @@ Historically, Matroska files and streams have used the following media types wit
 For better compatibility, a system **SHOULD** be able to handle both formats.
 Newer systems **SHOULD NOT** use the historic format and use the format that follows the format in [@!RFC6838] instead.
 
-Please register three media types, the [@!RFC6838] templates are below:
+IANA has registered three media types per the templates (see [@!RFC6838]) in the following subsections.
 
 ### For Files Containing Video Tracks
 
--- a/iana.md
+++ b/iana.md
@@ -21,7 +21,7 @@ Matroska files and streams are found in three main forms: audio-video files, aud
 
 Historically, Matroska files and streams have used the following media types with an "x-" prefix.
 For better compatibility, a system **SHOULD** be able to handle both formats.
-Newer systems **SHOULD NOT** use the historic format and use the format that follows the [@!RFC6838] format instead.
+Newer systems **SHOULD NOT** use the historic format and use the format that follows the format in [@!RFC6838] instead.
 
 Please register three media types, the [@!RFC6838] templates are below:
 
--- a/iana.md
+++ b/iana.md
@@ -19,7 +19,7 @@ The Chapter Codec IDs are to be allocated according to the "First Come First Ser
 
 Matroska files and streams are found in three main forms: audio-video files, audio-only, and occasionally with stereoscopic video tracks.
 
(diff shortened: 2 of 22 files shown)


robUx4:
@mbunkus @dericed if there's no objection I'll merge it this week-end. Then we can publish the other documents with the relevant changes (although we can't reference just RFC9559 yet).

mbunkus:
Thanks for all your work, Steve, and sorry for not reacting sooner. Pretty busy. I do have reviewing thir PR on my TODO list for today or tomorrow. So please don't merge before Sunday. Thanks.

robUx4:
No problem, there's no rush. Also we still have some reviewing steps before it's all finalized, at least the last call to make sure we agree on the final document. But I expect more steps to come, especially with the inconsistent backticking we use.

JeromeMartinez:
@robUx4 @mbunkus Thank you both!

index_matroska.md
```diff
@@ -39,22 +39,22 @@ fullname="Dave Rice"
...
 # Introduction
 
-Matroska is an audiovisual data container format. It was derived from a project called [@?MCF],
+Matroska is an audiovisual data container format. It was derived from a project called [@?MCF]
 but diverges from it significantly because it is based on EBML (Extensible Binary Meta Language) [@!RFC8794],
 a binary derivative of XML. EBML provides significant advantages in terms of future format extensibility,
 without breaking file support in parsers reading the previous versions.
 
-First, it is essential to clarify exactly "What an Audio/Video container is", to avoid any misunderstandings:
+To avoid any misunderstandings, it is essential to clarify exactly what an audio/video container is:
```
mbunkus, line 55: I know this has been in there forever, but maybe use "…exactly what an audiovisual data container is" in order to be consistent with the wording in the previous paragraph.

notes.md
```diff
@@ -135,45 +135,52 @@ Figure: Block Header with 1 octet TrackNumber
...
+Figure: Block Header with 2-Octet TrackNumber
 
 where:
 
-{newline="true" spacing="normal"}
-Track Number: 8, 16, 24, 32, 40, 48 or 64 bits
-: an EBML VINT coded track number
+{newline="false" spacing="normal"}
+Track Number:
+: 8, 16, 24, 32, 40, 48, or 64 bits. An EBML VINT-coded track number.
```
mbunkus, line 144: I'm slightly baffled why 56 bits is not included in the list 😂 Or rather, if it's a VINT, why 64 bits is there but 56 isn't — given as an 64 bit VINT sacrifices 8 bits to the length marker, leaving 56 bits for the value. Same applies to other places below.

notes.md
```diff
@@ -197,37 +204,44 @@ Figure: SimpleBlock Header with 1 octet TrackNumber
...
+Figure: SimpleBlock Header with 2-Octet TrackNumber
 
 where:
 
-{newline="true" spacing="normal"}
-Track Number: 8, 16, 24, 32, 40, 48 or 64 bits
-: an EBML VINT coded track number
+{newline="false" spacing="normal"}
+Track Number:
+: 8, 16, 24, 32, 40, 48, or 64 bits. An EBML VINT-coded track number.
```
mbunkus, line 213: Same objection as above

mbunkus (changes requested):

index_matroska.md
robUx4: This was one of the RFC editor questions about which one we should use for consistency. I think audiovisual is more generic (includes subtitles). "audio/video" should be kept only when those apply.

I did not make such editorial changes yet, as they may have their own opinion on the subject.

notes.md
robUx4: I plead guilty 88df108e245ccaa0024b5fe3dc66bfc1fc053d5d

chapter_codecs.md
```diff
@@ -39,15 +39,15 @@ and so on until there is no `Nested Chapter` with the same `ChapterTimeStart`.
...
+chapter **MUST** be executed, then the leave commands of its parent chapter, etc., until the
 common `Parent Chapter` or `Edition` element. The leave command of that `Parent Chapter` or `Edition` element
 **MUST NOT** be executed.
 * the enter commands (`ChapProcessTime`=1) of the `Nested Chapter` of the common `Parent Chapter` or `Edition` element,
 to reach the chapter we switch to, **MUST** be executed, then the enter commands of its `Nested Chapter`
 to reach the chapter we switch to **MUST** be executed, until that chapter is the chapter we switch to.
 The enter commands of that chapter **MUST** be executed as well.
 
-When the last Chapter finished playing -- i.e. its `ChapterTimeEnd` has been reached --
+When the last Chapter finished playing, i.e., its `ChapterTimeEnd` has been reached --
```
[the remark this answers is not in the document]
robUx4: Fixed. This is in the codec document, so no impact on RFC9559 (hence it was incorrectly edited).

codec_specs.md
```diff
@@ -841,7 +841,7 @@ Codec ID: S_TEXT/UTF8
 
 Codec Name: UTF-8 Plain Text
 
-Description: Basic text subtitles. For more information, see (#subtitles) on Subtitles.
+Description: Basic text subtitles. For more information; see (#subtitles) on Subtitles.
```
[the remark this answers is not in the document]
robUx4: Also fixed in #827.

iana.md
```diff
@@ -3,26 +3,27 @@
...
 
-Matroska files and streams are found in three main forms: audio-video files, audio-only and occasionally with stereoscopic video tracks.
+Matroska files and streams are found in three main forms: audio-video files, audio-only, and occasionally with stereoscopic video tracks.
 
-Historically Matroska files and streams have used the following media types with a "x-" prefix.
-For better compatibility a system **SHOULD** be able to handle both formats.
-Newer systems **SHOULD NOT** use the historic format and use the format that follows the [@!RFC6838] format instead.
+Historically, Matroska files and streams have used the following media types with an "x-" prefix.
+For better compatibility, a system **SHOULD** be able to handle both formats.
+Newer systems **SHOULD NOT** use the historic format and use the format that follows the format in [@!RFC6838] instead.
```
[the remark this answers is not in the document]
robUx4: Fixed in #827 as this MR is meant to match the answer to the first AUTH48 round (but also to check nothing wrong has been introduced).

notes.md
robUx4: Fixed in #827

index_matroska.md
robUx4: Fixed in #827

PR #827, opened 2024-05-05
extra AUTH48 editorial changes

robUx4:
On top of #826.

These are changes that need to be pushed in the next AUTH48 cycle. Keeping as draft for now as we may add more changes before the new AUTH48 reply.


robUx4:
There's also the https://github.com/ietf-wg-cellar/matroska-specification/tree/9559rebased branch which contains all the changes, the original AUTH48 XML, the changes we made to it and the XML reply file. Ultimately generating rfc9559.xml should match the XML from the RFC editors. But mmark is a bit limited in the formatting so some extra manual changes are needed.

robUx4:
These changes have been integrated in the v2 of the AUTH48 review.

PR #832, opened 2024-08-04
[AUTH48] add more element values IANA registries

robUx4:
Many elements may have new value in the future. We need to define how to expand these in the future.

The generation of the IANA text for each element(s) is not automated.

~~Draft on top of #837~~


JeromeMartinez (commented):
Didn't check the XSL but other parts seem good.

robUx4:
Gonna merge this in 2 steps:
- draft 22 is the current AUTH48 state.
- draft 23 will be AUTH48 plus the new IANA registries.

PR #833, opened 2024-08-21
cleanup Chapter Codec document

robUx4:
Some clarifications are also added about the uniqueness of the interpreter for the whole duration of an Edition. That allows internal variables to be shared between chapters. This is already the case in VLC for the DVD Menu (which wouldn't work otherwise).

The control track document is also updated the same way with regards to formatting.


robUx4:
Merging so we can update the expired documents with cleaner versions.

PR #835, opened 2024-08-21
Add MatroskaJS Chapter Codec

robUx4:
Rather than reinventing the wheel with a custom language, we can reuse and well-known one.

The goal of Matroska Script was to be similar to JavaScript/ECMAStrip/ActionScript.
But, in JS, integers can only use 56 bits. So not all ChapterUID (64-bit unsigned integer) can be addressed
with the current Matroska Script stable API.

Rather than adding a different GotoAndPlay that is compatible with JavaScript (using a string), we start from
scratch with ECMAScript 6th Edition (ES6).

The implementation of the Choice handling is almost complete in VLC by a Google Summer of Code student.

Draft on top of #833


robUx4:
Not sure it should go in the "historical" Chapter Codec document, ie make the specification for what existed so far. Or if it should be a document on its own. Maybe each command will need its own IANA registry.

chapter_codecs.md
```diff
@@ -72,6 +72,84 @@ The one and only command existing for the moment is `GotoAndPlay( ChapterUID );`
 same suggests, it means that, when this command is encountered, the `Matroska Player`
 **SHOULD** jump to the `Chapter` specified by the UID and play it, as long as this `Chapter` exists.
 
+## MatroskaJS
+
+This is the case when `ChapProcessCodecID` = 2. This is a script language build for
+Matroska purposes. It uses the [@!ECMAScript] 6th Edition (ES6) syntax. The commands are stored as text commands, in UTF-8.
```
Labnann, line 78: We  may want to use ES5 instead.

robUx4: That would be better (less constraint) but we don't have a working solution with that.

Also there are [things](https://compat-table.github.io/compat-table/es6/) that could make the JS usage cleaner like [let and const](https://hacks.mozilla.org/2015/07/es6-in-depth-let-and-const/) or static arrays and [typed arrays](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray). For example you could have a global function that gets the list of choices in an array and call the MatroskaJS engine. Then each (enter) chapter would just had the array and that call.

chapter_codecs.md
```diff
@@ -72,6 +72,84 @@ The one and only command existing for the moment is `GotoAndPlay( ChapterUID );`
 same suggests, it means that, when this command is encountered, the `Matroska Player`
 **SHOULD** jump to the `Chapter` specified by the UID and play it, as long as this `Chapter` exists.
 
+## MatroskaJS
+
+This is the case when `ChapProcessCodecID` = 2. This is a script language build for
```
rcombs, line 77: probably meant "built"?

robUx4: Good catch. There was the same issue for MatroskaScript as well.

PR #842, opened 2024-09-08
fix typo

retokromer:
(this one is … weird ;)


robUx4:
Thanks for checking this. In light of all the issues you found I went though the RFC9559 to make sure we don't have any other bugs before publishing. It seems OK now.

PR #846, opened 2024-10-05
[AUTH48] apply ultimate RFC Editor changes

robUx4:
Only the first editorial change is a real text change. The rest is just additions in IANA registry.


robUx4:
Merging as the latest version have been approved by the authors.

PR #854, opened 2024-10-22
add Matroska v5 document

robUx4:
There's a lot of work to be done, even the introduction is incorrect. But it contains the basic Matroska v5 that have already been merged in ebml_matroska.xml.

I kept the authors as in the other documents.


robUx4:
Merging this so we can keep adding to as we go along.

I except to have other things than just new elements. Look at [all the things we have pending for v5](https://github.com/ietf-wg-cellar/matroska-specification/issues?q=is%3Aissue+is%3Aopen+label%3Amatroska-v5).

PR #856, opened 2024-10-22
define the TrackType value for each codec types

robUx4:
And the value is mandatory. There is no default value.


robUx4:
Later we can probably define how some common values (display width, sampling frequency, etc) should be mapped from the original codecs so we don't have to say it for each codec.

PR #857, opened 2024-10-22
add V_AV1 mapping

robUx4:
With most of the mapping details from av1.md.

### DIFF
--- a/codec_specs.md
+++ b/codec_specs.md
@@ -136,6 +136,29 @@ These recommendations are based on [@!RFC6648, section 3].
 
 ## Video Codec Mappings
 
+### V_AV1
+
+Codec ID: `V_AV1`
+
+Codec Name: Alliance for Open Media AV1 Video codec
+
+Description: Only one `Sequence Header OBU`, as defined in section 6.4 of [@!AV1], is supported per Matroska Segment.
+Each `Block` contains one `Temporal Unit` containing one or more OBUs. Each OBU stored in the Block **MUST** contain its header and its payload.
+The OBUs in the `Block` follow the `Low Overhead Bitstream Format syntax`.
+They **MUST** have the `obu_has_size_field` set to 1 except for the last OBU in the frame, for which `obu_has_size_field` **MAY** be set to 0, in which case it is assumed to fill the remainder of the frame.
+A `SimpleBlock` **MUST NOT** be marked as a Keyframe if it doesn't contain a `Frame OBU`.
+A `SimpleBlock` **MUST NOT** be marked as a Keyframe if the first `Frame OBU` doesn't have a `frame_type` of `KEY_FRAME`.
+A `SimpleBlock` **MUST NOT** be marked as a Keyframe if it doesn't contains a `Sequence Header OBU`.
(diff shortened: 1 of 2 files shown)


codec_specs.md
```diff
@@ -136,6 +136,26 @@ These recommendations are based on [@!RFC6648, section 3].
...
+Description: Only one `Sequence Header OBU`, as defined in section 6.4 of [@!AV1], is supported per Matroska Segment.
+Each `Block` contains one `Temporal Unit` containing one or more OBUs. Each OBU stored in the Block **MUST** contain its header and its payload.
+The OBUs in the `Block` follow the `Low Overhead Bitstream Format syntax`.
+They **MUST** have the `obu_has_size_field` set to 1 except for the last OBU in the frame, for which `obu_has_size_field` **MAY** be set to 0, in which case it is assumed to fill the remainder of the frame.
+A `Block` inside a `BlockGroup` **MUST** use `ReferenceBlock` elements if the first `Frame OBU` in the `Block` has a `frame_type` other than `KEY_FRAME`.
+A `Block` inside a `BlockGroup` **MUST** use `ReferenceBlock` elements if the `Block` doesn't contain a `Sequence Header OBU`.
+A `Block` with `frame_header_obu` where the `frame_type` is `INTRA_ONLY_FRAME` **MUST** use a `ReferenceBlock` with a value of 0 to reference itself.
+
+Initialization: The `CodecPrivate` consists of the first 4 octets of the `AV1CodecConfigurationRecord` described in section 2.3 of [@!AV1-ISOBMFF]. The values in this bit field come from the `Sequence Header OBU`.
```
JeromeMartinez, line 153: What is the reason not to do as described in https://aomediacodec.github.io/av1-isobmff/#av1codecconfigurationbox-syntax with all bits rather than the first 32 ones only?

robUx4: The configOBUs in there are basically the Sequence Header OBU (must be first) and the metadata OBUs. Both contain values that should be mapped to Matroska fields/metadata. If we keep it we may have incoherences between the two.

But looking at the dav1d decoder in VLC we do this:
```c
    av1_OBU_sequence_header_t *sequence_hdr = NULL;
    if (dec->fmt_in->i_extra > 4)
    {
        // in ISOBMFF/WebM/Matroska the first 4 bytes are from the AV1CodecConfigurationBox
        // and then one or more OBU
        const uint8_t *obu_start = ((const uint8_t*) dec->fmt_in->p_extra) + 4;
        int obu_size = dec->fmt_in->i_extra - 4;
        if (AV1_OBUIsValid(obu_start, obu_size) && AV1_OBUGetType(obu_start) == AV1_OBU_SEQUENCE_HEADER)
            sequence_hdr = AV1_OBU_parse_sequence_header(obu_start, obu_size);
    }
```

ie we skip the first 4 octets and then look for the Sequence Header OBU. It is in fact crucial to be there, otherwise we can't detect hardware decoding early.

So I'll fix that.

JeromeMartinez (changes requested):

mbunkus (approved):
HDR (Dolby Vision) handling should also be specced (via BlockAdditionMappings). As in, not right now in this PR, but in general.

retokromer (approved):

PR #861, opened 2024-10-23
Rework main Tag example

robUx4:
It was not clear what it meant and how it translated to actual tags in the file.

The example was using unofficial tags (for legitimate use case) so `DATE_STARTED` and `DATE_ENDED` were added.

Plus some formatting cleanup.

### DIFF
--- a/matroska_tags.xml
+++ b/matroska_tags.xml
@@ -60,7 +60,7 @@ or for video's audio track you might use "English 5.1" This is akin to the "TIT2
       <description lang="en">URL corresponding to the tag it's included in.</description>
     </tag>
     <tag name="SORT_WITH" class="Nested Information" type="UTF-8">
-      <description lang="en">A child element to indicate what alternative value the parent tag can have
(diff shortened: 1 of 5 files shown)


matroska_tags.xml
```diff
@@ -251,6 +251,16 @@ an age in other countries or a URI defining a logo).</description>
     <tag name="DATE_PURCHASED" class="Temporal Information" type="UTF-8">
       <description lang="en">Information on when the file was purchased; see also (#commercial) on purchase tags.</description>
     </tag>
+    <tag name="DATE_STARTED" class="Temporal Information" type="UTF-8">
+      <description lang="en">When the information of the parent tag starts being valid.
+      The information of the parent tag is only valid between this date and the `DATE_ENDED` date of the same level.
+      The `DATE_ENDED` is **OPTIONAL**, in which case the end date is unknown.</description>
```
retokromer, line 257: I am not a native English speaker, but is this sentence clear? `In which case` sounds strange to me here.

(Also below, at line 262.)

robUx4: Good catch! I also fixed some bogus wording.

Issue #831, opened 2024-05-27
Tags: specify tags for ITU-R BS.1770 programme loudness (LUFS) and true-peak level (dBTP)


bb010g:
The _Matroska Media Container Tag Specifications_ should specify tags for _programme loudness_ and _programme true-peak level_ as specified by [Rec. ITU-R BS.1770] ([PDF][Rec. ITU-R BS.1770 (PDF)]) and as used by [EBU R 128] ([PDF][EBU R 128 (PDF)]). The programme loudness tag's value should be recorded in LUFS (called LKFS by the ITU), and the programme true-peak level tag's value should be recorded in dBFS. I propose the tag names **ITU_BS_1770_LUFS** and **ITU_BS_1770_DBFS** for these tags. These tags could be found at all TargetType levels (track, album, etc).

Currently, Tags specifies two tags for loudness information, REPLAYGAIN_GAIN and REPLAYGAIN_PEAK. These tags are based on the ReplayGain specification. However, ITU BS.1770 and EBU R 128 are increasingly being used in production & playback of both audio & video, so being able to encode that metadata in Matroska containers without a custom tag would be nice.

Note that loudness in LUFS is chosen instead of a gain in DB or a relative loudness in LU, relative to some target loudness, because of how target loudness levels (in LUFS) can be arbitrarily chosen. For example, EBU R 128 targets -23.0 LUFS, Qobuz targets -18 LUFS, Apple Music targets -16 LUFS, and YouTube targets -14 LUFS. REPLAYGAIN_GAIN is specified to always target 89dB SPL, but we have an opportunity to be less opinionated with ITU_BS_1770_LUFS. By being clear about this not being a relative loudness, we can hopefully avoid people changing tag values when their target changes. (Audio players can support user-specified target values for ReplayGain that aren't 89dB SPL, but specifying objective LUFS makes it more clear to implementers that target loudness should be controllable by the user.) Similarly, ITU_BS_1770_DBFS is an objective measurement that can be easily used to meet an arbitrary dBFS target by an audio player.

[...]

PR #864, opened 2024-10-24
Store the ReplayGain tag values as strings

robUx4:
This is how it's stored in Vorbis [1][2] and how
mkvmerge passes the value from one to the other [3]. There doesn't seem to be any player reading it from Matroska.

[1] https://code.videolan.org/videolan/vlc/-/blob/f1a90394435471e75fee6f13383079a3d0272ec4/modules/demux/xiph_metadata.c#L472
[2] https://github.com/FFmpeg/FFmpeg/blob/153a6dc8faafc4de263a493484ffc1dc2b5b26b2/libavformat/replaygain.c
[3] https://gitlab.com/mbunkus/mkvtoolnix/-/blob/7a32d31e7029b1ca4f323eb0f7fb975ec979b089/src/common/tags/vorbis.cpp#L78

I would have preferred a binary version of a float value to avoid parsing errors. But for backward compatibility it's probably better to keep it like this. Unless mkvmerge ends up turning Vorbis replay gain into a binary format ? I don't know the code enough to tell for sure, cc @mbunkus.

### DIFF
--- a/matroska_tags.xml
+++ b/matroska_tags.xml
@@ -308,12 +308,15 @@ The majority of the contemporary rock and pop music you hear on the radio these 
       <description lang="en">It is saved as a frequency in hertz to allow near-perfect tuning of instruments to
 the same tone as the musical piece (e.g., "441.34" in Hertz). The default value is 440.0 Hz.</description>
     </tag>
-    <tag name="REPLAYGAIN_GAIN" class="Technical Information" type="binary">
-      <description lang="en">The gain to apply to reach 89dB SPL on playback. This is based on the [@!ReplayGain] standard.
-Note that ReplayGain information can be found at all TargetType levels (track, album, etc).</description>
+    <tag name="REPLAYGAIN_GAIN" class="Technical Information" type="UTF-8">
+      <description lang="en">The gain to apply to reach 89dB SPL on playback. The value is computed according to the [@!ReplayGain] standard.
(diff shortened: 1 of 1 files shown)


robUx4:
That leaves `MCDI` as the only (official) binary tag. It may change with #831 but we may just use strings as for ReplayGain.

robUx4:
I tried muxing a [.ogg file](https://streams.videolan.org/streams/metadata/replaygain/The%20B-52%27s%20-%20Give%20Me%20Back%20My%20Man.ogg) with `REPLAYGAIN_GAIN` and `REPLAYGAIN_PEAK` and mkvmerge did turn them into string (or kept them as-is).

image

robUx4:
But the "dB" unit is there in the gain and the peak level is not in dB.

robUx4:
The original files has those strings:
```
replaygain_album_true_peak=1.117522
replaygain_track_true_peak=1.117522
replaygain_album_gain=-7.31 dB
replaygain_album_peak=1.072099
replaygain_track_gain=-7.31 dB
replaygain_track_peak=1.072099
```

robUx4:
BTW, given it influences the playback (it should, in an ideal world) I think it could be elements of the `TrackEntry\Audio`. We would need to put the "album" and "track" values. But if the file is a whole album with tracks split as chapters, we can't define individual tracks from there...

robUx4:
Merging this as this matches what exists in the wild.

PR #865, opened 2024-10-24
add EBU R128 loudness/peak value tags

robUx4:
It seems for the ITU_BS_1770_LUFS we can just tell to normalize to -23, and for other sources the values should be shifted ? cc @bb010g

### DIFF
--- a/cellar-tags/rfc_backmatter_tags.md
+++ b/cellar-tags/rfc_backmatter_tags.md
@@ -10,6 +10,30 @@
   </front>
 </reference>
 
+<reference anchor="EBU-R.128" target="https://tech.ebu.ch/publications/r128/">
+  <front>
+    <title>LOUDNESS NORMALISATION AND PERMITTED MAXIMUM LEVEL OF AUDIO SIGNALS</title>
+    <author/>
+     <date month="November" year="2023" />
+  </front>
+</reference>
+
+<reference anchor="EBU-TECH.3341" target="https://tech.ebu.ch/publications/tech3341">
+  <front>
+    <title>LOUDNESS METERING: ‘EBU MODE’ METERING TO SUPPLEMENT EBU R 128 LOUDNESS NORMALIZATION</title>
+    <author/>
+     <date month="November" year="2023" />
+  </front>
+</reference>
+
+<reference anchor="EBU-TECH.3342" target="https://tech.ebu.ch/publications/tech3342">
+  <front>
+    <title>LOUDNESS RANGE: A MEASURE TO SUPPLEMENT EBU R 128 LOUDNESS NORMALIZATION</title>
+    <author/>
+     <date month="November" year="2023" />
+  </front>
+</reference>
+
 <reference anchor="GS1" target="https://www.gs1.org/standards/barcodes-epcrfid-id-keys/gs1-general-specifications">
   <front>
     <title>GS1 General Specifications</title>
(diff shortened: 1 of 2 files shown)


bb010g:
> It seems for the ITU_BS_1770_LUFS we can just tell to normalize to -23, and for other sources the values should be shifted ? cc @bb010g

I would explicitly recommend against normalization to -23. If that means using a fixed-point encoding over a floating-point encoding to avoid the drop in accuracy when further from 0, then I think that's a better alternative. If a `ITU_BS_1770_LU` is specified in loudness units relative to -23 LUFS instead, then I'm sure that people will start encoding relative to their chosen LUFS instead of properly configuring their players to normalize to a target LUFS, like what's happened with ReplayGain tags. Using LUFS won't _ensure_ that non-compliant metadata will never happen, but it will hopefully reduce the number of cases because people will realize that they're shifting an absolute value.

robUx4:
> I would explicitly recommend against normalization to -23 LUFS. If that means using a fixed-point encoding over a floating-point encoding to avoid the drop in accuracy when further from 0, then I think that's a better alternative. 

No it's a floating point number (binary) value.

> Using LUFS won't ensure that non-compliant metadata will never happen, but it will hopefully reduce the number of cases because people will realize that they're shifting an absolute value.

But if the reference is floating that means we have to also store the reference so we can match different loudness levels. In the end the source value translated means the same thing as -6 / -29 stored separately. If the playback is targetting a general loudness of -18 it's the same amplication (or compression?) to apply regardless of the original loudness. Or am I missing something ?

robUx4:
In the end I prefer to keep the strict (relative) reference to EBU R128 rules because it seems the other ones may not use the exact measuring methods. It's better to keep the information that the LUFS value actually corresponds to the R128 recommendations, rather than hoping the iTunes or Qobuz values do too. If these sources don't have R128 values they may be shifted, but it's up to the person/system doing the conversion to R128 to assume the values are really equivalent.

From [Wikipedia](https://en.wikipedia.org/wiki/EBU_R_128#Adoption_and_aftermath):
> Through the 2010s, Internet streaming services have implemented loudness-based normalisation, even though each platform uses different methods and target levels

Rather than doing a guessed shift, a proper system would actually do the measurement again on the decoded signal to generate the values (with [libebur128](https://github.com/jiixyj/libebur128) for example) and write proper standardized values.

Also as for the values we need, there's the BWF version ([EBU Tech 3285](https://tech.ebu.ch/docs/tech/tech3285.pdf)) of the metadata that can be a good source:
* LoudnessValue: A 16-bit signed integer, equal to round(100x the Integrated Loudness
Value of the file in LUFS).
* LoudnessRange: A 16-bit signed integer, equal to round(100x the Loudness Range of the
file in LU).
* MaxTruePeakLevel: A 16-bit signed integer, equal to round(100x the Maximum True Peak
Value of the file in dBTP).
* MaxMomentaryLoudness: A 16-bit signed integer, equal to round(100x the highest value of the Momentary Loudness Level of the file in LUFS).
* MaxShortTermLoudness: A 16-bit signed integer, equal to round(100x the highest value of the
Short-term Loudness Level of the file in LUFS)

robUx4:
Both 32-bit and 64-bit IEEE 754 floating numbers are now allowed, just a in EBML. The size of the tag data tells which one it is.

PR #872, opened 2024-10-25
add VP9 codec private

robUx4:


codec_specs.md
```diff
@@ -439,7 +439,8 @@ Description: VP9 is an open and royalty free video compression format developed
 
 Codec BlockAdditions: A single-channel encoding of an alpha channel **MAY** be stored in `BlockAdditions`. The `BlockAddId` of the `BlockMore` containing these data **MUST** be 1.
 
-Initialization: none
+Initialization: An **OPTIONAL** `CodecPrivate` **SHOULD** be used. The data correspond to the "VP9 Codec Feature Metadata" defined in [@!WebMContainer].
```
JeromeMartinez, line 442:
```suggestion
Initialization: The `CodecPrivate` **MAY** contain the "VP9 Codec Feature Metadata" defined in [@!WebMContainer].
```

No need to use OPTIONAL here, especially with SHOULD. MAY or SHOULD, IMO MAY as it is not mandatory and it is maybe not our role to say if it recommended (or say why).

JeromeMartinez (changes requested):

codec_specs.md
robUx4: **MAY** contain is ambiguous. It means it may contain other things. But if it's there, it's exactly what should be in there.

The CodecPrivate is optional (I have plenty of VP9 files without it, that can't be decoded with hardware decoders because of that).

But it SHOULD (can't be MUST because it's optional) be used because it makes hardware decoding possible.

mbunkus: So we have the choice of either declaring the hardware devices to be out of spec if they don't work without CodecPrivate, or the muxers that don't write CodecPrivate. Either way is not so nice. Personally I'd prefer to say that the muxers are out of spec as that side is easier to change. Therefore I'd like to use **MUST** here.

robUx4: If it's a MUST we should at least tell that legacy muxers didn't put it so people don't consider the track invalid if it's not there. Software decoders will work just fine if it's not there. And for hardware decoding it may still possible to generate it from the first frame (I think).

Also, looking at https://www.webmproject.org/vp9/mp4/, it seems possible to use a `VPCodecConfigurationRecord` (via a `VPCodecConfigurationBox`) with VP8 as well. So technically we could have the same `CodecPrivate` for VP8 as well (with a fixed value). But there's no WebM or Matroska file with that `CodecPrivate` for VP8.

And in fact in the [WebM spec](https://www.webmproject.org/docs/container/) the rule is actually:
* The Codec ID SHOULD be "V_VP8" or "V_VP9".
* There is no CodecPrivate data for VP8. For VP9, CodecPrivate SHOULD contain a list of specific VP9 codec features (Level, Profile, Bit Depth and Chroma Subsampling values) as described in[ VP9 Codec Feature Format ](https://www.webmproject.org/docs/container/#vp9-codec-feature-metadata-codecprivate).

So IMO it should remain a **SHOULD** in our case. We could actually use the same wording.

robUx4: That means I could remove the OPTIONAL part. But I can already tell the RFC editors will want to know when it should not be set.

robUx4: > And for hardware decoding it may still possible to generate it from the first frame (I think).

It seems that mkvtoolnix can generate the `CodecPrivate` from the frame header: https://gitlab.com/mbunkus/mkvtoolnix/-/blob/7a32d31e7029b1ca4f323eb0f7fb975ec979b089/src/output/p_vpx.cpp#L43 + https://gitlab.com/mbunkus/mkvtoolnix/-/blob/7a32d31e7029b1ca4f323eb0f7fb975ec979b089/src/common/vp9.cpp

So it is doable. In other words in VLC I could generate it as well if it's missing to benefit from hardware decoding. In other words making it mandatory is possible. As long as we mention it's not always the case with legacy files.

mbunkus: Yes, that's exactly what mkvmerge does. Luckily this is pretty easy for VP9. For VP8 it simply doesn't do that, though if it's the same header with the same kind of data from the first frame, it should be as possible.

I don't really have a strong preference. I just think that no matter which way we chose, one part will be out of spec, and that should likely be the software side. I'd perfectly fine with stating that legacy files MAY come without it for histrorical reasons & that decoders SHOULD be able to handle that by deriving the data from the first frame. Or something similar.

robUx4: > decoders SHOULD attempt to decode if not present by looking at first frame in order not to make legacy files violate the specs?

Yes, ideally it should be done. It's not done in VLC because that would become a bit ugly but I'll have a look.

codec_specs.md
```diff
@@ -439,7 +439,9 @@ Description: VP9 is an open and royalty free video compression format developed
 
 Codec BlockAdditions: A single-channel encoding of an alpha channel **MAY** be stored in `BlockAdditions`. The `BlockAddId` of the `BlockMore` containing these data **MUST** be 1.
 
-Initialization: none
+Initialization: The `CodecPrivate` **MUST** contain a list of specific VP9 codec features as described in the "VP9 Codec Feature Metadata" section of [@!WebMContainer].
+Legacy muxers **MAY** not have provided the `CodecPrivate`. In that case it **SHOULD** be reconstructed by the decoder from the first frame.
```
JeromeMartinez, line 443:
```suggestion
In the case `CodecPrivate` is not present, the decoder **SHOULD** reconstruct it from the first frame.
```

MAY should not be used there as it is something a muxer conforming to spec usually may do, so a bit of rewording, similar to what was written in FFV1 for similar issues.

JeromeMartinez (changes requested):

codec_specs.md
JeromeMartinez:
```suggestion
Due to the existence of some old non compliant muxers which were not providing `CodecPrivate` , the decoder **SHOULD** reconstruct it from the first frame in the case `CodecPrivate` is not present.
```

Sorry, I forgot the second part of my thoughts.
Not "legacy" because it is like if it was at some point valid.

JeromeMartinez: > Otherwise I'll switch it to SHOULD

I don't know enough VP9, but if it is not as e.g. AVC which has SPS at every key frame but if  list of specific VP9 codec features is only in the first frame, I prefer MUST for a muxer because it is not good to have to read the first frame if we want to seek immediately to another frame.

robUx4: > I don't know enough VP9, but if it is not as e.g. AVC which has SPS at every key frame but if list of specific VP9 codec features is only in the first frame

The "Level, Profile, Bit Depth and Chroma Subsampling values" information described in the [WebM Guidelines](https://www.webmproject.org/docs/container/#webm-guidelines) are not supposed to change throughout the whole `Segment`. I don't know about the framing either, but I except at least keyframes to have this information. And on WebM it's a **SHOULD** so it is assuming that starting in the middle of a file without the `CodecPrivate` should still play fine.

robUx4: I applied your last proposal.

robUx4:
For the record the `CodecPrivate` is important to differentiate between 8-bit and 10-bit VP9 streams. Even Matroska/WebM scanners should not have to peak into the frames to know what the file contains. And as said about it helps picking the proper hardware decoder before "consuming" a (first) frame.

JeromeMartinez (approved):

PR #874, opened 2024-10-25
add A_AC4 codec

robUx4:
The documentation is freely available.

Unlike other codecs added until now, this one is not supported anywhere yet. So it may be kept for Matroska v5.


robUx4:
There's not even a decoder in FFmpeg, so no rush...

JeromeMartinez:
> So it may be kept for Matroska v5.

IMO codec spec is not tied to Matroska versioning, it is more a registry and there is no obligation to any player to support any stream format for matching a Matroska version.

robUx4:
I mean it will go in the v5 document, not in the codec document we're finalizing.

PR #880, opened 2024-11-02
Add more tag formatting constraints

robUx4:
And define some generic types that should be parsed the same (numbers, dates, country codes).

Refer to tag targets with the `TargetTypeValue` which is mandatory, unlike the `TargetType` string.

And fix some odd wording here and there.

### DIFF
--- a/matroska_tags.xml
+++ b/matroska_tags.xml
@@ -364,10 +364,10 @@ This holds the same information as the "MCDI" in [@!ID3v2.3].</description>
       <description lang="en">The TV Database [@!TheTVDB] tag which can include movies. The variable length digits string representing a "Series ID", "Episode ID" or "Movie ID" identifier **MUST** be prefixed with "series/", "episodes/", or "movies/", respectively.</description>
     </tag>
     <tag name="PURCHASE_ITEM" class="Commercial" type="UTF-8">
-      <description lang="en">URL to purchase this file. This is akin to the "WPAY" tag in [@!ID3v2.3].</description>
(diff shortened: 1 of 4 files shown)


matroska_tags.xml
```diff
@@ -365,17 +364,17 @@ This holds the same information as the "MCDI" in [@!ID3v2.3].</description>
...
+      <description lang="en">Information on where to purchase this album using the URL format defined in [@!RFC3986]. This is akin to the "WCOM" tag in [@!ID3v2.3].</description>
     </tag>
     <tag name="PURCHASE_OWNER" class="Commercial" type="UTF-8">
       <description lang="en">Information on the person who purchased the file. This is akin to the "TOWN" tag in [@!ID3v2.3].</description>
     </tag>
     <tag name="PURCHASE_PRICE" class="Commercial" type="UTF-8">
-      <description lang="en">The amount paid for entity. There **SHOULD** only be a numeric value in here. Only numbers,
-no letters or symbols other than ".". For instance, you would store "15.59" instead of "$15.59USD".</description>
+      <description lang="en">The amount paid for entity, using the Float number defined in (#number-tags-formatting).
+      The currency is not included. For instance, you would store "15.59" instead of "$15.59USD".</description>
```
retokromer, line 377: Is `$15.59USD` a form used anywhere IRL?

robUx4: It's definitely not written by a USAn 🤣 
However for this example it's serves the purpose of explaining that the unit should not be there. Neither before nor after the number.

PR #885, opened 2024-11-03
Add a rule for tag replacement

robUx4:
~~Draft on top of #884 as I'm not sure it's a rule that makes sense.~~ However it's implied by the fact lower levels should void tags that are set by the parent.

A better rule would be to RECOMMEND not to use an upper value if it doesn't apply to all children.

### DIFF
--- a/tagging.md
+++ b/tagging.md
@@ -150,6 +150,10 @@ If some tracks of that CD have no known
 "ARTIST", the value **MUST** be set to nothing, a void string "" as detailed in [@!RFC9559, section 24.2],
 so that the album "ARTIST" doesn't apply.
 
+If a tag with a given `TagName` is found at a `TargetTypeValue`,
+only values of that `TagName` are valid at that `TargetTypeValue` level.
+In other words, the `TagName` values from upper `TargetTypeValue` levels don't apply at that level.
+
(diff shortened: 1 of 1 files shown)


robUx4:
The wording is probably not right. In case of an "ARTIST" it can be the album artist or the track artist. One doesn't **replace** the other, they are complementary, both valid at their level.

[RFC9559](https://www.rfc-editor.org/rfc/rfc9559#section-24.2) says:

> If an upper value doesn't apply to a level but the actual value to use is not known, an empty TagString ([Section 5.1.8.1.2.5](https://www.rfc-editor.org/rfc/rfc9559#tagstring-element)) or an empty TagBinary ([Section 5.1.8.1.2.6](https://www.rfc-editor.org/rfc/rfc9559#tagbinary-element)) MUST be used as the tag value for this level. 

This is compatible with the fact it doesn't replace the value. The void is only used to explicitly not inherit an upper value.

robUx4:
I'll be merging this today and publish a new Tags document unless someone has issues.

IMO it's OK because it doesn't void the values above. Just clarifies how reset values for a level.

retokromer (approved):

PR #886, opened 2024-11-03
make the VP9 CodecPrivate recommended but not mandatory

### LINKED ISSUE #875: VP9 CodecPrivate should be SHOULD
I noticed a lot of my VP9/WebM test files don't have a `CodecPrivate` set. Looking at the source it's usually files created with FFmpeg. I looked at the libavformat code and even with the [latest code](https://github.com/FFmpeg/FFmpeg/blob/master/libavformat/matroskaenc.c#L1116), there is nothing to generate the `CodecPrivate` for VP9.

Even libwebm muxer sample code has [some code](https://github.com/webmproject/libwebm/blob/6c1e511f487eaea026d94d0c6494a617155b5cfb/mkvmuxer_sample.cc#L556) to set the Profile and Level values, but only if the values are provided in the command line.

IMO we should use **SHOULD** for the VP9 `CodecPrivate` rather because most files in the wild don't have this value set. That's also the recommandation in the WebM spec. And until someone fixes libavformat it's going to be missing from a lot of files.

robUx4:
libavformat doesn't set this information and it's optional in libwebm.

Fixes #875.

### DIFF
--- a/codec_specs.md
+++ b/codec_specs.md
@@ -463,8 +463,10 @@ Description: VP9 is an open and royalty free video compression format developed 
 
 Codec BlockAdditions: A single-channel encoding of an alpha channel **MAY** be stored in `BlockAdditions`. The `BlockAddId` of the `BlockMore` containing these data **MUST** be 1.
 
-Initialization: The `CodecPrivate` **MUST** contain a list of specific VP9 codec features as described in the "VP9 Codec Feature Metadata" section of [@!WebMContainer].
-Due to the existence of some old non compliant muxers which were not providing `CodecPrivate`, the decoder **SHOULD** reconstruct it from the first frame in the case `CodecPrivate` is not present.
+Initialization: The `CodecPrivate` **SHOULD** contain a list of specific VP9 codec features as described in the "VP9 Codec Feature Metadata" section of [@!WebMContainer].
+This piece of data helps to select a decoder on playback, but as many muxers don't provide the `CodecPrivate` for "V_VP9" it's not a hard requirement.
+It is possible for the decoder to reconstruct the "VP9 Codec Feature Metadata" from the first frame in case the `CodecPrivate` is not present.
+
 Note that the format differs from the `VPCodecConfigurationRecord` structure, as defined in [@VP-ISOBMFF].
 
 ## Audio Codec Mappings


codec_specs.md
```diff
@@ -463,8 +463,10 @@ Description: VP9 is an open and royalty free video compression format developed
 
 Codec BlockAdditions: A single-channel encoding of an alpha channel **MAY** be stored in `BlockAdditions`. The `BlockAddId` of the `BlockMore` containing these data **MUST** be 1.
 
-Initialization: The `CodecPrivate` **MUST** contain a list of specific VP9 codec features as described in the "VP9 Codec Feature Metadata" section of [@!WebMContainer].
-Due to the existence of some old non compliant muxers which were not providing `CodecPrivate`, the decoder **SHOULD** reconstruct it from the first frame in the case `CodecPrivate` is not present.
+Initialization: The `CodecPrivate` **SHOULD** contain a list of specific VP9 codec features as described in the "VP9 Codec Feature Metadata" section of [@!WebMContainer].
+Many muxers don't provide the `CodecPrivate` for "V_VP9" so it's not a hard requirement. However it is **RECOMMENDED** to provide this information that can help select a decoder on playback.
+It is possible for the decoder to reconstruct the "VP9 Codec Feature Metadata" from the first frame in the case `CodecPrivate` is not present.
```
mbunkus, line 468: Minor nitpicks:

- "…in the case…" is usually used simply as "…in case…"
- The whole sentence "Many muxers don't…" might read better as: "As many muxers don't provide the `CodecPrivate` for "V_VP9" tracks, it's not a hard requirement."
- I think there should be a comma after the "However" in "However it is…", as in: "However, it is…"

Apart from these I'm fine with the change. Thanks.

mbunkus (changes requested):

codec_specs.md
robUx4: I made the suggested changes. We do have a comma after all the "However" usage.

robUx4:
BTW for the record, I was checking how WebKit handles WebM and it turns out it's [using libwebm](https://github.com/WebKit/WebKit/blob/635a079e5128f24a69f90de3141943153e46e5bb/Source/WebCore/platform/graphics/cocoa/SourceBufferParserWebM.h#L38).

codec_specs.md
```diff
@@ -463,8 +463,10 @@ Description: VP9 is an open and royalty free video compression format developed
 
 Codec BlockAdditions: A single-channel encoding of an alpha channel **MAY** be stored in `BlockAdditions`. The `BlockAddId` of the `BlockMore` containing these data **MUST** be 1.
 
-Initialization: The `CodecPrivate` **MUST** contain a list of specific VP9 codec features as described in the "VP9 Codec Feature Metadata" section of [@!WebMContainer].
-Due to the existence of some old non compliant muxers which were not providing `CodecPrivate`, the decoder **SHOULD** reconstruct it from the first frame in the case `CodecPrivate` is not present.
+Initialization: The `CodecPrivate` **SHOULD** contain a list of specific VP9 codec features as described in the "VP9 Codec Feature Metadata" section of [@!WebMContainer].
+As many muxers don't provide the `CodecPrivate` for "V_VP9" so it's not a hard requirement. However, it is **RECOMMENDED** to provide this information that can help select a decoder on playback.
```
JeromeMartinez, line 467: IMO the second sentence is redundant (same meaning as the "SHOULD" one line above) and can be removed.

robUx4: They are not giving the same information. The first explain why it's not a MUST. The other why it should be a MUST but it's not.

JeromeMartinez: "However, it is **RECOMMENDED** to provide this information that can help select a decoder on playback." is synonym of "The `CodecPrivate` **SHOULD** contain".

Or I would rephrase for avoiding 2x the IETF wording (SHOULD & RECOMMENDED), overall it would be:
```suggestion
Initialization: The `CodecPrivate` **SHOULD** contain a list of specific VP9 codec features as described in the "VP9 Codec Feature Metadata" section of [@!WebMContainer].
This piece of data helps to select a decoder on playback, but as defacto many muxers don't provide the `CodecPrivate` for "V_VP9" so it's not a hard requirement.
It is possible for the decoder to reconstruct the "VP9 Codec Feature Metadata" from the first frame in the case `CodecPrivate` is not present.
```

robUx4: OK, I took your suggestion for the second sentence and slightly tweaked it.

JeromeMartinez (approved):

retokromer (approved):

PR #884, opened 2024-11-03
Make PARTS and TargetTypeValue rules more explicit

robUx4:
And add a basic audio example.

### DIFF
--- a/matroska_tags.xml
+++ b/matroska_tags.xml
@@ -41,7 +41,7 @@ All tags can be used "under" the "COUNTRY" tag like "LABEL", "RATING", etc.
 the total number of tracks of an audio CD).</description>
     </tag>
     <tag name="PART_NUMBER" class="Organization Information" type="UTF-8">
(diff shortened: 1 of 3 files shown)


robUx4:
The replacement rule is a bit tricky as for the TITLE the album Title should still be kept as such. Maybe we should not use any replacement and keep each element at its level.

matroska_tags.xml
```diff
@@ -11,7 +11,9 @@
     </class>
     <class name="Entities"/>
     <class name="Search and Classification"/>
-    <class name="Temporal Information"/>
+    <class name="Temporal Information">
+      <description lang="en">All tags in this section use the Date format defined in (#date-tags-formatting)..</description>
```
retokromer, line 15: Are the two periods at the end correct?

robUx4: No..

robUx4: Fixed. Thanks for your attention!

PR #891, opened 2024-11-10
Add T.35 reference

robUx4:


codec_specs.md
```diff
@@ -1131,13 +1131,13 @@ Block type name: Opaque data
...
 
-### ITU T.35 metadata
+### ITU T.35 Metadata
 
 Block type identifier: 4
 
 Block type name: ITU T.35 metadata
 
-Description: the `BlockAdditional` data is interpreted as ITU T.35 metadata, as defined by ITU-T T.35
+Description: the `BlockAdditional` data is interpreted as ITU T.35 metadata, as defined by [@!ITU-T.35]
```
robUx4, line 1140: Probably not a non normative reference because it's not the actual data description of various data.

robUx4:
It's still unclear how this is supposed to be mapped but we can take care of that later.

PR #898, opened 2024-11-11
update the ISO 14496-15 reference to the latest version

robUx4:
There's still not free version of this and we have 4 codecs that rely on it.


retokromer (approved):

robUx4:
Fun fact the chinese codec AVS3 is also possible in DVB and the referenced spec is in Chinese... (also not available for free) even though there's a IEEE version (also not free).

PR #902, opened 2024-11-16
reference FLAC as RFC9639

robUx4:
Draft for now as it won't build until it's officially published. For now it's available at https://www.rfc-editor.org/authors/rfc9639.html

### DIFF
--- a/codec_specs.md
+++ b/codec_specs.md
@@ -683,8 +683,7 @@ Codec ID: A_FLAC
 
 Codec Name: FLAC (Free Lossless Audio Codec)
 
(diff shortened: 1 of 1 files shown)


retokromer (approved):

retokromer:
RCF 9639 should be published quite soon.

ktmf01:
Perhaps the whole section can be replaced by a reference to section 10.2 of RFC 9639?

robUx4:
> Perhaps the whole section can be replaced by a reference to section 10.2 of RFC 9639?

That would indeed be a lot better.

robUx4:
It will create a nice link to the appropriate section in the HTML version.

PR #910, opened 2024-11-17
refer older MPEG audio/video to their ISO specs

robUx4:
Which are not freely available 30+ years later.


retokromer:
>Which are not freely available 30+ years later.

On the contrary, they are outrageously expensive.

retokromer (approved):

PR #911, opened 2024-11-19
fix the AV1 CodecPrivate

robUx4:
It's the `AV1CodecConfigurationBox`, not the whole `AV1CodecConfigurationRecord`. It was already defined that way in https://github.com/ietf-wg-cellar/matroska-specification/blob/master/codec/av1.md#codecprivate-1

The bug comes from 9c9b0c34a25b316109263ae6c79e93d03497aee5

### DIFF
--- a/codec_specs.md
+++ b/codec_specs.md
@@ -152,7 +152,7 @@ A `Block` inside a `BlockGroup` **MUST** use `ReferenceBlock` elements if the fi
 A `Block` inside a `BlockGroup` **MUST** use `ReferenceBlock` elements if the `Block` doesn't contain a `Sequence Header OBU`.
 A `Block` with `frame_header_obu` where the `frame_type` is `INTRA_ONLY_FRAME` **MUST** use a `ReferenceBlock` with a value of 0 to reference itself.
 
-Initialization: The `CodecPrivate` consists of the `AV1CodecConfigurationRecord` described in section 2.3 of [@!AV1-ISOBMFF].
+Initialization: The `CodecPrivate` consists of the `AV1CodecConfigurationBox` described in section 2.3 of [@!AV1-ISOBMFF].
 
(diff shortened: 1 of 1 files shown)


codec_specs.md
```diff
@@ -152,7 +152,7 @@ A `Block` inside a `BlockGroup` **MUST** use `ReferenceBlock` elements if the fi
 A `Block` inside a `BlockGroup` **MUST** use `ReferenceBlock` elements if the `Block` doesn't contain a `Sequence Header OBU`.
 A `Block` with `frame_header_obu` where the `frame_type` is `INTRA_ONLY_FRAME` **MUST** use a `ReferenceBlock` with a value of 0 to reference itself.
 
-Initialization: The `CodecPrivate` consists of the `AV1CodecConfigurationRecord` described in section 2.3 of [@!AV1-ISOBMFF].
+Initialization: The `CodecPrivate` consists of the `AV1CodecConfigurationBox` described in section 2.3 of [@!AV1-ISOBMFF].
```
mbunkus, line 155: What!? Nooooo, `Record` is correct! `Box` is wrong as `Box` contains the leading `avC1` bytes, which aren't used in Matroska.

JeromeMartinez: > What!? Nooooo, Record is correct! Box is wrong as Box contains the leading avC1 bytes, which aren't used in Matroska.

+1, [av1.md#codecprivate-1](https://github.com/ietf-wg-cellar/matroska-specification/blob/master/codec/av1.md#codecprivate-1) is wrong, https://aomediacodec.github.io/av1-isobmff/#av1codecconfigurationbox-syntax AV1CodecConfigurationBox (so Box) header is not in Matroska (not 4 bytes for size and no "avc1"string).

robUx4: Damn it, MP4 confuses me.

robUx4:
`AV1CodecConfigurationRecord` is the short one and is correct.

PR #917, opened 2024-11-24
add A_AAC codec ID

robUx4:
The "stripped ADTS header" seems to correspond more or less to the `raw_data_block()` in the spec.


JeromeMartinez (approved):
I would use SHOULD instead of MUST for the legacy CodecIDs as it is expected to be supported by decoders, legacy encoders may use that and I prefer to avoid considering such stream as invalid but not blocking.

retokromer (approved):

robUx4:
I double checked that `A_AAC` is matching the `raw_data_block()` in FFmpeg (done through `aac_adtstoasc`).

PR #918, opened 2024-11-24
Add more codec references

robUx4:
And cleanup some text


JeromeMartinez (approved):
I wouldn't describe UCI and VfW here as they are implementations and not a technical description of the stream based on specs, but not blocking.

robUx4:
> I wouldn't describe UCI and VfW here as they are implementations and not a technical description of the stream based on specs, but not blocking.

Yes, for now I kept a similar text. I need to dig deeper in the specs to see if there's a proper reference to explain the differences. I don't think we need to care about hacks needed in AVI to store the same codecs.

PR #920, opened 2024-11-24
include the Wavpack document in the codec spec

robUx4:
And rework the text a bit to be more readable. There's probably a bit more work needed.

### DIFF
--- a/wavpack.md
+++ b/wavpack.md
@@ -7,7 +7,9 @@
 
 [@!WAVPACK] stores each data in variable length frames. That means each frame can have a different number of samples.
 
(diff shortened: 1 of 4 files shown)


wavpack.md
```diff
@@ -114,41 +72,47 @@ To save space and avoid redundant information in Matroska we remove data from th
...
-}
-[ block data # 3 ]
-...
+[ correction block data ]
 ```
 
-* BlockAdditional (level 1)
+#### Multichannel
+
+The `BlockAdditional` data contain the following header with the data from the each `WavpackHeader` of the WavPack block of the correction file
```
retokromer, line 114: `contain` > `contains`

robUx4: "data" is plural in English (and latin). If it's changed to `BlockAdditional` element, it would work. And it may be more explicit too.

PR #921, opened 2024-11-27
Add ``application/x-font-otf`` as a legacy MIME type

moi15moi:
In a recent MKVToolNix [commit](https://gitlab.com/mbunkus/mkvtoolnix/-/commit/d858c2bf138009b8316b32c4554397ecb609d938), ``application/x-font-otf`` was defined as a legacy MIME type. It might be a good idea to include a mention of this in the documentation

### DIFF
--- a/attachments.md
+++ b/attachments.md
@@ -91,6 +91,8 @@ Therefore, it is **RECOMMENDED** for a `Matroska Player` to support the followin
 
 * application/x-font-ttf: TrueType fonts, equivalent to `font/ttf`
 
+* application/x-font-otf: OpenType Layout fonts, equivalent to `font/otf`
+
 * application/vnd.ms-opentype: OpenType Layout fonts, equivalent to `font/otf`
 
 * application/font-sfnt: Generic SFNT Font Type, equivalent to `font/sfnt`


robUx4:
Good catch! We will need to prepare an errata for this.

moi15moi:
Should I do it? If so, how?

robUx4:
You can't. We have to create that for the IETF, once this is merged.

mcr:
Steve Lhomme ***@***.***> wrote:
    > Good catch! We will need to prepare an errata for this.

I don't think we can't add a mime type with errata, as it requires expert review.

Please bring this up on the list: it might be that the Media type experts can
just bless x-font-otf as a legacy.  If not, we might need an actual
additional RFC, but let's discuss this.

robUx4:
> Please bring this up on the list: it might be that the Media type experts can just bless x-font-otf as a legacy.  If not, we might need an actual additional RFC, but let's discuss this.

Could we just add this in the Matroska v5 document ? It will contain new elements but also other things.

mcr:
Steve Lhomme ***@***.***> wrote:
    > Could we just add this in the Matroska v5 document ? It will contain
    > new elements but also other things.

maybe.
I don't know the timing of that.  or even if we have to bother.
I'll put this on the agenda.

moi15moi:
Any news about this PR?
I will rebase it if it has a chance to get merge.

rjsparks:
@moi15moi - It's complicated - see https://mailarchive.ietf.org/arch/msg/cellar/P39JqNi0xxMgZRdbm_1IX_uGk84/
This PR won't be the way to get it registered in any case.

PR #926, opened 2024-12-15
bump Codec draft version after 2024-12-15 v14 tag

robUx4:


robUx4:
Since the v14 draft is expired, let's update it.

PR #935, opened 2024-12-22
allow multiple comment blocks in WebVTT BlockAdditions

### LINKED ISSUE #394: Multiple WebVTT comment blocks
The WebVTT codec mapping currently has this to say about WebVTT comments:

"The third and all following lines contain all WebVTT Comment Blocks that precede the current WebVTT Cue Block. These may be absent."

This does not say how multiple comment blocks are to be stored. mkvmerge separates them by a single \n, but this has the disadvantage that it doesn't allow to distinguish between different comment blocks and a single comment block that just happens to contain a NOTE in the middle after a newline. (Indeed, two comment blocks get merged when doing a mkvmerge->mkvextract roundtrip.) I therefore propose to use two line feed characters to distinguish different comment blocks.

Furthermore, mkvmerge strips comment blocks of their terminating WebVTT line terminator; this is currently against the spec, but I don't see a reason why this shouldn't be allowed. It is similar to not including the terminating line terminator of a WebVTT cue block.

robUx4:
And allow stripping the new lines at the end.

Fixes #394

### DIFF
--- a/subtitles.md
+++ b/subtitles.md
@@ -397,6 +397,8 @@ Each Matroska Block may be accompanied by one `BlockAdditions` element. Its form
 
 3.  The third and all following lines contain all `WebVTT comment block`(s) that precede
     the current `WebVTT cue block`. These may be absent.
+    Each `WebVTT comment block` includes its `WebVTT line terminator` and is followed by one line feed character (U+0x000a).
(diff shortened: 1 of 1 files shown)


robUx4:
cc @mkver

retokromer (approved):

PR #951, opened 2025-02-23
make tag strictness paragraph non normative

robUx4:
This is just generic text that can't be enforced one way or another.


retokromer (approved):
Indeed.

SpencerDawkins (approved):
This is perfect.

PR #952, opened 2025-02-23
rename 'Identifiers" section to 'External Identifiers"

robUx4:
Due to the xml2rfc HTML formatting, identifiers is a special keyword. This results in the section being presented slightly differently. And the links in the IANA section don't work.


robUx4:
Merging to fix the links, we'll see if there can be a better name.

PR #961, opened 2025-02-23
make country code string formatting a MUST

robUx4:
The SHOULD was added in 2734e4053ff6ba12d59bf78906ecb89e7b905fa8. But it was not present in the original spec which just defined the Matroska country code format.


retokromer (approved):

SpencerDawkins (approved):
This looks perfect.

PR #962, opened 2025-02-23
make the INSTRUMENTS always a child of other tags

robUx4:
But we can't set a definitive list of potential parents. There might be more in the future.

### DIFF
--- a/matroska_tags.xml
+++ b/matroska_tags.xml
@@ -68,7 +68,7 @@ Or "Marley Bob" and "Marley Robert Nesta" (no comma needed).</description>
     </tag>
     <tag name="INSTRUMENTS" class="Nested Information" type="UTF-8">
       <description lang="en">The instruments that are being used/played, separated by a comma.
-It **SHOULD** be a child of the following tags: "ARTIST", "LEAD_PERFORMER", or "ACCOMPANIMENT".</description>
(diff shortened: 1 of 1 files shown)


matroska_tags.xml
```diff
@@ -68,7 +68,7 @@ Or "Marley Bob" and "Marley Robert Nesta" (no comma needed).</description>
     </tag>
     <tag name="INSTRUMENTS" class="Nested Information" type="UTF-8">
       <description lang="en">The instruments that are being used/played, separated by a comma.
-It **SHOULD** be a child of the following tags: "ARTIST", "LEAD_PERFORMER", or "ACCOMPANIMENT".</description>
+It **MUST** be a child of the another tag, including the following: "ARTIST", "LEAD_PERFORMER", or "ACCOMPANIMENT".</description>
```
SpencerDawkins, line 71:
```suggestion
It **MUST** be a child of another tag, including the following: "ARTIST", "LEAD_PERFORMER", or "ACCOMPANIMENT".</description>
```

SpencerDawkins (changes requested):
I just corrected a nit - otherwise, this is perfect.

matroska_tags.xml
robUx4: Applied. (the Github commit system doesn't produce nice commits in the history, nor credit the author of the change)

SpencerDawkins (approved):
This now seems perfect.

PR #963, opened 2025-02-23
turn the ADDRESS country code into a MAY

robUx4:
We probably don't care too much as it's not meant to be parsed, especially as there's no definitive format.
Anything that can be put on an envelope should be OK.

### DIFF
--- a/matroska_tags.xml
+++ b/matroska_tags.xml
@@ -74,7 +74,7 @@ It **SHOULD** be a child of the following tags: "ARTIST", "LEAD_PERFORMER", or "
       <description lang="en">Email corresponding to the tag it's included in, using the "Addr-Spec" format defined in [@!RFC5322, section 3.4.1].</description>
     </tag>
     <tag name="ADDRESS" class="Nested Information" type="UTF-8">
-      <description lang="en">The physical address of the entity. The address **SHOULD** include a country code using the Country Code format defined in (#country-code-tags-formatting).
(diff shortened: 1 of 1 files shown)


matroska_tags.xml
```diff
@@ -74,7 +74,7 @@ It **SHOULD** be a child of the following tags: "ARTIST", "LEAD_PERFORMER", or "
       <description lang="en">Email corresponding to the tag it's included in, using the "Addr-Spec" format defined in [@!RFC5322, section 3.4.1].</description>
     </tag>
     <tag name="ADDRESS" class="Nested Information" type="UTF-8">
-      <description lang="en">The physical address of the entity. The address **SHOULD** include a country code using the Country Code format defined in (#country-code-tags-formatting).
+      <description lang="en">The physical address of the entity. The address **MAY** include a country code using the Country Code format defined in (#country-code-tags-formatting).
```
SpencerDawkins, line 77:
```suggestion
      <description lang="en">The physical address of the entity. The address **MAY** include a country code using the Country Code format defined in (#country-code-tags-formatting).
```
This one is a bit complicated. It's saying 

> **MAY** include a country code using the Country Code format defined in (#country-code-tags-formatting)

but I think you're saying something like 

> **MAY** include a country code.
> If a country code is included, the Country Code format defined in (#country-code-tags-formatting) is preferred.

robUx4: Yes it does make sense and avoids a potential confusion. 

I turned the requirement into a RECOMMENDED because existing tags may already use other formats. And there may be other, more standardized ways to define a physical address.

retokromer (approved):

SpencerDawkins (approved):
This now seems perfect.

PR #964, opened 2025-02-23
reword the LYRICS text

robUx4:
doublon is not used in English. And it's clearer to explain the double edition.

### DIFF
--- a/matroska_tags.xml
+++ b/matroska_tags.xml
@@ -105,7 +105,7 @@ This is akin to the "TPE1" tag in [@!ID3v2.3] when the `TargetTypeValue` is 30 (
     <tag name="LYRICS" class="Entities" type="UTF-8">
       <description lang="en">The lyrics corresponding to a song, in case audio synchronization is not known
 or as a duplicate of a subtitle track. Editing this value, when it is a duplicate of a subtitle track, **SHOULD** also result
-in editing the subtitle track for more consistency.</description>
+in editing the subtitle track for more consistency, and vice versa.</description>
     </tag>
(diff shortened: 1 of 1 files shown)


retokromer (approved):

matroska_tags.xml
```diff
@@ -103,9 +103,9 @@ This is akin to the "TPE1" tag in [@!ID3v2.3] when the `TargetTypeValue` is 30 (
       <description lang="en">The name of a person who arranged the piece (e.g., Ravel).</description>
     </tag>
     <tag name="LYRICS" class="Entities" type="UTF-8">
-      <description lang="en">The lyrics corresponding to a song (in case audio synchronization is not known
-or as a doublon to a subtitle track). Editing this value, when subtitles are found, **SHOULD** also result
-in editing the subtitle track for more consistency.</description>
+      <description lang="en">The lyrics corresponding to a song, in case audio synchronization is not known
+or as a duplicate of a subtitle track. Editing this value, when it is a duplicate of a subtitle track, **SHOULD** also result
```
SpencerDawkins, line 107: Is this a SHOULD, or a MUST?

robUx4: Given there is nothing in the file that tells that this LYRICS tag correspond to the data of a particular subtitle track, this can't be enforced by a tool. It can only be a human that needs to extract the subtitles, edit them, remux them. So it's a best effort option. Failure to do it will not make the file invalid.

This could be RECOMMENDED. I have no strong opinion about either. But I don't think it's a MUST.

SpencerDawkins (approved):
This seems reasonable, and thank you, @robUx4, for explaining the plan to me.

PR #966, opened 2025-02-23
make less normative reference to official tag reasons

robUx4:


tagging.md
```diff
@@ -533,8 +533,9 @@ This corresponds to this layout of EBML elements:
 
 # Official Tags
 
-The following is a complete list of the supported Matroska Tags. While it is possible
-to use Tag names that are not listed below, this is **NOT RECOMMENDED** as compatibility will
+The following is a complete list of the supported Matroska Tags.
+As stated in (#why-official-tags-matter) it is better to use official tags in this as much as possible,
```
SpencerDawkins, line 537:
```suggestion
As stated in (#why-official-tags-matter) it is better to use official tags as much as possible,
```

SpencerDawkins (changes requested):
This is a minor nit - I couldn't parse "in this" clearly.

tagging.md
robUx4: Applied. I don't remember what I want to express with "this".

retokromer (approved):

retokromer (approved):

SpencerDawkins (approved):
This is now perfect.

Issue #969, opened 2025-02-26
rework the files layout for more clarity


robUx4:
As discussed during the last meeting, we can make it easier for people for find what they need to edit but splitting each document we generate into its own folder with its own sources. There is no shared files (apart from ebml_matroska.xml).

This will likely break existing merge requested that are not merged yet.

PR #971, opened 2025-03-09
Move most documents to produce the XML output in separate folders

robUx4:
First step towards #969. The file moves are clean git rename so rebasing existing code (include Pull Request seems straightforward).

The files used to generate other files are not moved for now.

The .xml/.html/.txt files are still generated at the root.


robUx4:
Looking at the [resulting tree](https://github.com/robUx4/matroska-specification/tree/folder-split), it's still hard to find where the document sources are. I'll prefix with "cellar-"

robUx4:
Although a rebase is straightforward, Github generates merge conflicts, so each PR will need to be rebased manually...

arjunsatarkar:
Just to note, I think this or a related change has broken all the links in the sidebar ("Main Menu") of <https://ietf-wg-cellar.github.io/matroska-specification/>.

PR #974, opened 2025-03-09
jekyll: remove more build/generated files

robUx4:
It copies a lot of markdown files but doesn't seem to turn them into HTML...


robUx4:
It seems the proper place to handle the website is in https://github.com/Matroska-Org/infrastructure

Should we remove all these files from here @mbunkus ?

PR #978, opened 2025-04-06
add "i.e." to EBML path notice

robUx4:
Originally added by 2df8f5aaa173b4ee64ecc967b1cf0a96a447813c


retokromer:
Why do you leave the two dashes before the "i.e."?

robUx4:
> Why do you leave the two dashes before the "i.e."?

That's how it was done since RFC 8794 so I kept doing it that way.

retokromer (approved):

PR #986, opened 2025-04-06
reword the v7 vobsub requirement

robUx4:


cellar-codec/subtitles.md
```diff
@@ -21,9 +21,9 @@ Here is a list of pointers for storing subtitles in Matroska:
 The first image format that is a goal to import into Matroska is the VobSub subtitle format.
 This subtitle type is generated by exporting the subtitles from a DVD [@?DVD-Video].
 
-The requirement for muxing VobSub into Matroska is v7 subtitles (see first line of the .IDX file).
-If the version is smaller, you must remux them using the SubResync utility from
-VobSub 2.23 (or MPC) into v7 format. Generally any newly created subs will be in v7 format.
+If the subtitle version in the .IDX file is less than v7,
+the subtitles must be remuxed using the SubResync utility from VobSub 2.23 (or MPC) into v7 format.
```
retokromer, line 25: Isn’t the `must` here a `MUST`?

rjsparks: I don't think so? 8174/2119 keywords describe what implementations have to do for protocol interoperability. I think this should be something more like "one way to remux the subtitles is to use the SubResync utility from..."

robUx4: I added @rjsparks's text and some more about why it should be remuxed.

rjsparks (approved):

retokromer (approved):

PR #995, opened 2025-04-27
BlockAddIDValue 4 is not recommended

robUx4:
Rebased version of #745 using the normative RECOMMENDED word as [suggested](https://github.com/ietf-wg-cellar/matroska-specification/pull/745#discussion_r1149215591) and some links to the T.35 block addition spec.

I'll merge this quickly but it can still be changed before we send it to the RFC editors (and also discussed in the upcoming meeting).


robUx4:
poke again @JeromeMartinez

PR #996, opened 2025-04-27
Clarification about when CodecPrivate is mandatory

robUx4:
Rebased version #446 where a lot of the added Initialization where already added earlier.
This one line is the only actual addition that is left.

I'll merge this quickly but it can still be changed before we send it to the RFC editors (and also discussed in the upcoming meeting).


SpencerDawkins (approved):
@robUx4, thank you for helping me understand what we're doing here.

PR #998, opened 2025-04-27
ensure CodecIDs don't use values already in the registry

robUx4:


SpencerDawkins (approved):
This is perfect, and captures your last clarification to me.

retokromer (approved):

PR #1001, opened 2025-04-27
remove comment in vobsub example

robUx4:
We are not defining the vobsub constraints


SpencerDawkins (commented):
Sometimes making problematic text go away is the right thing to do ...

PR #1016, opened 2025-06-13
Fix a sample blockmoreid

dericed:


robUx4 (approved):
LGTM

There's an issue with the line length of the generated text, but I'll fix it afterwards:

```
draft-ietf-cellar-codec-15.xml(1926): Warning: Too long line found (L3393), 96 characters longer than 72 characters: 
             <BlockAdditional>01:00:00:00</BlockAdditional><!--presented as a string for readability but should use binary encoding defined in the associated mapping-->
```

PR #1019, opened 2025-06-29
add an ABNF description of the date and time format in tags

robUx4:
Inspired by https://www.rfc-editor.org/rfc/rfc3339#page-12


robUx4:
I removed leftover text copied from RFC3339 that doesn't apply with our simplified version.

PR #1080, opened 2026-03-31
Add codec mapping for VC-1

arch1t3cht:
VC-1 can currently only be muxed as a `V_MS/VFW/FOURCC` track, which can create complications when demuxing since such tracks are usually muxed with decoding timestamps rather than presentation timestamps. This commit adds a proper codec mapping for VC-1 emulating its carriage in ISO-BMFF files in order to improve this situation.


cellar-codec/codec_specs.md
```diff
@@ -441,6 +441,16 @@ In addition the packing of RGB, YUV, etc. pixels **MUST** be declared with a `Tr
 
 Initialization: none
 
+### V_VC1
+
+Codec ID: V_VC1
+
+Codec Name: VC-1
+
+Description: Individual pictures (which could be a frame, or 2 fields having the same timestamp) of VC-1 stored as described in [@!SMPTE.RP2025-2007].
```
robUx4: This seems a bit vague. Given the mapping is the same as ISOBMFF for the CodecPrivate, it would be easier/cleaner to reference the corresponding ISOBMFF structure as well. Or a similar wording as in "5 VC-1 Sample definition" depending on the profile.

robUx4 (changes requested):

cellar-codec/codec_specs.md
arch1t3cht: Fair point, I originally used the H.264 mapping as a base. I now updated it to simply reference the "VC-1 sample" definition.

robUx4 (approved):

arch1t3cht:
Thanks!

robUx4:
No, thank YOU ;)

PR #1081, opened 2026-04-06
tags: rename official targettypes to assigned

robUx4:
Same as for tag names.


eckelcu (approved):
Looks good to me. Thanks.

retokromer (approved):

PR #1082, opened 2026-04-06
Use less "we" subject

robUx4:
There is one remaining which IMO is OK: "we would end up with 7 "TITLE_" tag names"

On top of #1081 as we change the same line.


retokromer (approved):

eckelcu (changes requested):
The changes look good.

For the remaining instance of "we" in "we would end up with 7 "TITLE_" tag names”; how about “… otherwise there would be a need for 7 “TITLE” tag names.”?

robUx4:
I opted for an even more simple wording `otherwise 7 "TITLE_" tag names would needed`.

PR #1083, opened 2026-04-06
tags: remove normative emphasis of IANA guideliness

robUx4:
The text has been moved out of the IANA considerations but was still normative.

Follow-up to #1075


retokromer (approved):

eckelcu (commented):
I think this was ok, and perhaps better, as it was.

robUx4:
> I think this was ok, and perhaps better, as it was.

It's not clear whether you agree to merge this or not. I think it's goes in the right direction given there should not be any normative text in that part of the document. I'll merge it as it is. If we need more changes it can still be done...

PR #1094, opened 2026-04-26
Rename draft-ietf-cellar-matroska5-01 to draft-ietf-cellar-matroskanext-00

robUx4:
For now the `ebml_matroska.xml` stil refers to new elements as `minver="5"`.

The `cellar-matroska5` folder is also renamed `cellar-matroskanext`


retokromer (approved):

robUx4:
Dropping in favor of #1101 after the naming has been agreed in the last meeting.

PR #1120, opened 2026-08-09
codec: discard unknown codecs

robUx4:


cellar-codec/codec_specs.md
```diff
@@ -14,6 +14,12 @@ but rather to list those codecs that are currently supported in Matroska and the
 need a well defined `CodecID` so that all developers supporting Matroska will use the
 same `CodecID`.
 
+Because the list of codec is extensive and will grow in the future, Matroska Players are not
+expected to be able to handle all the possible codec defined in this document.
+There is also no minimal set of codec that a Matroska Player should be capable of decoding.
+When a Matroska Player encounters a `TrackEntry` with a `CodecID` it cannot handle, it should discard
+the data related to this track.
```
SpencerDawkins, line 21:
```suggestion
Because the list of codecs is extensive and will grow in the future, Matroska Players are not
expected to be able to handle all the possible codecs the Players might encounter.
There is also no minimal set of codecs that a Matroska Player is expected to be capable of decoding.
When a Matroska Player encounters a `TrackEntry` with a `CodecID` it cannot handle, it MUST discard
the data related to this track.
```

robUx4: OK with the plural changes. But we can make normative decisions for the player behavior. They could completely reject a file if they don't know a codec. Or they could launch another player to play it, etc.

robUx4: https://github.com/ietf-wg-cellar/matroska-specification/pull/1127

PR #1121, opened 2026-08-09
codec: add a reference to C for the Wavpack pseudo code

robUx4:


cellar-codec/rfc_backmatter_codec.md
```diff
@@ -193,6 +193,17 @@
   </front>
 </reference>
 
+<reference anchor="ISO9899" target="https://www.iso.org/standard/74528.html">
```
retokromer: This version is no longer available on their site.

robUx4: Indeed, but that's C23, not C11 anymore :(

robUx4: (the 2018 was actually C17)

retokromer: >Indeed, but that's C23, not C11 anymore :(

I don’t know if the old ones can be accessed. In the past – i.e. 15 years ago or so – that was not possible. Sadly.

(On my end, I moved to C17, but I cannot remember why exactly.)

retokromer (approved):

PR #1124, opened 2026-08-09
codec: add a set of rules Designated Experts need to review IANA entries

robUx4:
Plus make sure the new registry goes in the Matroska group

### DIFF
--- a/cellar-codec/codec_iana.md
+++ b/cellar-codec/codec_iana.md
@@ -2,7 +2,9 @@
 
 ## Matroska Codec IDs Registry
 
-This document defines registries for Codec IDs stored in the `CodecID` element.
+This document defines a new registry for Codec IDs stored in the `CodecID` element.
+IANA is requested to create the this registries under the existing "Matroska" registry group.
+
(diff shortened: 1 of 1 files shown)


cellar-codec/codec_iana.md
```diff
@@ -13,6 +13,15 @@ The details of the string format are found in (#codec-id).
 To register a new Codec ID in this registry, one needs a Codec ID string, a TrackType value,
 a description, a Change Controller, and an optional Reference to a document describing the Codec ID.
 
+Designated Experts should verify these rules are matched:
+
+* the different items defined in (#defining-matroska-codec-support) are properly defined,
```
retokromer: Since these are complete sentences, I would personally use a semicolon (or a capital first letter and a period) and not a comma to separate the different points.

retokromer (commented):
a nit, otherwise LGTM

cellar-codec/codec_iana.md
robUx4: OK, it seems more common that way. If there's a problem the RFC editors will say so.

